PR #398, opened 2013-01-17
Adding support for multiple authentication providers per route

### LINKED ISSUE #307: Support multiple authentication strategies per route with preferences
For example, accept Hawk and Basic on a single endpoint. Return 401 WWW-Authenticate header with multiple challenges in order of preference.

geek:
Closes #307


fedirch:
is it or will be it released?

Issue #696, opened 2013-03-20
Provide easier debugging defaults


kpdecker:
Earlier I hit a case where my code was causing a 500 response. This produced zero feed back other than the 500 response in my development environment.

After chatting with @hueniverse it turns out there are events that can be used to debug this situation but you have to write custom code to handle this.

This wasn't very accessible as someone who is not familiar with the framework as it masks any stack trace or other debugging information. It would be much friendlier to users to perform some sort of logging of this data by default.

Issue #695, opened 2013-03-20
Allow for true 404 handler


kpdecker:
Should allow for multiple handlers for a given route to allow for implementation of 404 handlers for cases that the default can not handle them, i.e. directory handling.

This is particularly important if trying to serve a directory from `/{path*}` as the documented manner of handling 404s conflicts with this route.

PR #725, opened 2013-03-29
Debug enhancements

### LINKED ISSUE #703: taking a look at http data posted before 400 response
Hi,
I have a post method that always returns 400 bad request, I'd like to take a look at what has been posted from another program, trying to use a 'pre' as following, but nothing was shown, any idea how to take a look at data in http post before the 400 bad request response? thanks,

<pre><code>
var fetch1 = function (request, next) {
  console.log(request)
  next('Hello');
};

var submit = {
  pre: [

    { method: fetch1, assign: 'm3' }
  ],

  handler: function (request) {})
]
</code></pre>

hueniverse:
Closes #703 #698 #696

### DIFF
--- a/docs/Reference.md
+++ b/docs/Reference.md
@@ -21,6 +21,7 @@
     - [CORS](#cors)
     - [State](#state)
     - [Timeout](#timeout)
+    - [Debug](#debug)
 <p></p>
 - [**Server Events**](#server-events)
 <p></p>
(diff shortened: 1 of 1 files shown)


andreas-wagner:
How exactly does this close 'Allow for true 404 handler #695' which was in #698?
Real question is how do I catch the 404 of a directory handler, so i can for example show the user a custom error page?

andreas-wagner:
I've accomplished it by using a hook with server.ext('onPreResponse', function (request, reply) { ... });.
Is it better to additionally use the catch-all server.route({ method: '*', path: '/{p*}', handler: handler });?

PR #2309, opened 2014-12-19
Fixes #2308 by logging boom error object instead of just message

### LINKED ISSUE #2308: Logging boom errors from handlers should send boom error to log not just message
When the response value is a `boom` object the following code is used to log (from [here](https://github.com/hapijs/hapi/blob/master/lib/handler.js#L78-L81)):
```
if (response.isBoom) {
    request._log(['handler', 'error'], { msec: timer.elapsed(), error: response.message });
    return callback(response);
}
```

This does not send the entire error instance but only the message. It is can be valuable to be able to log additional information such as the status code.

Proposal:
```
if (response.isBoom) {
    request._log(['handler', 'error'], { msec: timer.elapsed(), error: response });
    return callback(response);
}
```

Sending a PR with the change and a test.

dschenkelman:


lib/handler.js
```diff
@@ -76,7 +76,7 @@ internals.handler = function (request, callback) {
         // Check for Error result
 
         if (response.isBoom) {
-            request._log(['handler', 'error'], { msec: timer.elapsed(), error: response.message });
+            request._log(['handler', 'error'], { msec: timer.elapsed(), error: response });
```
hueniverse: This would be a breaking change. Instead, just add `data` field with the full response.

dschenkelman:
Is there anything else that needs to be done from my side? I understand you might be busy, just asking in case something is expected.

PR #2324, opened 2014-12-28
Remove catch call for promise replies

### LINKED ISSUE #2323: Promise support
Would it be appropriate to change [this line](https://github.com/hapijs/hapi/blob/master/lib/response.js#L375) in `response.js` from:

`this.source.then(onDone, onDone).catch(onDone);`

to

`this.source.then(onDone, onDone).then(null, onDone);`

The reason for this is that some promise implementations still don't have `catch` implemented.
A major example is Mongoose/mpromise.
Open to discussion.

bendrucker:
Adds compatibility for promises that implement the bare A+ spec

Closes #2323

### DIFF
--- a/lib/methods.js
+++ b/lib/methods.js
@@ -101,7 +101,7 @@ internals.Methods.prototype._add = function (name, method, options, realm) {
                 return methodNext(err);
             };
 
-            result.then(onFulfilled, onRejected).catch(onRejected);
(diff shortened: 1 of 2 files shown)


hueniverse:
What about https://github.com/hapijs/hapi/blob/master/lib/methods.js#L104?

bendrucker:
Right you are. Amended.

PR #2343, opened 2015-01-09
throw an appropriate error when payload validation is used with GET or HEAD routes

### LINKED ISSUE #2347: Improve error message when validation.payload is set but type is GET
Currently you get an error looking like: 

```
/Users/wraithan/devel/javascript/alert-alpaca/node_modules/hapi/lib/route.js:126
Hoek.assert(!this.settings.validate.payload || this.settings.payload.parse, 'R
                                                                    ^
TypeError: Cannot read property 'parse' of null
    at new module.exports.internals.Route (/Users/wraithan/devel/javascript/alert-alpaca/node_modules/hapi/lib/route.js:126:73)
    at internals.Connection._addRoute (/Users/wraithan/devel/javascript/alert-alpaca/node_modules/hapi/lib/connection.js:344:17)
    at internals.Connection._route (/Users/wraithan/devel/javascript/alert-alpaca/node_modules/hapi/lib/connection.js:336:18)
    at internals.Plugin._apply (/Users/wraithan/devel/javascript/alert-alpaca/node_modules/hapi/lib/plugin.js:432:14)
    at internals.Plugin.route (/Users/wraithan/devel/javascript/alert-alpaca/node_modules/hapi/lib/plugin.js:407:10)
    at Object.<anonymous> (/Users/wraithan/devel/javascript/alert-alpaca/index.js:7:8)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
```

It looks like there used to be a nicer error that was removed in a refactor:

https://github.com/hapijs/hapi/commit/4299e9e7513da024e577f8a287431c4578a14d15#diff-f34e409694d3ba334dabbea190f64c1dL117

damirv:
When a user mistakenly uses a payload validation on a GET or HEAD route, an appropriate error is given, instead of

```
TypeError: Cannot read property 'parse' of null
   at new module.exports.internals.Route (... node_modules/hapi/lib/route.js:126:73)
```

This might help new users that make this mistake more quickly to uncover what's going on.

Closes #2347

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -123,6 +123,7 @@ exports = module.exports = internals.Route = function (options, connection, real
         }
     }
 
+    internals.checkPayloadValidation(this);
     Hoek.assert(!this.settings.validate.payload || this.settings.payload.parse, 'Route payload must be set to \'parse\' when payload validation enabled:', options.method, options.path);
     Hoek.assert(!this.settings.jsonp || typeof this.settings.jsonp === 'string', 'Bad route JSONP parameter name:', options.path);
 
@@ -245,6 +246,8 @@ internals.compileRule = function (rule) {
 
 
 internals.Route.prototype.lifecycle = function () {
+
+    internals.checkPayloadValidation(this);
 
     var cycle = [];
 
@@ -387,4 +390,9 @@ internals.parseJSONP = function (request, next) {
     }
 
     return next();
+};
+
+// check that payload validation is not used with a GET or HEAD method
+internals.checkPayloadValidation = function (self) {
+    Hoek.assert(!self.settings.validate.payload || self.settings.payload, 'Payload validation cannot be used with \'get\' or \'head\' HTTP methods');
 };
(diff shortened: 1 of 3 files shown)


hueniverse:
It needs to ignore it at the route lifecycle setup step, not assert. This is because you can now set a global route validation default for all routes.

damirv:
Sure, where can I find the route lifecycle setup code?

hueniverse:
Setup is here: https://github.com/hapijs/hapi/blob/master/lib/route.js#L247
But method can be `*` so need to skip it here: https://github.com/hapijs/hapi/blob/master/lib/validation.js#L19

damirv:
I will try to fix this tomorrow.

damirv:
I am unfamiliar with global route validation, is there any documentation on that? (couldn't find anything)

Currently, I have added it to the two suggested places. However, I also added the check where I first made the change, because that is where error originally occurred.

I also included a simple test case. If you point me to the doc on global route validation, I can add that test case as well and pass the 100% coverage requirement.

hueniverse:
Solved slightly differently.

PR #2345, opened 2015-01-09
Emit response only when response is available

### LINKED ISSUE #2344: response event with null response
>  'response' - emitted after a response to a client request is sent back. Single event per request.

What if a request is terminated by client so that server cannot send back a response?

I actually encountered request.response is null for `response`event - see hapijs/good#285 

My use case is basically token auth with mysql backend. I used following components

- hapi@8.1.0
- hapi-auth-bearer-token@3.1.0
- good@5.1.1
- good-console@4.1.0
- mysql@2.5.4

At the end of wrk test (and it happens only at the last few requests), I got this error:

```
Debug: internal, implementation, error
    TypeError: Cannot read property 'statusCode' of null
    at internals.NetworkMonitor._onResponse (/somewhere/node_modules/good/lib/network.js:40:38)
    at emit (events.js:95:17)
    at handler (/somewhere/node_modules/hapi/node_modules/kilt/lib/index.js:54:44)
    at emit (events.js:95:17)
    at internals.Request._finalize (/somewhere/node_modules/hapi/lib/request.js:378:21)
    at internals.Request._reply (/somewhere/node_modules/hapi/lib/request.js:344:21)
    at /somewhere/node_modules/hapi/lib/request.js:325:18
    at done (/somewhere/node_modules/hapi/node_modules/items/lib/index.js:22:21)
    at /somewhere/node_modules/hapi/lib/request.js:314:24
    at iterate (/somewhere/node_modules/hapi/node_modules/items/lib/index.js:35:13)
```
I could reproduce request.response is null in test code - see #2345

From a close look (my first look on hapi internal so correct me if I'm wrong), currently `internals.Request.prototype._finalized` is called even `_isBailed`=true. `_isBailed` is set to true when the request (not response) is closed. So.. probably it happens when request is closed before response is actually created somehow.

One possible solution is just changing the doc `response` event can be emitted even response is not send back... which is misleading :(

Or emit `response`only when response is actually send back. This may break existing app which relies on this bug, but sounds better to me. Those app probably should use `tail` event.

chulkilee:
Fixes #2344

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -375,7 +375,9 @@ internals.Request.prototype._reply = function (exit) {
 internals.Request.prototype._finalize = function () {
 
     this.info.responded = Date.now();
-    this.connection.emit('response', this);
+    if (this.response) {
+        this.connection.emit('response', this);
+    }
 
     this._isFinalized = true;
     this.addTail = undefined;
--- a/test/payload.js
+++ b/test/payload.js
@@ -79,8 +79,9 @@ describe('payload', function () {
         server.connection();
         server.route({ method: 'POST', path: '/', config: { handler: handler } });
 
-        server.once('response', function (request) {
+        server.once('tail', function (request) {
 
+            expect(request.response).to.null()
             expect(request._isBailed).to.equal(true);
             done();
         });


chulkilee:
@hueniverse could you review this pull request?

hueniverse:
This changes the contract and is not the right approach.

PR #2359, opened 2015-01-22
Remove '{}' payload from cors OPTIONS response

kanongil:
While I guess the payload is allowed, it is quite redundant and adds a  `content-type` header.


kanongil:
Note that travis failed because of hapijs/lab#295.

jzorn:
Oh that's a nice one!

PR #2362, opened 2015-01-22
Don't respond to connections until listening is started

kanongil:
This also fixes the _removes connection event listeners after it stops_ test to do what it says.


hueniverse:
What exactly is the problem it is trying to solve? Why would there be connections before the listener is active?

kanongil:
The test I fixed was supposed to check that a stopped server will remove attached connection listeners. The code did not do this, so I fixed it by delaying the attaching until the server is actually listening (after someone has called start).

If you don't care about the extra listener, the test should probably be revised to check that the listener count matches after each `stop()` call.

PR #2363, opened 2015-01-22
Refuse to handle incoming request after server is stopped

kanongil:
As described in joyent/node#9066, `node` can continue to emit `request` events on closed http servers. This patch works around this by checking if the server is stopping before responding to incoming requests.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -58,6 +58,7 @@ exports = module.exports = internals.Connection = function (server, options) {
     // Connection facilities
 
     this._started = false;
+    this._stopping = false;
     this._connections = {};
     this._onConnection = null;          // Used to remove event listener on stop
     this._registrations = {};           // Tracks plugin for dependency validation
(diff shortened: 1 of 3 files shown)


lib/connection.js
```diff
@@ -202,6 +206,10 @@ internals.Connection.prototype._dispatch = function (options) {
 
     return function (req, res) {
 
+        if (self._stopping) {
```
hueniverse: Why not just check `_started`?

kanongil: Because `_started` is always `false` when using `inject()`.

kanongil:
I'm sure there are other ways to fix this, but this seemed quite simple. Another approach could be to remove the `request` listener when close is called. However, we would not be able to `end()` the underlying connection before the timeout, which I'm not entirely certain is a good idea anyway, as it will probably abort pending pipelined responses.

lib/connection.js
hueniverse: Not if you also check for `req.connection`.

PR #2370, opened 2015-01-26
Add xss protection to validation response

shinyaohira:
Validation field should also be escaped.
```json
{"statusCode":400,"error":"Bad Request","message":"&lt;script&gt;&lt;&#x2f;script&gt; is not allowed","validation":{"source":"query","keys":["<script></script>"]}}
```


kanongil:
Why should it be escaped? I would think that it is the receivers responsibility not to expose validation keys without escaping.

shinyaohira:
Because attackers don't escape it intentionally. It was fixed by #697, but I think it is incomplete.

PR #2383, opened 2015-01-31
Update README

danilovaz:
added using hapi and use plugins, this information is the same site


hueniverse:
Thanks for the pull request.

hapijs.com is the main site for info. People should use that as it reflects the current published state. This information was removed from the readme for a reason.

danilovaz:
for this, readme is much better. anyway

PR #2400, opened 2015-02-11
Override server realm with route realm inside of handler

### LINKED ISSUE #2278: request.server.render throws "Uncaught error: Missing views manager" in plugin
I am trying to upgrade from v6 to v8. I have a plugin that I register like so
```javascript
exports.register = function (server, options, next) {

   	server.views({
        engines: { jade: require('jade') },
		compileOptions: {
			pretty: true
		},
        path:  __dirname + '/views'
    });

   server.route({
		method: 'POST', 
		path: '/path/to/resource', 
		handler:resource.create, 
		config:{
			validate:{				
				payload: require('./schemas/resource')
			}
		}
	})
```

and in my route handler
```javascript
var create = function(request,reply){
	
	var context = {}
	
	request.server.render("name_of_view", context, null, function(err, rendered, settings){
```

 ```Hoek.assert(vision.manager, 'Missing views manager');``` throws because manager is undefined.

I stepped into ```server.views({..})```, and ```vision.manager``` does get created.

savardc:
Use the realm for the current route inside of handler so they have access to state added during the plugin registration

Fix #2278

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -89,6 +89,8 @@ internals.handler = function (request, callback) {
     var reply = request.server._replier.interface(request, request.route.realm, finalize);
     var bind = request.route.settings.bind;
 
+    request.server.realm = request.route.realm;
+
     // Execute handler
 
     request.route.settings.handler.call(bind, request, reply);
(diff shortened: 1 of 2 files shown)


hueniverse:
This is the wrong way to fix it. I am still thinking of how I want to change the realm implementation to handle his case.

PR #2402, opened 2015-02-11
Improved validation of route method label

kanongil:
This patch improves route method option handling in 3 ways:

 1. It fixes a bug in the validation which allows a nested array (eg.  `['GET', ['POST']]`) but later crashes when creating the route.
 1. It enforces the non-`HEAD` method restriction from the API docs.
 1. It limits the allowed value to token characters from https://tools.ietf.org/html/rfc7230#section-3.2.6.

Finally, I have included a couple of tests for `HEAD` responses, validating the existing implementation.

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -35,11 +35,13 @@ exports = module.exports = internals.Route = function (options, connection, real
     Hoek.assert(options.handler || (options.config && options.config.handler), 'Missing or undefined handler:', options.method, options.path);
     Hoek.assert(!!options.handler ^ !!(options.config && options.config.handler), 'Handler must only appear once:', options.method, options.path);            // XOR
     Hoek.assert(options.path === '/' || options.path[options.path.length - 1] !== '/' || !connection.settings.router.stripTrailingSlash, 'Path cannot end with a trailing slash when connection configured to strip:', options.method, options.path);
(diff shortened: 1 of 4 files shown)


lib/route.js
```diff
@@ -35,11 +35,13 @@ exports = module.exports = internals.Route = function (options, connection, real
     Hoek.assert(options.handler || (options.config && options.config.handler), 'Missing or undefined handler:', options.method, options.path);
     Hoek.assert(!!options.handler ^ !!(options.config && options.config.handler), 'Handler must only appear once:', options.method, options.path);            // XOR
     Hoek.assert(options.path === '/' || options.path[options.path.length - 1] !== '/' || !connection.settings.router.stripTrailingSlash, 'Path cannot end with a trailing slash when connection configured to strip:', options.method, options.path);
+    Hoek.assert(/^[a-zA-Z0-9!#\$%&'\*\+\-\.^_`\|~]+$/.test(options.method), 'Invalid method name:', options.method, options.path);
```
arb: Wouldn't it be faster and probably easier to read to just have a method object and reference that?

```js
var _methdods = { get: true, post: true...}
Hoek.assert(_methods[options.method.toLowerCase()])
```
or something along those lines?

kanongil: I don't know why you are concerned about speed in route creation. As for whitelisting, that would make sense as a future option but right now I think it is a breaking change.

This patch is mainly concerned with catching route names that should never be allowed, like `GET /`.

PR #2405, opened 2015-02-11
Support inline dependencies on plugins

### LINKED ISSUE #2332: Replacing plugin.dependency() with attributes key
Would it be a cleaner idea to replace `plugin.dependency()` with an extension to `register.attributes`?

Rather than

```js
var resolved = function (server, next) {

    server.route({}) // etc
    next()
}

var register = function (server, options, next) {

    server.dependency("lout", resolved)
    next()
}

register.attributes = {
    name: "test",
    version: "1.0.0"
}
```

it'd be possible to do something like this

```js
var register = function (server, options, next) {

    server.route({})
    next()
}

register.attributes = {
    name: "test",
    version: "1.0.0",
    dependencies: "lout"
}
```

I can't see why this wouldn't work out at first look.

Marsup:
Fixes #2332.

I have a small doubt on L248 of this patch, `server` is the same as `selection` right ?

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -216,10 +216,7 @@ internals.Plugin.prototype.register = function (plugins /*, [options], callback 
         };
 
         Hoek.assert(item.name, 'Missing plugin name', hint);
-        Hoek.assert(typeof item.dependencies === 'string' ||
-            Array.isArray(item.dependencies) &&
-            item.dependencies.every(internals.isString),
-            'Dependencies must be a string or an array of strings');
+        Schema.assert('dependencies', item.dependencies, 'must be a string or an array of strings');
 
         registrations.push(item);
     }
@@ -255,12 +252,6 @@ internals.Plugin.prototype.register = function (plugins /*, [options], callback 
 
         item.register(selection, item.options || {}, next);
     }, callback);
-};
-
-
-internals.isString = function (str) {
-
-    return typeof str === 'string';
 };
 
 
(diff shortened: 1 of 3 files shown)


lib/plugin.js
```diff
@@ -211,10 +211,16 @@ internals.Plugin.prototype.register = function (plugins /*, [options], callback
             name: attributes.name || (attributes.pkg && attributes.pkg.name),
             version: attributes.version || (attributes.pkg && attributes.pkg.version) || '0.0.0',
             multiple: attributes.multiple || false,
-            options: plugin.options
+            options: plugin.options,
+            dependencies: attributes.dependencies || []
         };
 
         Hoek.assert(item.name, 'Missing plugin name', hint);
+        Hoek.assert(typeof item.dependencies === 'string' ||
```
hueniverse: Why not a joi.assert here?

Marsup:
@hueniverse like that or do you want to check the whole item object ? I don't know if doing it now would be a breaking change for some people.

hueniverse:
This is fine, but why a breaking change for some people?

lib/plugin.js
```diff
@@ -230,6 +233,21 @@ internals.Plugin.prototype.register = function (plugins /*, [options], callback
...
         }
 
+
+        if (item.dependencies.length) {
+
+            // Declare dependencies
+
+            selection.dependency(item.dependencies, function (server, next) {
+
+                // Deferred registration when all dependencies are resolve
```
hueniverse: This is a change in behavior. Dependency does not imply order. If you want to dictate order of registration you can't use this mechanism.

Marsup: I don't think there's an order to have here, dependencies in registration all just have to be there before the plugin begins his own registration. How is that a change if there wasn't any mechanism for that before ?

Marsup:
@hueniverse breaking because depending on how we validate the object, we might reject existing configurations if it's too strict.

expr:
Sorry about not working on this. Glad it's landed though :-)

hueniverse:
@Marsup not if we allow unknown keys :-)

PR #2417, opened 2015-02-18
Closes #2414

### LINKED ISSUE #2414: test occasionally fails
When running npm test on hapi locally, I see the following test fail occasionally:

  344) Request _execute() handles aborted requests:
          
                Timed out (2000ms) - Request _execute() handles aborted requests

This doesn't happen all the time.  I chatted with @cjihrig as well and he said he noticed this on 0.10 latest (I noticed it when I was using 0.12).

lloydbenson:
This fixes the occasional test issue but it may be hiding a bug.

### DIFF
--- a/test/request.js
+++ b/test/request.js
@@ -223,9 +223,9 @@ describe('Request', function () {
 
                 var check = function () {
 
-                    if (total > 0) {
+                    if (total) {
                         createConnection();
-                        setImmediate(check);
+                        setTimeout(check, 10);
                     }
                     else {
                         done();
--- a/test/request.js
+++ b/test/request.js
@@ -223,7 +223,7 @@ describe('Request', function () {
 
                 var check = function () {
 
-                    if (total) {
+                    if (total > 0) {
                         createConnection();
                         setImmediate(check);
                     }


kanongil:
I think a more stable solution is to change
```javascript
client.on('data', function () {
```
to
```javascript
client.once('data', function () {
```

This will ignore any extra 'data' emits.

lloydbenson:
Thanks for the suggestion!  This does seem to work as well.  I'll update the PR with this as it seems a bit better.  I still see things like:

2
2
2
1
-3

and 

2
2
-1

Again not sure if this is totally normal or not.

lloydbenson:
I spoke too soon, I still had the > 0 when I said it was passing with once.  As you can see above if I do once, it still skips and hangs the timeout.

kanongil:
Ah, I see real the issue now. `setImmediate(check)` can be called any number of times before the responses are processed, meaning that the test itself is flawed.

Regarding a proper fix, I can't really say, as I have a hard time to figure out what is actually tested. Eg. it contains no `expect()` calls.

Unless someone has a more specific idea of what is tested, I say that the most appropriate fix is to remove the test. Maybe @geek that created it in 1937aefc9ce64e23cd7cb0a6b8e29865cc8d793b?

PR #2418, opened 2015-02-18
Add regex to allow leading $ and _

### LINKED ISSUE #2401: Method name RegExp
Currently, the method name is validated with [this regexp](https://github.com/hapijs/hapi/blob/2a9ce7196e9db45a99a271ba36da4df5384066ae/lib/methods.js#L38):

```javascript
/^[a-zA-Z]\w*(?:\.[a-zA-Z]\w*)*$/
```

You could allow any [identifier name](http://www.ecma-international.org/ecma-262/5.1/#sec-7.6), that is, any string starting with `$` or `_`. For instance, I need to create a method with a name starting with underscore and I can't.

```
Error: Invalid name: _common.injectQueryClientAndAncestors
```

ckross01:
Fixes #2401
added to the regex to allow leading $ and _
added two tests to validate that the leading $ _ is allowed for a method name

### DIFF
--- a/lib/methods.js
+++ b/lib/methods.js
@@ -35,7 +35,7 @@ internals.Methods.prototype.add = function (name, method, options, realm) {
 };
 
 
-exports.methodNameRx = /^[_$a-zA-Z]\w*(?:\.[a-zA-Z]\w*)*$/;
+exports.methodNameRx = /^[_$a-zA-Z]\w*(?:\.[_$a-zA-Z]\w*)*$/;
 
 
 internals.Methods.prototype._add = function (name, method, options, realm) {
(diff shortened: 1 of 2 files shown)


lib/methods.js
```diff
@@ -35,7 +35,7 @@ internals.Methods.prototype.add = function (name, method, options, realm) {
 };
 
 
-exports.methodNameRx = /^[a-zA-Z]\w*(?:\.[a-zA-Z]\w*)*$/;
+exports.methodNameRx = /^[_$a-zA-Z]\w*(?:\.[a-zA-Z]\w*)*$/;
```
hueniverse: Probably worth adding that to the other text block so you can do `_a._b`

PR #2428, opened 2015-02-23
Create an entry at server.plugins on plugin registration

fampinheiro:
I need to check if a plugin is registered and reading the [documentation](http://hapijs.com/api#serverplugins) I was trying to do it testing the presence of `server.plugins['plugin-name']`.

Since the `server.plugins['plugin-name']` is only assigned when the expose method is called I moved that logic to the plugin construction.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -37,6 +37,7 @@ exports = module.exports = internals.Plugin = function (server, connections, env
     this.plugins = this.root._plugins;
     this.settings = this.root._settings;
     this.version = Package.version;
+    this.plugins[env] = {};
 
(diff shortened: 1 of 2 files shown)


hueniverse:
This would be a breaking change as the only value set right now is what is being exposed. This will now initialize the key with something else. If you need to check if a plugin is loaded, the plugin should actively register this information. If you are trying to work with someone else's plugin, open a new issue requesting the ability to know which plugin is loaded. Note that it is tricky when a plugin is loaded multiple times.

fampinheiro:
Ok and thanks for the heads up on the possibility of issues in a multi plugin environment.

Issue #2427, opened 2015-02-21
Issue uploading file with io.js or node > 0.11.6


dschenkelman:
When trying to migrate my app to either node 0.12 or io.js (any version) I started seeing an issue that is not always reproduced (seems to be a race condition of some sort) when uploading files.

Basically, when uploading a file larger than that allowed by the payload `maxBytes` we get an error:
```
{ [Error: write EPIPE]
  code: 'EPIPE',
  errno: 'EPIPE',
  syscall: 'write',
  address: undefined }
```

This does not happen if the size of the file is under the limit.

I put together a repository that reproduces the issue: https://github.com/dschenkelman/hapi-upload-iojs.

pandeysoni:
You can visit for working code in https://github.com/pandeysoni/Hapi-file-upload-download

    /*
     * upload file
     */
    
    exports.uploadFile = {
        payload: {
            maxBytes: 209715200,
            output: 'stream',
            parse: false
        },
        handler: function(requset, reply) {
            var form = new multiparty.Form();
            form.parse(requset.payload, function(err, fields, files) {
                if (err) return reply(err);
                else upload(files, reply);
            });
        }
    };
    
    /*
     * upload file function
     */
    
    var upload = function(files, reply) {
        fs.readFile(files.file[0].path, function(err, data) {
            checkFileExist();
            fs.writeFile(Config.MixInsideFolder + files.file[0].originalFilename, data, function(err) {
                if (err) return reply(err);
                else return reply('File uploaded to: ' + Config.MixInsideFolder + files.file[0].originalFilename);
    
            });
        });
    };

dschenkelman:
Hey @pandeysoni,

Thanks for the example. Just to clarify, it is not that I'm not able to get file upload to work, I have a working application using 0.10.30.

The issue is that in versions of node > 0.11.6 and in io.js instead of always getting a 400 in the request I sometimes get the error mentioned above.

pandeysoni:
Hi @dschenkelman, I cloned your github repo and i tested with node : v0.10.31 and it is throwing error with version 0.10 also.

dschenkelman:
Have just tried running 10 times with node 0.10.30 and did not get the error. Running 5 times with io.js 1.2.0 I got the error 4 times.

kanongil:
I had a look at this. I am unable to reproduce the error on node 0.12. I ran the client 50'ish times.

Which system are you on?

pandeysoni:
It's on windows 7.

[...]

PR #2429, opened 2015-02-24
Fix for #2427

kanongil:
This includes a new test that fails on all version of node.

I'm not entirely sure that the fix I have implemented is the best way to handle this issue but it seems to do the trick without messing up any other tests.


hueniverse:
Removing the fix does not break the test.

kanongil:
@hueniverse Are you sure? I just checked on master and it still fails for me.

FYI, I'm on node v0.10.36 on Yosemite.

kanongil:
@hueniverse I can see you have made your own fix, which is probably wise. However it has issues on node 0.10 as seen in this travis log: https://travis-ci.org/hapijs/hapi/jobs/53393303

It seems your fix requires streams3 for full support.

hueniverse:
@kanongil yeah, I need to find a way around that to first drain the stream. Have been busy with other stuff.

PR #2439, opened 2015-03-03
pass context to response schema validation

nlf:
Extends response schema validation to receive the same default context as input validation currently does.


hueniverse:
Missing documentation.

nlf:
@hueniverse need anything else from me on this?

PR #2499, opened 2015-04-13
Add raw payload

rysenko:
See https://github.com/hapijs/subtext/pull/8 for argumentation why it's sometimes preferred to have both parsed and unparsed payload.


hueniverse:
Need to resolve the subtext issue first.

Issue #2525, opened 2015-05-07
update changelog


AdriVanHoudt:
more a reference issue, i'll give it a go but if anyone wants to try feel free ;)

AdriVanHoudt:
Ok so somewhere there is a tag of 2.8 screwing up the changelog generation :neutral_face: 
This is as far as I got. Pinging @diffsky for help
Also I had to set `--remote hapijs/hapi` to make it work
```js
$ make changelog MDCHANGELOG_TOKEN=0ccd76d9778b184d6e7cd3cf305b7a61b711e204
[hapijs/hapi] fetching 1356/1356 issues [====================] 100% 0.0s
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
[hapijs/joi] fetching 334/334 issues [====================] 100% 0.0s
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
[hapijs/qs] fetching 48/48 issues [====================] 100% 0.0s
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
[moment/moment] fetching 712/712 issues [====================] 100% 0.0s
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
[hapijs/shot] fetching 17/17 issues [====================] 100% 0.0s
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
[hapijs/catbox] fetching 80/80 issues [====================] 100% 0.0s
[hapijs/shot] fetching 17/17 issues [====================] 100% 0.0s
[hapijs/joi] fetching 334/334 issues [====================] 100% 0.0s
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
[hapijs/inert] fetching 16/16 issues [====================] 100% 0.0s
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
[hapijs/wreck] fetching 41/41 issues [====================] 100% 0.0s
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
Enter passphrase for key '/c/Users/adri/.ssh/id_rsa':
[hapijs/hoek] fetching 98/98 issues [====================] 100% 0.0s
[hapijs/shot] fetching 17/17 issues [====================] 100% 0.0s
[...]

PR #2527, opened 2015-05-08
#2525 update changelog and generation command

diffsky:
mdchangelog to 0.7 supports restricting dependents to the same github organization. This PR:
- uses the new flag
- uses the remote flag to allow folks to run mdchangelog from their forks
- has an updated changelog

Should mdchangelog be a devdependency?


AdriVanHoudt:
closes #2525

PR #2532, opened 2015-05-12
Dynamic authentication scopes

nlf:
I took the simplest approach that I could think of here. Let me know if anything should be changed.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -125,11 +125,11 @@ internals.Auth.prototype._setupRoute = function (options, path) {
         options = Hoek.applyToDefaults(this.settings.default, options);
     }
 
-    if (typeof options.scope === 'string') {
-        options.scope = [options.scope];
-    }
-
     if (options.scope) {
+        if (typeof options.scope === 'string') {
+            options.scope = [options.scope];
+        }
+
         for (var i = 0, il = options.scope.length; i < il; ++i) {
             if (/{([^{}]+)}/g.test(options.scope[i])) {
                 options.hasScopeParameters = true;
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -125,6 +125,10 @@ internals.Auth.prototype._setupRoute = function (options, path) {
         options = Hoek.applyToDefaults(this.settings.default, options);
     }
 
+    if (typeof options.scope === 'string') {
+        options.scope = [options.scope];
+    }
+
     Hoek.assert(options.strategies.length, 'Route missing authentication strategy:', path);
 
     options.mode = options.mode || 'required';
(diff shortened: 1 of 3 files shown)


lib/auth.js
```diff
@@ -269,11 +269,25 @@ internals.Auth.prototype._authenticate = function (request, next) {
 
         // Check scope
 
+        var expandScope = function (scope) {
+
+            return scope.replace(/{([^{}]*)}/g, function (_, context) {
```
Marsup: Are you sure you don't you want + instead of * ?

nlf:
Thinking it may be wise to white list the parameters the dynamic scope is able to access, allowing anything on the request seems like asking for trouble

nlf:
Ok, I think I feel ok about this since it only allows the scope to contain things the user is sending in. This should eliminate the possibility of a malicious user exploiting this feature to leak data.

lib/auth.js
```diff
@@ -269,11 +269,29 @@ internals.Auth.prototype._authenticate = function (request, next) {
 
         // Check scope
 
+        var expandScope = function (scope) {
```
hueniverse: Does this needs to be a function? Why not just inline the logic inside the for loop?

lib/auth.js
```diff
@@ -269,11 +269,29 @@ internals.Auth.prototype._authenticate = function (request, next) {
...
+                    params: request.params,
+                    query: request.query,
+                    payload: request.payload
+                }, context);
+            });
+        };
+
         if (config.scope) {
+            if (typeof config.scope === 'string') {
+                config.scope = [config.scope];
```
hueniverse: This is wasteful. We avoid creating objects like this outside of the config step. This code runs on every request and in 99% of the cases, scope will be a single string.

test/auth.js
```diff
@@ -628,6 +628,30 @@ describe('authentication', function () {
...
+            server.connection();
+            server.auth.scheme('custom', internals.implementation);
+            server.auth.strategy('default', 'custom', true, { users: { steve: { scope: 'one-test' } } });
+            server.route({
+                method: 'GET',
+                path: '/{id}',
+                config: {
+                    handler: function (request, reply) { return reply(request.auth.credentials.user); },
+                    auth: {
+                        scope: 'one-{params.id}'
```
hueniverse: Need to test partial `{}` (e.g. one side missing), along with multiple `{}` in one scope if that feature is kept.

lib/auth.js
nlf: if by multiple bits, you mean would they be allowed to compose their scope with multiple dynamic values, then yes definitely.

lib/auth.js
nlf: This was because the linter throws a "don't make functions within a loop" warning, so I moved it out of the loop

nlf:
added the tests you wanted, moved the code to make sure the scope is an array to route setup, and tweaked the loop stuff a little. i kept the logic of the replacing in a function to avoid the "don't create functions in a loop" linter warning.

lib/auth.js
```diff
@@ -270,10 +274,21 @@ internals.Auth.prototype._authenticate = function (request, next) {
...
+
+                return Hoek.reach({
+                    params: request.params,
+                    query: request.query,
+                    payload: request.payload
+                }, context);
+            };
+
+            for (var i = 0, il = config.scope.length; i < il; ++i) {
+                config.scope[i] = config.scope[i].replace(/{([^{}]+)}/g, expandScope);
```
hueniverse: You need to set a flag when a route contains a scope with valid `{}` template parameter, then only call this replace logic if that flag is set. Otherwise you are wasting cycles for the 99% not using this feature.

PR #2533, opened 2015-05-12
clean up lint

nlf:
after https://github.com/cjihrig/hapi-scope-start/pull/1 is merged this will clean up all the remaining linter warnings

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -121,7 +121,7 @@ internals.Connection.prototype._init = function () {
             self.info.uri = (self.settings.uri || (self.info.protocol + '://' + self.info.host + ':' + self.info.port));
         }
 
-        self._onConnection = function _onConnection(connection) {
+        self._onConnection = function _onConnection (connection) {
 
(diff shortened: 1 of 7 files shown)


Marsup:
Why aren't you also setting the lab options to enforce it in the same PR ?

nlf:
It's already set, isn't it?

Marsup:
Nope.

nlf:
What's missing, I thought it was just the `-L` flag?

Marsup:
--lint-errors-threshold and --lint-warnings-threshold. Probably want to set it to 0.

Marsup:
Why's that makefile still here btw ? :)

nlf:
good question :) if there are no objections i can remove it in this PR also

Marsup:
Not without adding the equivalent commands in the npm scripts.

Marsup:
Wow, it seems there's a LOT of work to do on linting, good luck !

PR #2555, opened 2015-05-22
fix dynamic scopes

nlf:


lib/auth.js
```diff
@@ -283,18 +283,23 @@ internals.Auth.prototype._authenticate = function (request, next) {
         // Check scope
 
         if (config.scope) {
+            var scopes;
```
hueniverse: assign config here and override later if needed

PR #2569, opened 2015-05-28
Make sure auth filter passes on request auth artifacts when injecting, too

anneveling:
Previous issue 2501/2502 turned out to do only part of what we would need to use auth artifacts in the application and test it to via server inject. Test added


AdriVanHoudt:
tests fail because of some linting errors:
```js
Linting results:
	test/auth.js:
		Line 1120: hapi/hapi-scope-start - Missing blank line at beginning of function.
		Line 1132: hapi/hapi-scope-start - Missing blank line at beginning of function.
```

PR #2570, opened 2015-05-28
Added "error" event listener

### LINKED ISSUE #2561: Cannot catch EADDRINUSE from server.start()
When I try to call `server.start(callback)`, and the port is in use, the script crashes with a message:

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: listen EADDRINUSE
    at errnoException (net.js:904:11)
    at Server._listen2 (net.js:1042:14)
    at listen (net.js:1064:10)
    at net.js:1146:9
    at dns.js:72:18
    at process._tickCallback (node.js:419:13)
```

I would expect the `callback` to be called instead (with the error, of course).

arb:
Added error listener to listener object. Closes #2561.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -149,17 +149,26 @@ internals.Connection.prototype._start = function (callback) {
         return process.nextTick(callback);
     }
 
-    this.listener.once('error', function (err) {
+    var self = this;
+    var onError = function (err) {
 
+        self._started = false;
         return callback(err);
-    });
+    };
+    this.listener.once('error', onError);
+
+    var finalize = function () {
+
+        self.listener.removeListener('error', onError);
+        callback();
+    };
 
     if (this.type !== 'tcp') {
-        this.listener.listen(this.settings.port, callback);
+        this.listener.listen(this.settings.port, finalize);
     }
     else {
         var address = this.settings.address || this.settings.host || '0.0.0.0';
-        this.listener.listen(this.settings.port, address, callback);
+        this.listener.listen(this.settings.port, address, finalize);
     }
 };
 
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -149,6 +149,11 @@ internals.Connection.prototype._start = function (callback) {
         return process.nextTick(callback);
     }
 
+    this.listener.once('error', function (err) {
+
+        return callback(err);
+    });
+
     if (this.type !== 'tcp') {
         this.listener.listen(this.settings.port, callback);
     }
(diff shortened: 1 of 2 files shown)


kanongil:
This approach can cause the callback to be triggered multiple times if an error is triggered on the listener for some other reason at a date after the `listening` event has been triggered.

You should probably also change `Connection#info.started` to `0`, to indicate that it stopped running.

arb:
@kanongil Yeah I knew :-( But I think I have a solution that isn't vulnerable to that. I also though about trying to call `_stop` on error but I'm not sure if we really need to do that.

Marsup:
Why don't you unregister that listener once the server is started ?

lib/connection.js
```diff
@@ -149,12 +149,26 @@ internals.Connection.prototype._start = function (callback) {
         return process.nextTick(callback);
     }
 
+    var self = this;
+    var startupErr = function (err) {
```
hueniverse: Event handler should be called `onError`

lib/connection.js
```diff
@@ -149,12 +149,26 @@ internals.Connection.prototype._start = function (callback) {
...
+
+        self._started = false;
+        return callback(err);
+    };
+    this.listener.once('error', startupErr);
+
+    var afterListen = function () {
```
hueniverse: This is usually called `finalize`

test/connection.js
```diff
@@ -399,6 +399,20 @@ describe('Connection', function () {
                 });
             });
         });
+
+        it('will return an error if the port is aleady in use', function (done) {
+
+            var server = new Hapi.Server();
+            server.connection({ port: 9000 });
```
hueniverse: test should first pass `0` (or nothing) and then reuse the port assigned.

lib/connection.js
```diff
@@ -149,12 +149,26 @@ internals.Connection.prototype._start = function (callback) {
         return process.nextTick(callback);
     }
 
+    var self = this;
```
hueniverse: `self` always at top of function with empty line.

lib/connection.js
```diff
@@ -149,12 +149,26 @@ internals.Connection.prototype._start = function (callback) {
...
+    var onError = function (err) {
+
+        self._started = false;
+        return callback(err);
+    };
+    this.listener.once('error', onError);
```
hueniverse: While at it, add another empty line above this :-)

PR #2574, opened 2015-05-31
server.dependency: call after with (server, options, next). Closes #2520

### LINKED ISSUE #2520: API wart: server.dependency after method lacks options
I find myself having to do this a fair bit:

```
function register(server, options, next) {
    server.dependency(['auth-plugin'], function (_server, _next) {
        after(_server, options, _next));
    }
}

function after(server, options, next) {
    // ... register routes, etc
}
```

Sure, I could use closures… but if the `after` API is designed around closures, why are we passing it `server`? If there's a reason to pass `server`, there's a reason to pass `options`. 

Do we have a bug category for "API changes waiting for enough friends to justify a major version bump"?

garthk:
This fixes #2520 without breaking any other tests, but I'm not entirely happy with it:

* The arguments list for the `Plugin` constructor is getting a little long
* ~~That check for `ext.arity` in `Server.start` doesn't explain itself~~
* ~~`Plugin.after` is getting too clever with its arguments~~

(Update: cleaned up a little in later commits, now squashed in.)

### DIFF
--- a/API.md
+++ b/API.md
@@ -841,8 +841,9 @@ Used within a plugin to declares a required dependency on other [plugins](#plugi
 - `after` - an optional function called after all the specified dependencies have been registered
   and before the server starts. The function is only called if the server is started. If a circular
   dependency is detected, an exception is thrown (e.g. two plugins each has an `after` function
-  to be called after the other). The function signature is `function(server, next)` where:
+  to be called after the other). The function signature is `function(server, options, next)` where:
     - `server` - the server the `dependency()` method was called on.
+    - `options` - the options object passed to the plugin during registration.
     - `next` - the callback function the method must call to return control over to the application
       and complete the registration process. The function signature is `function(err)` where:
         - `err` - internal error condition, which is returned back via the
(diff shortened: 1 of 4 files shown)


garthk:
Heh. Finding those lint errors was tough; I started with the `lint` branch, which has ~200 lines of lint complaints. Travis had a shorter list, but with line numbers based on `master`.

hueniverse:
Can't check the number of arguments in a callback function like this. It is never a good idea because you don't know how the method is implemented internally. It can have more than 3 arguments or no arguments at all and use `arguments`.

garthk:
Cleaned up the mess I'd made of _afters by pre-binding the `after` method's leading arguments. Removed checks against `arguments.length` as a side-effect. Cost: another private API method.

Any hints on how to rebase to `master`? I've been using Mercurial for half a decade.

garthk:
Rebased and squashed. Travis failing for `io.js` because `npm install` failed.

FWIW, I won't be at all offended if you ditch the PR because it doesn't pass muster. I'm delighted just to be able to *file* a PR.

PR #2575, opened 2015-05-31
Editor support improvements for Atom, Sublime, and others:

garthk:
`npm test` uses `lab` to enforce the code style. We can help contributors by also having their editor flag the problems as they type. This PR adds:

* Local `eslint` so lint plugins can report on divergence from Hapi JavaScript style
* `.editorconfig` so editorconfig plugins can apply Hapi indentation and line termination

I copied the `.eslintrc` from `lab`, and adjusted the `plugins` entry to make Atom happy.

Hazard: need to track `lab` changes to ~~`.eslintrc` and~~ its `package.json` dependencies on `eslint` and `eslint-plugin-hapi`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.js]
+charset = utf-8
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+[{package.json,.travis.yml,.eslintrc}]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,1 @@
+{ "extends": "./node_modules/lab/lib/linters/eslint/.eslintrc" }
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "accept": "1.x.x",
-    "ammo":  "1.x.x",
+    "ammo": "1.x.x",
     "boom": "^2.5.x",
     "call": "2.x.x",
     "catbox": "^4.2.x",
@@ -45,6 +45,8 @@
   "devDependencies": {
     "bluebird": "2.x.x",
     "code": "1.x.x",
+    "eslint": "0.19.x",
+    "eslint-plugin-hapi": "^1.0.0",
     "handlebars": "2.x.x",
     "lab": "5.x.x",
     "wreck": "5.x.x"


Marsup:
ESlint now offers the possibility to include another rc, I'd prefer that.

AdriVanHoudt:
Using the extends option from eslint would indeed be way better, maybe worth it to separate config in it's own module?

garthk:
I reckon [`extends`](http://eslint.org/docs/user-guide/configuring.html#extending-configuration-files) is a great idea, and the example in the docs is pushing the rules into their own module. Win!

Blocking problems:

* `lab` takes a dependency on `eslint@0.19.x`, which doesn't support `extends` (fix: hapijs/lab#366)
* `lab`'s version of `.eslintrc` needs its `plugins` prefix removed
* the Atom [`linter-eslint`](https://github.com/AtomLinter/linter-eslint) package doesn't seem to respect `extends`, even when using a local `eslint`

AdriVanHoudt:
PR should be done later today, needs a small eslint version bump.
The config should be revised anyway when making e separate module.
And I suggest creating an issue there, I've tried Atom recently and bumped
into the same issue so getting that fixed would be great.

On Sun, 31 May 2015 11:34 Garth Kidd <notifications@github.com> wrote:

> I reckon extends
> <http://eslint.org/docs/user-guide/configuring.html#extending-configuration-files>
> is a great idea, and the example in the docs is pushing the rules into
> their own module. Win!
>
> Blocking problems:
>
>    - lab takes a dependency on eslint@0.19.x, which doesn't support
>    extends (fix: hapijs/lab#366 <https://github.com/hapijs/lab/pull/366>)
>    - lab's version of .eslintrc needs its plugins prefix removed
>    - the Atom linter-eslint <https://github.com/AtomLinter/linter-eslint>
>    package doesn't seem to respect extends, even when using a local eslint
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hapijs/hapi/pull/2575#issuecomment-107149828>.
>
-- 

Met vriendelijke groeten

Adri Van Houdt

Marsup:
Yet another module, not sure of the benefit.

garthk:
`hapi` delegates lint rules to `lab`, so I think we already get a tick for "own module" for the time being. Can we make changing *which module* a separate issue, @AdriVanHoudt?

garthk:
I've revised the `.eslintrc` and `.editorconfig`. They'll work when:

- [ ] @AdriVanHoudt lands hapijs/lab#366, or `lab` otherwise upgrades its `eslint`
- [ ] someone removes the `eslint-plugin-` prefix from the `hapi` entry in `lab/lib/linters/eslint/.eslintrc`

Anything else we need?

AdriVanHoudt:
@Marsup there is https://github.com/cjihrig/eslint-plugin-hapi so that's already semi it's own module
@garthk sure
and removing the prefix doesn't break stuff?

AdriVanHoudt:
PR done, but needs revision

hueniverse:
I'm going to pass on this for now.

PR #2591, opened 2015-06-09
Some lint cleanup

csrl:
Here are two commits removing unused variables and parameters.

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -66,7 +66,7 @@ internals.prerequisites = function (request, callback) {
 internals.handler = function (request, callback) {
 
     var timer = new Hoek.Bench();
-    var finalize = function (response, data) {
+    var finalize = function (response) {
 
         if (response === null) {                            // reply.continue()
             response = Response.wrap(null, request);
@@ -251,7 +251,7 @@ internals.pre = function (pre) {
     return function (request, next) {
 
         var timer = new Hoek.Bench();
-        var finalize = function (response, data) {
+        var finalize = function (response) {
 
             if (response === null) {                            // reply.continue()
                 response = Response.wrap(null, request);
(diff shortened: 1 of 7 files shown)


lib/index.js
```diff
@@ -3,9 +3,4 @@
 var Server = require('./server');
 
 
-// Declare internals
```
AdriVanHoudt: this is per style guide so don't remove it

csrl: Can you point me to where that is called out in the style guide?  I don't see it.

AdriVanHoudt: http://hapijs.com/styleguide#user-content-module-globals

csrl: Ok, I think you are confused by that.  The style guide states that you "can only have two top level globals" referring to "internals" and "exports".  It does not say they are required to be there even if they are unused.  "can" vs "must".

And going by the lack of such in other modules that don't use them, I'd say it is fine to remove this one.

AdriVanHoudt: ok true the guide is somewhat lax, pinging @hueniverse for clarity

hueniverse: I always have them because I am so used to expecting them there that I never need to bother check if it is declared before using.

csrl:
Ok, I've pushed a new set of cleanup commits with that in mind.

lib/handler.js
```diff
@@ -66,7 +66,7 @@ internals.prerequisites = function (request, callback) {
 internals.handler = function (request, callback) {
 
     var timer = new Hoek.Bench();
-    var finalize = function (response, data) {
```
hueniverse: I rather keep `data` here because it's part of the reply interface callback spec even if it is not used.

lib/handler.js
```diff
@@ -251,7 +251,7 @@ internals.pre = function (pre) {
     return function (request, next) {
 
         var timer = new Hoek.Bench();
-        var finalize = function (response, data) {
+        var finalize = function (response) {
```
hueniverse: Same here.

hueniverse:
I'll just add this manually to another lint work ongoing.

PR #2593, opened 2015-06-10
Passing options to qs to remove inconsistent results.

kpjf:
The default of 20 on the arrayLimit creates inconsistent results, sometimes you get a regular zero indexed array, and if you have more than twenty items you get an associative array. Setting to 0 means that for all data passed in you get the same data type back.

This needs to be used in tandem with a pull request for subtext (https://github.com/hapijs/subtext/pull/7).


hueniverse:
This is not the right solution. The qs default is correct for security reasons. We need the option to override but not for everyone. Also, this PR does not include any tests or addresses the ability to pass qs options for query processing. I'll implement it differently.

PR #2602, opened 2015-06-16
Documentation typo fix. Closes #2601.

### LINKED ISSUE #2601: Error in API.md docs regarding plugin dependencies
The documentation on specifying dependencies for plugins is inconsistent:

In the `server.dependency()` section, the sample code says this:

```
exports.register = function (server, options, next) {

    return next();
};

register.attributes = {
    name: 'test',
    version: '1.0.0',
    dependency: 'yar'
};
```

However, the Plugins section says this:

> The plugin function must include an attributes function property with the following:
>
> `name` - required plugin name string. The name is used as a unique key. Published plugins should use the same name as the name field in the 'package.json' file. Names must be unique within each application.
> `version` - optional plugin version. The version is only used informatively to enable other plugins to find out the versions loaded. The version should be the same as the one specified in the plugin's 'package.json' file.
> `multiple` - if true, allows the plugin to be registered multiple times with the same server. Defaults to false.
> `dependencies` - optional string or array of string indicating a plugin dependency. Same as setting dependencies via server.dependency().

Further discussion on Gitter revealed that the correct key is `dependencies` and not `dependency` and referenced https://github.com/hapijs/hapi/blob/93292ce5a53e6c51077fadabb59831bb53c3688c/lib/plugin.js#L231

reedog117:
Addresses typo regarding plugin dependencies. Closes #2601.

### DIFF
--- a/API.md
+++ b/API.md
@@ -874,7 +874,7 @@ exports.register = function (server, options, next) {
 register.attributes = {
     name: 'test',
     version: '1.0.0',
-    dependency: 'yar'
+    dependencies: 'yar'
 };
 ```
 


devinivy:
Looks right to me.

PR #2621, opened 2015-06-29
propagate plugin.register errors via callback(err); resolves #2620

### LINKED ISSUE #2620: registering plugins can throw; API suggests callback(error) will be triggered instead
## sample code
```js
var Hapi = require('hapi');

server.connection({ port: 3000 });

const Foo = {
  register: function (server, options, next) {

    next();
  }
};

Foo.register.attributes = {
  name: 'foo',
  version: '1.0.0'
}

server.register([Foo, Foo], function(error) {

  console.log('register callback: ', error);
});
```

## Expectation:
Console will log:
```
register callback: Error: Plugin foo already registered ...
```

## Actuality:
```
........./node_modules/hapi/node_modules/hoek/lib/index.js:723
    throw new Error(msgs.join(' ') || 'Unknown error');
          ^
Error: Plugin foo already registered...
```

## Use case:
I'm building apps composed of plugins via Glue. Those plugins may (do) aregister other plugins. If/when the nested registration fails, for example due to already being registered [or any other error], an Error is thrown -- but it is not propagated back to my root Hapi application. The process dies without any output whatsoever and is quite difficult to debug. I confirmed the root cause of this by drilling down into the Hapi codebase and wrapping the registration code in the `plugins` file with a try/catch and confirming the error stems from there (again, specifically because a plugin has already been registered).

## Sugested solution:
Wrap the registration code in a `try/catch`, call `next(error)` rather than simply throwing the exception. This would enable me to 1) more easily debug my app by actually seeing the error, 2) potentially ignore some errors that aren't deemed dangerous -- for example, I may be ok with a second attempt to register a plugin failing as I only require that it was registered at least once.

## What on earth am I trying to do?
I'm attempting to combine what could be multiple isolated Hapi applications into one server that needs to be deployed/maintained. I could write a single Hapi server, require/register my one key plugin, and start the server. Repeat for each plugin/component. Instead, I'm attempting to load multiple such plugins and serve them all from one place. As mentioned, this can hurt at the point where each of your primary plugins may depend on another plugin -- for example, something that is parsing out the Authorization header and setting it to `request.app.foo`. And that is what led me to to create this issue :+1:

jmonster:
see #2620 for details; I'm open to suggestions..!

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -158,6 +158,17 @@ internals.Plugin.prototype._clone = function (connections, plugin) {
 
 internals.Plugin.prototype.register = function (plugins /*, [options], callback */) {
 
+    var callback = (typeof arguments[1] === 'object' ? arguments[2] : arguments[1]);
+
+    try {
+        this._register.apply(this, arguments);
+    } catch(err) {
+        callback(err);
+    }
+};
+
+internals.Plugin.prototype._register = function (plugins /*, [options], callback */) {
+
     var self = this;
 
     var options = (typeof arguments[1] === 'object' ? arguments[1] : {});


jmonster:
A big (huge!) downside to this PR is that when a error is thrown, other plugins won't be registered.. so while this enables you to *see* the error, it's not safe to ignore it and move forward -- you won't know which plugins were successfully registered vs which were not attempted [because of the throw]

PR #2642, opened 2015-07-12
Failing test for `server.ext` with complex deps.

devinivy:
Using `server.ext` option `before` as an array misbehaves, essentially ignoring the dependencies.

Root cause in topo has a pending PR: https://github.com/hapijs/topo/pull/7.  I've verified that patching topo allows all tests to pass.

### DIFF
--- a/test/plugin.js
+++ b/test/plugin.js
@@ -1823,7 +1823,7 @@ describe('Plugin', function () {
                             server.connections[2].inject('/', function (res3) {
 
                                 expect(res3.result.deps).to.equal('|3|2|');
-                                expect(res1.result.complexDeps).to.equal('|three|two|one|');
+                                expect(res3.result.complexDeps).to.equal('|three|two|one|');
                                 done();
                             });
                         });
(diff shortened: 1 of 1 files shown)


hueniverse:
Can you rewrite this test not to change existing tests? Just add a new `it()` block that doesn't touch all the other tests. At some point I want to break up all the common plugins used in there.

devinivy:
Should be resolved now!

test/plugin.js
```diff
@@ -1831,6 +1831,35 @@ describe('Plugin', function () {
...
+            server.connection({ labels: ['0'] });
+
+            var handler = function (request, reply) {
+
+                return reply(request.app.complexDeps);
+            };
+
+            server.route({ method: 'GET', path: '/', handler: handler });
+
+            server.register([internals.plugins.deps1, internals.plugins.deps2, internals.plugins.deps3], function (err) {
```
hueniverse: Don't use the internals plugin here. Create three new ones inside the test.

devinivy: Makes sense, given what you said.  I made those updates for you.

PR #2658, opened 2015-07-23
Regenerated changelog with --sanitize option on mdchangelog. Intended…

mtharrison:
… to fix broken markup on hapijs.com


mtharrison:
Related issue to add this option: https://github.com/creativelive/mdchangelog/issues/31

PR #2660, opened 2015-07-23
Allow objects as valid values for route notes

dmlittle:
Allow objects as valid values for route notes for more complex notes. Notes can now be a string, an array of strings, or an object with string values.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -249,7 +249,8 @@ internals.routeConfig = internals.routeBase.keys({
     description: Joi.string(),
     notes: [
         Joi.string(),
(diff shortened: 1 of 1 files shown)


dmlittle:
Bump

dmlittle:
Bump

Marsup:
You might want to explain the motivation behind this change.

hueniverse:
This is a breaking change and I don't see the value. If you need to store something other than a simple string, you can put it in the route config `app` object and use whatever format you want. The problem with this is that any system already using notes (e.g. lout) will now have to figure out what to do with this and how to present it.

PR #2662, opened 2015-07-25
Added schema validation when creating a server method using object

mtharrison:
This is designed to catch simple mistakes like this:

    server.method({
        name: 'myMethod',
        method: function () { ... },
        cache: {
            expiresIn: 1000
        }
    });

which should really be:

    server.method({
        name: 'myMethod',
        method: function () { ... },
        options: {
            cache: {
                expiresIn: 1000
            }
        }
    });

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -275,8 +275,8 @@ internals.method = Joi.object({
 
 
 internals.methodObject = Joi.object({
-    name: Joi.string(),
-    method: Joi.func(),
+    name: Joi.string().required(),
+    method: Joi.func().required(),
     options: Joi.object()
 });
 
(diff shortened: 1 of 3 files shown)


lib/schema.js
```diff
@@ -274,6 +274,13 @@ internals.method = Joi.object({
 });
 
 
+internals.methodObject = Joi.object({
+    name: Joi.string(),
```
Marsup: Shouldn't name and method be required ?

mtharrison: Yes. At the moment that's still taken care of here: 

https://github.com/hapijs/hapi/blob/master/lib/methods.js#L42-L43

But I can add it to this schema too to be more explicit about it.

AdriVanHoudt:
:+1:

netei:
Could be usefull indead

PR #2671, opened 2015-07-29
Update topo to 1.0.3

### LINKED ISSUE #2657: `server.ext` dependencies not ordered using `before` as array
Using `server.ext` option `before` as an array misbehaves, essentially ignoring the dependencies.  There's a PR with a failing test here: #2642.  The issue lives in topo, is directly related to https://github.com/hapijs/topo/issues/9.

devinivy:
Resolves #2657 and #2671.  See failing test at #2642.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -105,7 +105,7 @@
             }
         },
         "topo": {
-            "version": "1.0.2"
+            "version": "1.0.3"
         },
         "vision": {
             "version": "2.0.1"


AdriVanHoudt:
can you rename this to be like the issue, makes for a better changelog ;) http://hapijs.com/updates

mtharrison:
The changelog has picked it up from the issue. Is it normal practice to make both a PR and an issue for an update? We could get doubling up of plugin stuff in the changelog if so.

screen shot 2015-07-30 at 09 01 00

AdriVanHoudt:
yeah but it didn't do it for the others

Issue #2666, opened 2015-07-28
Support for plugin alias option


dylancwood:
# Background

For discoverability, plugins are often named quite verbosely, with the word *hapi* in them. Further, they are often hyphenated (e.g. `hapi-redis`). See https://github.com/sandfox/node-hapi-redis/issues/9 for a specific example.

# Problem

1. The hyphens used by most authors prevent developers from referencing the plugin with dot notation. They have to use bracket notation to reference the plugin (e.g. `server.plugins['hapi-redis']). IMO this adds up to wasted time, and ugly code. 
1. While it makes sense to prefix a plugin name with the the word *hapi* when interacting with it outside of hapi (e.g. `npm i hapi-redis`), it is redundant inside of hapi (e.g. `server.plugins['hapi-redis']`).

# Solution

Packages need to be discoverable, so prefixing them with *hapi-* is important. Further, it is important that the plugin name match the package name so that it is easy to understand where the plugin comes from. In order to maintain both of these requirements, and solve the above problem, I suggest that we support a user-defined `alias` option when registering a plugin. This idea was actually originally suggested by @Marsup on gitter when I asked for ideas about this problem. 

Here is how I envision the `alias` option being used:

```js
server.register([
    {
      register: require('hapi-redis'),
      options: {},
      alias: 'redis'
    },
    require('boom')
  ],
  { /*hapi-options*/ },
callbackFn
);
```

At first glance, I think that this could be accomplished in [lib/plugin.js](https://github.com/hapijs/hapi/blob/master/lib/plugin.js#L340) just after line 340:

```js
internals.Plugin.prototype.expose = function (key, value) {

    Hoek.assert(this.realm.plugin, 'Cannot call expose() outside of a plugin');

    var plugin = this.realm.plugin;
    this.root.plugins[plugin] = this.root.plugins[plugin] || {};
    // need to figure out how to grab alias :-(
    this.root.plugins[alias] = this.root.plugins[plugin];
    this.root.plugins[
    if (typeof key === 'string') {
        this.root.plugins[plugin][key] = value;
    }
    else {
        Hoek.merge(this.root.plugins[plugin], key);
    }
};
```

# Discussion

**What do you all think about adding a plugin alias for easier namespacing of plugins?**

hueniverse:
[...]

PR #2672, opened 2015-07-30
add plugin alias and tests

dylancwood:
# Background

See #2666 

# Solution

Add alias to `options` object passed to `server.register`.

# Discussion

The alias is passed to the `Plugin` constructor, and assigned to `this.alias`. From there, it is referenced in `expose()`. I am open to other suggestions on how to pass the **alias** to the `expose()` method. 

This option is only valid when registering one plugin at a time. If `options.alias` is specified when registering multiple plugins at once, and error is thrown.

I am currently not verifying that the alias does not overwrite an existing plugin's namespace in `server.plugins`. Is this something that is important? Since the alias is user-specified, it seems unlikely that they would unintentionally overload another existing plugin.

*Documentation updates still to come.* I wanted to iterate on this PR before working on the documentation.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -37,6 +37,7 @@ exports = module.exports = internals.Plugin = function (server, connections, env
     this.plugins = this.root._plugins;
     this.settings = this.root._settings;
     this.version = Package.version;
+    this.alias = options.alias || null;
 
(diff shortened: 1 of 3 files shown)


hueniverse:
Reverted for now.

Issue #2673, opened 2015-08-03
304 response sends: no-cache regardless of routes.cache settings


atomantic:
configuring 
```
routes.cache = {
       expiresIn: 900000,
       privacy: "public"
}
```
is working great for 200 responses, but when the if-non-match or if-modified-since matches and converts the response to a 304, the cache-control header becomes "no-cache". Normally, this is OK, as a browser will use the existing cached asset instead of updating the cache-control with the 304 cache-control header, but this makes a reverse proxy like Akamai unhapi.

This RFC suggests a 304 should always return the same cache-control that a 200 would return: https://tools.ietf.org/html/rfc7232#section-4.1
```
The server generating a 304 response MUST generate any of the following 
header fields that would have been sent in a 200 (OK) response 
to the same request: 
Cache-Control, Content-Location, Date, ETag, Expires, and Vary.
```

kanongil:
@atomantic See https://github.com/hapijs/hapi/issues/2676#issuecomment-128298036 for a work-around.

@hueniverse I am looking into fixing this. I have created a failing test case but I am not sure about the right way to implement the actual fix.

I am leaning towards amending the default `cache.statuses` option with the 304 header. However, this does not fix the issue for anyone who is already using the option (with 200 and no 304).

The other option is to implicitly add the 304 response as cacheable whenever 200 is included in the `cache.statuses` list. This is simple to do while converting the list to an object for the lookup.

hueniverse:
It needs to be explicit in the logic, not the config.

PR #2703, opened 2015-08-12
Question: Don't add must-revalidate cache-control value.

wprl:
This may be more of a question than a solution.  

Chrome audit reports any response with `Cache-Control: must-revalidate` as an error that the resource is "explicitly non-cacheable."  

Chrome audit seems to contradict [the RFC](https://tools.ietf.org/html/rfc7234#section-5.2.2.1) here, but perhaps `must-revalidate` is problematic in practice, even if it shouldn't be problematic in theory.

Removing `must-revalidate` fixes the Chrome audit results for me.  Could it be removed, or perhaps be optional?  Is Chrome audit reporting a false error?

### DIFF
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -425,7 +425,7 @@ internals.cache = function (response) {
 
         var ttl = (response.settings.ttl !== null ? response.settings.ttl : request._route._cache.ttl());
         var privacy = (request.auth.isAuthenticated || response.headers['set-cookie'] ? 'private' : request.route.settings.cache.privacy || 'default');
-        response._header('cache-control', 'max-age=' + Math.floor(ttl / 1000) + ', must-revalidate' + (privacy !== 'default' ? ', ' + privacy : ''));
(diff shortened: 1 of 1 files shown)


hueniverse:
@mnot can you help answer this?

kanongil:
This is likely another manifestation of the bug in #2673. Chrome is probably receiving a 304 response with incorrect `cache-control: no-cache`.

wprl:
I enabled sending `Cache-Control` for 304s, so I think I already addressed the issue with 304s in the code.  My issue is that this causes Chrome audit errors:

```
Cache-Control: max-age=36000, private, must-revalidate
```

Removing the `must-revalidate` clears the "non-cacheable" error from the audit:

```
Cache-Control: max-age:36000, private
```

mnot:
Based upon a quick look, Chrome Audit is saying a few different misleading things about caching.

See:
  https://code.google.com/p/chromium/issues/detail?id=520353

wprl:
Looks like Paul Irish created an issue for it: https://codereview.chromium.org/1294473004

I think I can safely assume it's a bug in the audit tool.

PR #2712, opened 2015-08-13
update all failAction references to with allowed value of callback

camerow:
Docs now reflect failAction takes an optional callback parameter.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2106,6 +2106,7 @@ following options:
         - `'error'` - return a Bad Request (400) error response. This is the default value.
         - `'log'` - report the error but continue processing the request.
         - `'ignore'` - take no action and continue processing the request.
+        - `function` - takes a callback with optional params '(request, reply, source, error)'.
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -2106,6 +2106,7 @@ following options:
         - `'error'` - return a Bad Request (400) error response. This is the default value.
         - `'log'` - report the error but continue processing the request.
         - `'ignore'` - take no action and continue processing the request.
+        - `function` - takes a callback with optional params '(request, reply, source, error)'.
```
hueniverse: I wouldn't call is optional params.

- a function with signature `function(request, reply, source, err)` where:
   - `request` - ...

and so on. See other function options style in the doc.

camerow: Ok. I see what you mean. Like I said, definitely new to this so some of the phrasing and subtle details often escape me.

camerow:
I believe that was the change you needed. If not let me know. This is my first official public repo contribution. Even though it was incredibly small, it's got me excited. =) Thanks.

lloydbenson:
Welcome to the community!

Issue #2566, opened 2015-05-27
Lifecycle hooks on routes


Marsup:
What would you think of allowing the lifecycle hooks (all but onRequest obviously) to be set per route ?
The filtering I have to do to match routes in those functions just pollutes the readability of it.
Having global hooks when they're not used as such probably also is a tiny waste of resource however minor.

hueniverse:
I've long wanted to change the ext functionality to be per:
- server
- connection
- plugin
- route

But I am worried about the overhead of checking for all those. Right now when nothing is set it is a simple null check at each ext point. For this to work I will need to change the lifecycle array generation to actually build the array per route as well as rebuild it when extensions are added later after route config.

Maybe for 9.0.

devinivy:
This would be wonderful.  @hueniverse are you worried about overhead at the time of server start, or throughout the course of a single request?

hueniverse:
Only care about hot code. Executing the request lifecycle is hot.

seanrucker:
Looking forward to this feature; here's my use case...

I'm using Hapi as an API Gateway. I have four clients accessing this API, each has its own set of routes. One of the clients is an Ember app which requires the API response to conform to the [JSON API spec] (http://jsonapi.org). I'm using a Hapi plugin (inspired by [hapi-json-api.js](https://gist.github.com/wraithgar/a3ec2e0a4295d5d784c9)) to handle some of the boilerplate code required to munge the response into something JSON API compliant. The plugin uses Hapi lifecycle hoks which unfortunately has the effect of making _all_ my routes conform to JSON API.

hueniverse:
What would the interface look like for adding extensions to one or more routes? I got ideas but want to see what other people think.

devinivy:
I see three options,
1. Add an `ext

devinivy:
I see three options, some of which could probably work in tandem.

1. Add an `ext()` method to the route public interface.
    ```js
    server.lookup('route-id').ext('onPreAuth', extensionFn, options);
    ```

2. Add an ext option to route config.  Adding options would start to look messy.
    ```js
    server.route({
        method: 'GET',
        path: '/',
        config: {
            id: 'route-id',
            handler: function (request, reply) { return reply('ok'); },
            ext: { onPreAuth: [extensionFn] }
        }
    });
    ```

3. Adding a `routes` option to `server.ext()`.  This addresses `ext()` options and the ability to add it to multiple routes.
    ```js
[...]

PR #2726, opened 2015-08-19
Route-level extension management.

devinivy:
This is part of the work for #2566.  Wanted to get some feedback, @hueniverse, before moving forward if you'd like me to.  This is meant to be a refactor of how lifecycle hooks are managed so that routes and connections each have their own list of dependency-managed hooks.

The approach is to let each connection keep track of route-level extensions in addition to their own `onRequest` extensions.  A route will pull-in its connection's route-level extensions when it's added.  When a new extension is added to a connection, it's also added to each route's (and special route's) list of extensions.

No tests have been added yet.  The one failing test (`188) Methods allows genreateTimeout to be false`) is failing on the current build.  There should be no API changes in this PR.  If this looks good it can be used to add lifecycle hooks to route configurations.

**Edit** Looks like some PRs have been merged in time to see no failing tests.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -347,7 +347,7 @@ internals.Connection.prototype._ext = function (event, nodes, options) {
         // Add extension to existing routes
         var routes = this._router.table();
         for (var i = 0, il = routes.length; i < il; ++i) {
-            routes[i]._extNodes(event, nodes, options);
+            routes[i]._ext(event, nodes, options);
         }
 
         // And to special routes (notFound, etc.)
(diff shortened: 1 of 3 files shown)


hueniverse:
Thanks for the effort, but this is the kind of refactor that I am unlikely to take from a PR. It has to be something I do from scratch - just how my process works. It looks reasonable (basically keep a route level list and init it with whatever was already added to the connection pre-route).

hueniverse:
In general, if I don't mark an issue with `help wanted` or `new contributors`, I am not ready to take pull requests on it...

devinivy:
No worries– just wanted to dig into some unfamiliar territory.

Issue #2705, opened 2015-08-12
Start server without callback fails


felansu:
When starts the server without callback as it described at example on [home page ](http://hapijs.com/)

**Code example**

```javascript
var Hapi = require('hapi');

// Create a server with a host and port
var server = new Hapi.Server();
server.connection({ 
    host: 'localhost', 
    port: 8000 
});

// Add the route
server.route({
    method: 'GET',
    path:'/hello', 
    handler: function (request, reply) {
       reply('hello world');
    }
});

// Start the server
server.start();
```

**Exception**
```
/home/felansu/Projects/QRBWEB/node_modules/hapi/lib/server.js:242
        return callback();
               ^
TypeError: undefined is not a function
    at /home/felansu/Projects/QRBWEB/node_modules/hapi/lib/server.js:242:16
    at /home/felansu/Projects/QRBWEB/node_modules/hapi/node_modules/hoek/lib/index.js:841:22
    at process._tickDomainCallback (node.js:381:11)
```

**Version report**
```
└─┬ hapi@9.0.1
  ├── accept@1.1.0
  ├── ammo@1.0.1
  ├── boom@2.8.0
  ├── call@2.0.2
  ├── catbox@6.0.0
  ├── catbox-memory@1.1.1
  ├── cryptiles@2.0.4
  ├─┬ heavy@3.0.0
  │ └─┬ joi@5.1.0
  │   ├── isemail@1.1.1
  │   └── moment@2.10.6
  ├── hoek@2.14.0
  ├── iron@2.1.3
  ├── items@1.1.0
  ├─┬ joi@6.6.1
  │ ├── isemail@1.1.1
  │ └── moment@2.10.6
  ├── kilt@1.1.1
  ├─┬ mimos@2.0.2
  │ └── mime-db@1.16.0
  ├── peekaboo@1.0.0
  ├── qs@4.0.0
  ├── shot@1.6.0
  ├── statehood@2.1.1
  ├─┬ subtext@2.0.0
  │ ├── content@1.0.2
  │ ├─┬ pez@1.0.0
  │ │ ├── b64@2.0.0
  │ │ └─┬ nigel@1.0.1
  │ │   └── vise@1.0.0
  │ └── wreck@6.1.0
  └── topo@1.0.3
```

cleydsonjr:
+1

mtharrison:
This is a new requirement in hapi 9. See release notes here: https://github.com/hapijs/hapi/issues/2682

[...]

PR #2739, opened 2015-08-26
Updated error when calling server.start with no callback

fiveisprime:
Updated the start method of server to verify that a callback was
passed. If no callback is passed, a friendly error is thrown.

    Error: A callback is required when calling server.start

This is compared to the old error which required some digging
to find the cause of the problem.

    TypeError: callback is not a function

Related to #2705

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -127,6 +127,8 @@ internals.Server.prototype.start = function (callback) {
 
     var self = this;
 
+    Hoek.assert(typeof callback === 'function', 'Missing required start callback function');
+
     if (this._state === 'initialized') {
         return this._start(callback);
     }
--- a/test/server.js
+++ b/test/server.js
@@ -233,6 +233,17 @@ describe('Server', function () {
             }).to.throw('Cannot start server before plugins finished registration');
             done();
         });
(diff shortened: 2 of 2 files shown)


kanongil:
Nice. You should probably align the test and message with the stop() test here: https://github.com/hapijs/hapi/blob/master/lib/server.js#L258.

fiveisprime:
Updated! Thanks for the feedback. :+1:

arb:
Can you add a test for this as well?

fiveisprime:
Done.

spiralni:
@fiveisprime thanks for your contribution!

It's very nice to have descriptive exceptions. I like the way "windsor" library shows its errors, very verbose and including way to fix them.
Maybe we can include a little example showing how it must to be done. It is just an idea.

This is an example: 

exception_thumb2

Source: http://kozmic.net/2011/02/16/windsors-longest-exception-message-just-got-longer/

fiveisprime:
@spiralni That's awesome! I can do something like this, but in a different pull
request since it will touch a lot more than just the start.

I love this idea. :)

PR #2741, opened 2015-08-27
Fix incorrect docs for Route Prerequisites custom failAction

wookiehangover:
I was really bummed to discover that despite the documentation, providing a custom `failAction` function is not supported in Route Prerequisites.

This removes the documentation error, which looks to be a copypasta mistake in 98a3f02cf03f1c02633ea3ba779d2878f18d217d

BTW, this would be a great feature to have, are there any plans to implement it / is anybody working on it?


hueniverse:
No one has ever asked for it and in practice, this is why we have all the other extension points.

wookiehangover:
Fair enough, thanks @hueniverse!

Issue #2746, opened 2015-08-29
"failAction" does not accept custom error function as stated in the documentation


maddin77:
According to [Route options](http://hapijs.com/api#route-options) in the documentation, `payload.failAction`, `response.failAction`, `state.failAction`, `validate.failAction` and `pre.failAction` all accept a custom error handler function besides predefined string values.

However, as defined in [schema.js](https://github.com/hapijs/hapi/blob/master/lib/schema.js), `validate.failAction` is the only one that actually accepts a function as a valid value. Therefore trying to use a custom function as `failAction` for any of the others will throw
```
[1] "failAction" must be a string
```
on startup.

There's #2741 to remove the part of the documentation for the request prerequisites, so i'm wondering if it's supposed to be like in the documentation, or as defined in the schema.

hueniverse:
As you can tell from #2741 being merged, it's not supported.

PR #2748, opened 2015-09-01
Fix mis-documentation of various failAction options

devinivy:
As pointed out in #2746 there are various places in the docs that indicate a `failAction` callback can be provided.  Turns out that's not quite the case!  It can only be used with route validation options.  This PR removes it from cookie parsing, payload parsing, and response validation sections of the docs (#2741 fixed for route prerequisites).  Wouldn't mind if someone made double sure that these options can go, but as far as I can tell, the tests, schema, and codebase indicate that they can.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2106,13 +2106,6 @@ following options:
         - `'error'` - return a Bad Request (400) error response. This is the default value.
         - `'log'` - report the error but continue processing the request.
         - `'ignore'` - take no action and continue processing the request.
-        - a custom error handler function with the signature `function(request, reply, source, error)` where:
-            - `request` - the [request object](#request-object).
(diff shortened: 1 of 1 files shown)


devinivy:
For anyone reviewing, here's where I'm looking to make sure `failAction` isn't being used as a callback.

- Cookie parsing: https://github.com/hapijs/hapi/blob/e72b46e4f8342bd784201b72242c70d4ac106034/lib/route.js#L349-L358
- Payload parsing: https://github.com/hapijs/hapi/blob/e72b46e4f8342bd784201b72242c70d4ac106034/lib/route.js#L380-L387
- Response validation: https://github.com/hapijs/hapi/blob/93292ce5a53e6c51077fadabb59831bb53c3688c/lib/validation.js#L187-L194

PR #2762, opened 2015-09-09
Fixed broken windows tests

simon-p-r:
When testing hapi on windows I found 2 broken tests, this PR is to fix this one https://github.com/hapijs/hapi/blob/master/test/connection.js#L1089-L1118 whilst the I cannot fix the other test here https://github.com/hapijs/hapi/blob/master/test/connection.js#L64-L73 as this is beyond me at the moment. 

Thanks

### DIFF
--- a/test/connection.js
+++ b/test/connection.js
@@ -1111,9 +1111,7 @@ describe('Connection', function () {
 
                 server.inject('/', function (res) {
 
-                    var EOL = Os.EOL;
-                    var html = '<div>' + EOL + '    <h1>hola!</h1>' + EOL + '</div>' + EOL;
-                    expect(res.result).to.equal(html);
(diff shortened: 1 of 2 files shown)


test/connection.js
```diff
@@ -1111,7 +1111,9 @@ describe('Connection', function () {
 
                 server.inject('/', function (res) {
 
-                    expect(res.result).to.equal('<div>\n    <h1>hola!</h1>\n</div>\n');
```
hueniverse: Better to use a regular expression (e.g. `expect(res.result).to.match(...)`

simon-p-r:
Like this?

hueniverse:
No, keep the string but where it has `\n` adjust to also allow Windows new lines.

PR #2773, opened 2015-09-18
Detects and rejects malformed response headers

jefflembeck:
This works around a byte truncation flaw in node core versions > 0.10
where the high bit is stripped from headers leading to a possible header
injection.


aredridel:
This has security implications for folks passing arbitrary data to redirects in any fashion, even within otherwise legit URLs.

PR #2775, opened 2015-09-20
Add next to docs example for server.after

toboid:
Tiny update to docs to show correct usage of `next` in `server.after` callback so that correct pieces are in place for anyone copying the example.

### DIFF
--- a/API.md
+++ b/API.md
@@ -475,9 +475,11 @@ var Hapi = require('hapi');
 var server = new Hapi.Server();
 server.connection({ port: 80 });
 
-server.after(function () {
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -475,9 +475,11 @@ var Hapi = require('hapi');
 var server = new Hapi.Server();
 server.connection({ port: 80 });
 
-server.after(function () {
+server.after(function (server, next) {
```
hueniverse: Change `server` to `srv` since they are not the same.

toboid: Sure, this now updated.

PR #2776, opened 2015-09-20
Add preload flag to HSTS header and fix casing for includeSubDomains.

analytically:
See https://scotthelme.co.uk/hsts-preloading/ for more information.

Also fixes casing issue in HSTS header: includeSubDomains - http://tools.ietf.org/html/rfc6797.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2161,7 +2161,7 @@ following options:
       be set to that number. Defaults to `true`. You may also specify an object with the
       following fields:
         - `maxAge` - the max-age portion of the header, as a number. Default is `15768000`.
-        - `includeSubdomains` - a boolean specifying whether to add the `includeSubdomains`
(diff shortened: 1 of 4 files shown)


devinivy:
I can't say for sure whether or not this will be an accepted feature, but the PR will likely need to adjust the casing of `includeSubDomains` vs `includeSubdomains` in the tests and include a test for the new option to be considered.  Additionally, the options schema would need to be adjusted around here: https://github.com/hapijs/hapi/blob/master/lib/schema.js#L116.  It does seem like the casing of `includeSubDomains` in this PR is preferred, based upon other examples on the web.

hueniverse:
@nlf?

nlf:
this gets a +1 from me, looks good

devinivy:
Is this our first case of releasing a feature to 10.x and hapi-tls?

PR #2790, opened 2015-09-27
Add cache stats to server methods

### LINKED ISSUE #2763: get cache stats for a server method?
It looks like catbox stores [stats](https://github.com/hapijs/catbox#api-1) for gets/sets/hits/stales/etc.  Is there any way to access this for a server method?  From the looks of it, drop is the only thing exposed to the server method.

toboid:
Adds catbox stats to server methods cache object. Closes #2763

### DIFF
--- a/API.md
+++ b/API.md
@@ -1212,12 +1212,14 @@ to use the built-in cache and share across multiple request handlers without hav
 common module.
 
 Methods are registered via `server.method(name, method, [options])` where:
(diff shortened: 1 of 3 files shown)


hueniverse:
Missing documentation.

toboid:
Documentation added. Wasn't completely sure how to document this so any guidance would be appreciated if this isn't the best approach. Thanks.

PR #2795, opened 2015-09-29
Added documentation for the Request.generateResponse method

vertiman:


API.md
```diff
@@ -2600,6 +2600,28 @@ server.ext('onRequest', function (request, reply) {
...
+    if (!thing) {
+      return request.generateResponse(Boom.notFound());
+    }
+
+    return thing;
+  });
+  return reply(result);
+};
+```
+
```
hueniverse: Example should comply with the style guide.

API.md
```diff
@@ -2600,6 +2600,28 @@ server.ext('onRequest', function (request, reply) {
 });
 ```
 
+#### `request.generateResponse(source, [options])`
+
+_Always available._
+
+Returns an [`response`](#response-object) which you can pass into the [reply interface](#reply-interface).  Useful inside a promise to create a response object to return up to the [reply interface](#reply-interface).
```
hueniverse: s/Useful/For example

vertiman:
@hueniverse changes done.  Thanks for the feedback.

API.md
```diff
@@ -2600,6 +2600,29 @@ server.ext('onRequest', function (request, reply) {
...
+- `source` - the object to set as the source of the [reply interface](#reply-interface).
+- `options` - options for the method, optional.
+
+```js
+var handler = function (request, reply) {
+
+    var result = promiseMethod().then(function (thing) {
+    
+        if (!thing) {
+            return request.generateResponse(Boom.notFound());
```
hueniverse: This doesn't explain why you would need it because you can return an error as-is. You need this method if the promise wants to return a reply with some options like status code and headers.

API.md
```diff
@@ -2600,6 +2600,29 @@ server.ext('onRequest', function (request, reply) {
 });
 ```
 
+#### `request.generateResponse(source, [options])`
+
+_Always available._
+
+Returns an [`response`](#response-object) which you can pass into the [reply interface](#reply-interface).  For example inside a promise to create a response object to return up to the [reply interface](#reply-interface).
```
hueniverse: Returns **a**...

Move the "For example" to after the arguments before the example and end the first sentence with "where:" like we do elsewhere int he document. Also no empty line before the arguments.

PR #2796, opened 2015-09-29
don't duplicate accept-encoding in vary header

ldesplat:
This is in relation hapijs/h2o2#17 .

accept-encoding value in vary header was being duplicated. This fixes the issue and adds a test case to cover the scenario.

I could have fixed it instead in [lib/response.js](https://github.com/hapijs/hapi/blob/master/lib/response.js#L129), to not add a value if it already exists in a header but I thought that was too brittle to do there.

PS: forgot to say, I will make a lts PR if this is accepted.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -143,6 +143,15 @@ internals.Response.prototype.vary = function (value) {
         this.headers.vary = value;
     }
     else if (this.headers.vary !== '*') {
+
+        var values = this.headers.vary.split(',');
+
+        for (var i = 0, il = values.length; i < il; ++i) {
+            if (values[i] === value) {
+                return this;
+            }
+        }
+
         this._header('vary', value, { append: true });
     }
 
--- a/test/response.js
+++ b/test/response.js
@@ -286,6 +286,26 @@ describe('Response', function () {
                 done();
             });
         });
+
+        it('sets Vary header with multiple similar and identical values', function (done) {
+
+            var handler = function (request, reply) {
+
+                return reply('ok').vary('x').vary('xyz').vary('xy').vary('x');
+            };
+
+            var server = new Hapi.Server();
+            server.connection();
+            server.route({ method: 'GET', path: '/', handler: handler });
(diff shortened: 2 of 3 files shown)


lib/transmit.js
```diff
@@ -227,7 +227,7 @@ internals.transmit = function (response, callback) {
 
     // Content-Encoding
 
-    if (request.headers['accept-encoding']) {
+    if (request.headers['accept-encoding'] && (!response.headers.vary || response.headers.vary.indexOf('accept-encoding') === -1) ) {
         response.vary('accept-encoding');
```
hueniverse: Better done inside he response.vary()  method in a generic way.

ldesplat:
You are right and much easier. Updated the PR. Added 1 more test and kept the old one since I thought it was still relevant.

lib/response.js
```diff
@@ -143,7 +143,20 @@ internals.Response.prototype.vary = function (value) {
...
     }
     else if (this.headers.vary !== '*') {
-        this._header('vary', value, { append: true });
+
+        var values = this.headers.vary.split(',');
+        var foundMatch = false;
+
+        for (var i = 0, il = values.length; i < il; ++i) {
+            if (values[i] === value) {
+                foundMatch = true;
```
hueniverse: Why not just return here?

ldesplat: I would have but there appears to be opportunities to return early in this function and it was not done. So, I was following the current flow and decided to `break`. Returning early does save an allocation and I am already doing `split` so I just changed it. Thanks!

ldesplat:
Any update on this? I would love to get it going so I can do a PR for lts and get h2o2 updated! Thank you :)

ldesplat:
@nlf could this be included in LTS branch? h2o2 tests fail when upgrading to hapi 9.x.x otherwize.

nlf:
it will be. i'll release a new LTS after 10.2.0 is released

PR #2800, opened 2015-09-30
Fixing server.views reference in Error Response example

ryanmurakami:
An Error Response example showing how to render a view doesn't register vision before calling server.views. Also, inert is unused in this example, since there are no static files being returned.


ryanmurakami:
Saw this was already opened :-/

PR #2829, opened 2015-10-06
Expanded `registrations` API. Added test. Updated docs.

arb:
Added more information to the plugin results of `registrations`.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -253,6 +253,14 @@ internals.Plugin.prototype.register = function (plugins /*, [options], callback 
         selection.realm.modifiers.route.vhost = item.options.routes.vhost;
         selection.realm.pluginOptions = item.pluginOptions || {};
 
+        var registrationData = {
+            version: item.version,
+            name: item.name,
+            multiple: item.multiple,
+            options: item.pluginOptions,
+            attributes: item.register.attributes
+        };
+
         // Protect against multiple registrations
 
         if (!item.connections) {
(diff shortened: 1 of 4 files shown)


devinivy:
Any reason the `'0.0.0'` default version is necessary?

arb:
Yes the `0.0.0` default is necessary. If a module doesn't have an `attributes.pkg` object or a `attributes.version`, then the plugin version ends up being `undefined`. Based on the tests and the `default` values in the plugin schema, I think the intent was to set the version to 0.0.0 in any instance where the version is not provided.

hueniverse:
Hmm. The schema has the '0.0.0' default. Is it not working right?

lib/plugin.js
```diff
@@ -264,7 +272,7 @@ internals.Plugin.prototype.register = function (plugins /*, [options], callback
                 Hoek.assert(item.multiple, 'Plugin', item.name, 'already registered');
             }
             else {
-                self.root._registrations[item.name] = { version: item.version };
+                self.root._registrations[item.name] = registrationData;
```
AdriVanHoudt: why not just `item` as a whole?

arb:
@hueniverse it works right if your plugin has `register.attributes.pkg = {}`, the the Joi default on "version" pops and defaults it to "0.0.0". In instances where you don't have anything on `pkg`, only the default value of `{}` is set to `pkg`; not on `version`.

hueniverse:
The we need to fix the schema, not the code.

arb:
@hueniverse updated schema so we could remove the `|| '0.0.0'` code change.

lib/plugin.js
```diff
@@ -253,6 +253,14 @@ internals.Plugin.prototype.register = function (plugins /*, [options], callback
         selection.realm.modifiers.route.vhost = item.options.routes.vhost;
         selection.realm.pluginOptions = item.pluginOptions || {};
 
+        var registrationData = {
+            version: item.version,
+            name: item.name,
+            multiple: item.multiple,
```
hueniverse: This is redundant given you have attributes.

arb:
This should be good for a second review.

PR #2865, opened 2015-10-21
Expose original handler options as a property of the handler function

csabapalfi:
When handler options is an Object we just turn it to a function and there's no way to access the original handler options.

We could simply add the original handler options object as a property of the handler function and this is what this PR does.

(My use case is to provide access to proxy options to my plugin which sets up mock replies using nock for local integration testing)


csabapalfi:
Actually I no longer need this and even when I did I could work around it. Closing now.

PR #2870, opened 2015-10-23
Fix empty content-length handling for gzip and 204 responses

kanongil:
The transmit logic incorrectly considers missing `content-length` headers as having a payload length of `0`, while it usually signals that chunked transfer encoding should be used (with any actual length).

This patch fixes this by always parsing the header value, and testing using integer comparisons instead of the current truthiness tests.

I have added 2 new tests to verify the chunked response behavior. I also fixed an existing compression test, which was actually not working as intended due to not signalling a compressible mime type.


kanongil:
Note that this bug actually causes incorrect responses when chunked transfers are used along with `response.emptyStatusCode === 204`.

PR #2885, opened 2015-10-29
Update hapijs/subtext to 2.0.2 from 2.0.1

gergoerdosi:


gergoerdosi:
@hueniverse Any chance to merge and release this as a patch version? This issue is a blocker for us, we can't do deployments because npm commands fail with an error.

PR #2887, opened 2015-10-30
Test on node v5

gergoerdosi:


.travis.yml
```diff
@@ -3,5 +3,7 @@ language: node_js
 node_js:
   - 4.0
   - 4
+  - 5.0
+  - 5
```
hueniverse: Just 5. we explicitly support 4.0 but no fixed 5 version

AdriVanHoudt: does hapi 'officially' support 5?

Marsup: We tested for 0.12 and yet never supported it.

AdriVanHoudt: so that is a no?

gergoerdosi: Removed 5.0. I don't see why we shouldn't support 5 (and other non-lts releases). We didn't support 0.12 because it had a lot of issues. But it's not the case with 5 (hopefully).

jzorn:
I think what @Marsup says is that they were testing v0.12 even though they did not support it and therefore no reason exists not to test on v5.

hueniverse:
We should test on 5 as a service to the node community so that we can surface issues early.

Issue #2894, opened 2015-11-02
CORS  headers to include 'Origin'


nettaru:
Safari is sending 'Origin' in its headers, and therefor is denied by hapi, which makes the OPTIONS request to fail and return 404:
![screen shot 2015-11-02 at 15 46 10](https://cloud.githubusercontent.com/assets/1736657/10884706/c6a2a7a8-8179-11e5-9fe7-def0d38f6ad4.png)

need to add 'Origin' to the allowed headers to fix the issue:
https://github.com/hapijs/hapi/blob/master/lib/defaults.js#L86

mtharrison:
You're making a cross origin request there. Have you enabled CORS on your /token route?

nettaru:
yes. it's set to true on all the routes in the server configuration like so
```
server: {
    port: 3000
    router: {
      stripTrailingSlash: true
    }
    routes: {
      cors: true
    }
  }
```

mtharrison:
Is that a Glue manifest? Are you using the latest versions of everything? The `routes` option is supposed to sit under a `connections` object:

```javascript
var manifest = {
    server: {
        connections: {
            routes: {
                cors: true
            }
        }
    },
    connections: [
        ...
    ],
    plugins: [
        ...   
    ]
};
```

nettaru:
not using Glue. The object above is my serverConfig object, and i'm using it like so in my app.js: `server.connection(serverConfig)`.

I'm using hapi 11.0.2 because I'm getting the following error for 11.0.3:
![screen shot 2015-11-02 at 16 37 48](https://cloud.githubusercontent.com/assets/1736657/10885818/79e254f8-817f-11e5-8997-541b8f9bbe24.png)

devinivy:
hapi v10 and v11 are meant to be used with node v4, which is probably the issue you've run into above!  Up until 11.0.3 hapi worked on node v0.10 essentially by coincidence.

I'm curious if you're able to verify your cors issue using curl or a similar client, by checking to see that the origin header is the only difference for you between a working cors request and a failing cors request.  Does adding `'origin'` to `cors.additionalHeaders` fix your issue?

darsee:
> Does adding 'origin' to cors.additionalHeaders fix your issue?

It seems to fix the issue for me.

nettaru:
@devinivy thanks, updating node did help the terminal issue.
As for the second question - it does fix the issue. when i do the same request from chrome there is no problem since the header is different (includes only access and content-type):
[...]

PR #2895, opened 2015-11-02
add Origin to CORS header

nettaru:
solves #2894 

Add 'Origin' to CORS headers for Safari requests.
Removed duplicate test.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2127,7 +2127,7 @@ following options:
       ('Access-Control-Max-Age'). The greater the value, the longer it will take before the
       browser checks for changes in policy. Defaults to `86400` (one day).
     - `headers` - a strings array of allowed headers ('Access-Control-Allow-Headers').
-      Defaults to `['Accept', 'Authorization', 'Content-Type', 'If-None-Match']`.
+      Defaults to `['Accept', 'Authorization', 'Content-Type', 'If-None-Match', 'Origin']`.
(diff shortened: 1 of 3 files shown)


API.md
```diff
@@ -2127,7 +2127,7 @@ following options:
       ('Access-Control-Max-Age'). The greater the value, the longer it will take before the
       browser checks for changes in policy. Defaults to `86400` (one day).
     - `headers` - a strings array of allowed headers ('Access-Control-Allow-Headers').
-      Defaults to `['Accept', 'Authorization', 'Content-Type', 'If-None-Match']`.
+      Defaults to `['Accept', 'Authorization', 'Content-Type', 'Origin', 'If-None-Match']`.
```
AdriVanHoudt: make then alphabetically sorted, also below

kanongil:
While this fixes the issue, I'm not sure this is the best way to handle it. I would prefer that hapi just adds the `origin` header as an implied allowed header for preflight requests, not changing the `access-control-allow-headers` output. This should be simple to do by adding it to the `settings._headers` list, if missing.

This would also fix any code where `cors.headers` has been specified in the config.

hueniverse:
@kanongil I agree. This seems like a bug in the browser implementation since Origin is clearly allowed or this whole thing doesn't work.

PR #2931, opened 2015-11-05
Lookup route during OPTIONS by using request.info.hostname. Closes #2930

### LINKED ISSUE #2930: Route not correctly looked up during CORS OPTIONS request when using vhost
Migrated from https://github.com/hapijs/glue/issues/38

mtharrison:
During a normal request the route is looked up using the [value of `request.info.hostname`](https://github.com/hapijs/hapi/blob/master/lib/request.js#L331):
```javascript
const match = this.connection._router.route(this.method, this.path, this.info.hostname);
```

However during an OPTIONS request, it's currently being looked up via the [value of `request.headers.host`](https://github.com/hapijs/hapi/blob/master/lib/cors.js#L101):
```javascript
const route = request.connection.match(method, request.path, request.headers.host);
```
These values are different when not using port 80, therfore this can cause the route not to be found correctly when there's a vhost setting on a route as per https://github.com/hapijs/glue/issues/38 https://github.com/hapijs/hapi/issues/2930.

### DIFF
--- a/lib/cors.js
+++ b/lib/cors.js
@@ -98,7 +98,7 @@ internals.handler = function (request, reply) {
 
     // Lookup route
 
-    const route = request.connection.match(method, request.path, request.headers.host);
+    const route = request.connection.match(method, request.path, request.info.hostname);
     if (!route) {
         return reply(Boom.notFound());
     }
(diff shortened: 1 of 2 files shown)


devinivy:
Thanks for tracking this down– looks good to me.

PR #2940, opened 2015-11-11
Fails to set a global route auth config

giovannicalo:
After the [recent special routes changes](https://github.com/hapijs/hapi/commit/98d34046392006540b64a799b6ca58150d08df2c#diff-259dec4414a400a6e895f16ff1d0ca3bR377), a server configured to use an authentication strategy by default for all routes...
```
const server = new Hapi.Server({ connections: { routes: { auth: "foo" } });
server.connection(...);
...
server.auth.strategy("foo", ..., ...);
```
... would fail to start and instead throw an error.
```
Unknown authentication strategy: foo in path: /{p*}
```
I also doubt the special routes are supposed to require authentication in the first place, as they didn't till the aforementioned commit.

This fixes it.

However, it might be better to just ignore authentication-related options for routes where `method === "_special"`, instead of passing `config: { auth: false }` every time.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -380,8 +380,8 @@ internals.Connection.prototype._addRoute = function (config, plugin) {
 
 internals.Connection.prototype._defaultRoutes = function () {
 
-    this._router.special('notFound', new Route({ method: '_special', path: '/{p*}', handler: internals.notFound }, this, this.server, { special: true }));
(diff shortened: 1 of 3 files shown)


gergoerdosi:
Can you add a test that fails without your changes and passes with them?

PR #2962, opened 2015-12-01
allow origin header to be capitalized

FuzzySockets:
If origin header is capitalized, hapi returns a 404 for CORS preflight (OPTIONS) request.

### DIFF
--- a/lib/cors.js
+++ b/lib/cors.js
@@ -156,13 +156,13 @@ exports.headers = function (response) {
 
     response.vary('origin');
 
-    if (!request.headers.origin ||
-        !internals.matchOrigin(request.headers.origin, settings)) {
+    const origin = request.headers.origin || request.headers.Origin;
 
+    if (!origin || !internals.matchOrigin(origin, settings)) {
         return;
     }
 
-    response._header('access-control-allow-origin', request.headers.origin);
+    response._header('access-control-allow-origin', origin);
 
     if (settings.credentials) {
         response._header('access-control-allow-credentials', 'true');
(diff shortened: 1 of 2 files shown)


hueniverse:
Huh? node sets all headers to lowercase. Where are you getting uppercase headers?

bboysathish:
@hueniverse I think he is trying to say, a request originates from browser/mobile client which sets the `origin` as `Origin` then hapi throws error.

FuzzySockets:
@bboysathish That's correct.  

@hueniverse So at this point the headers should already be all lowercase? I can't for the life of me figure out why the OPTIONS preflight request is returning a 404. I've configured CORS to a T based on the latest documentation.  Tried replicating the request in the unit tests with origin set with an uppercase "Origin" header as it is in my original request, but I guess this is not the issue?

bboysathish:
@FuzzySockets  Can you post that unit test in a gist and share here.

FuzzySockets:
@bboysathish https://gist.github.com/FuzzySockets/b132e464b34fdccbdb5b

FuzzySockets:
@bboysathish This was the failing unit test, but the request that is failing in my application is a POST with some data.  Let me know if you want the headers for that as well.

FuzzySockets:
https://github.com/visionmedia/superagent/issues/679 Looks like it's probably related to the front-end xhr library.

@hueniverse I see what you're talking about here: https://github.com/nodejs/node/blob/master/lib/_http_incoming.js#L128.  I assume by the time the node request hits server.inject this has already been run.

hueniverse:
If a request comes in through the node http parser, it is forced to lowercase. I don't know what you are doing and how the request is coming.

PR #2972, opened 2015-12-13
Add validation check for stripUnknown route response option

johnbrett:
This might save someone some time debugging in the future. Currently we can add:
````javascript
server.route({
    method: 'GET',
    path: '/',

        return reply('ok');
    },
    config: {
        response: {
            schema: Joi.string(),
            options: {
                stripUnknown: true
            }
        }
    }
});
````
Which actually wouldn't strip any unknown keys without the `modify` key set true. With this PR if we add a route with response option `stripUnknown` set true without `modify` set to true, it will throw a validation error.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -129,7 +129,8 @@ internals.routeBase = Joi.object({
         modify: Joi.boolean(),
         options: Joi.object()
     })
-        .without('modify', 'sample'),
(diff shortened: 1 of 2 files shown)


hueniverse:
You made travis sad... Just remove the trailing spaces.

PR #2997, opened 2015-12-31
Allow custom expect-100 logic

fiws:
Hello,

while testing upload related stuff with hapi I noticed that there currently is no option to prevent [node from sending `100-continue`](https://nodejs.org/dist/latest-v5.x/docs/api/http.html#http_event_checkcontinue) for an requests where the client expects it (`Expect: 100-continue` Header).

This PR adds the possibility to send the 100-continue when needed (for example after validation). For now this has to be done via `request.raw.req.sendContinue()`.
I thought about adding a method to `reply`, but there already is `reply.continue()` which could be easily confused.

I've also added an option (`autoContinue`) which sends `100-continue`s to requests which expect it. (This defaults to true, so hapi still behaves the same way)

This is my first PR to hapi. I'm expecting that I did everything wrong that can be done wrong ;)
I'm open for criticism and am expecting changes to this PR before the merge.
Also I'm not sure if the tests are testing enough.

hapi new year's eve! :fireworks: :tada: :heart:

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -86,6 +86,7 @@ exports = module.exports = internals.Connection = function (server, options) {
 
     this.listener = this.settings.listener || (this.settings.tls ? Https.createServer(this.settings.tls) : Http.createServer());
     this.listener.on('request', this._dispatch());
+    this.listener.on('checkContinue', this._dispatch());
     this._init();
 
     this.listener.on('clientError', (err, socket) => {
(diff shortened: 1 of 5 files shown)


hueniverse:
This is actually a lot more complicated because of how the request lifecycle works and where certain conditions are validated. I am not sure it's worth the effort.

hueniverse:
I am going to close this because after 2 months no one has expressed any need for this and it's pretty intrusive. I am also not sure this is enough given everything else that can happen to a request. Thanks.

PR #3014, opened 2016-01-06
Update JSON response stringify for better performance

feego:
Hi, I've been doing some profiling to a REST API I've been working on and noticed that, for large JSON responses, a lot of time was being spent stringifying the actual responses. After some digging, I found out that the performance of the native V8 JSON.stringify degrades very quickly when the method is called with more than 1 argument (the Javascript object to stringify itself), even if the other arguments are null or undefined.

I've done some simple but yet conclusive tests with JSON.stringify and got some results to share:

* Object to stringify: [fathers.txt](https://github.com/hapijs/hapi/files/80244/fathers.txt)
* Script used:

```javascript
// Copy pasted data into variable fathers.
var fathers = ....;

// Test 1.
var startTime = Date.now();
JSON.stringify(fathers, null, null);
console.log(Date.now() - startTime);
// Prints 215.

// Test 2.
startTime = Date.now();
JSON.stringify(fathers);
console.log(Date.now() - startTime);
// Prints 20.
```

Hapi is currently stringifying all JSON response objects calling JSON.stringify always with all the 3 arguments, even if the last 2 are null. I updated the JSON.stringify call in Hapi to use this last 2 arguments only when they have non falsy values.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -497,7 +497,8 @@ internals.Response.prototype._streamify = function (source, next) {
         const replacer = options.replacer || this.request.route.settings.json.replacer;
         const suffix = options.suffix || this.request.route.settings.json.suffix || '';
         try {
-            payload = JSON.stringify(payload, replacer, space);
+            var arguments = replacer || space ? [payload, replacer, space] : [payload];
+            payload = JSON.stringify.apply(JSON, arguments);
         }
         catch (err) {
             return next(err);
--- a/lib/response.js
+++ b/lib/response.js
@@ -499,7 +499,7 @@ internals.Response.prototype._streamify = function (source, next) {
         try {
             if (replacer || space) {
                 payload = JSON.stringify(payload, replacer, space);
-            } 
+            }
             else {
                 payload = JSON.stringify(payload);
             }
--- a/lib/response.js
+++ b/lib/response.js
@@ -497,8 +497,11 @@ internals.Response.prototype._streamify = function (source, next) {
         const replacer = options.replacer || this.request.route.settings.json.replacer;
         const suffix = options.suffix || this.request.route.settings.json.suffix || '';
         try {
-            var arguments = replacer || space ? [payload, replacer, space] : [payload];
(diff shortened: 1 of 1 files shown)


hueniverse:
Interesting. I don't mind the optimization but if should be implemented using an if statement, not apply. That should also fix the exception your code is throwing.

feego:
Thanks. And done.

leesei:
I've made a jsPerf for this finding:
[JSON.stringify() w/ and w/o param · jsPerf](http://jsperf.com/json-stringify-w-and-w-o-param)

I think it's a v8 bug, opened issue [here](https://bugs.chromium.org/p/v8/issues/detail?id=4730).

leesei:
It is [confirmed](https://bugs.chromium.org/p/v8/issues/detail?id=4730#c1) by the v8 team that this is the expected behavior.
This v8 specific optimization LGTM.

mcollina:
Passing here by change: you should use http://npm.im/fast-safe-stringify by @davidmarkclements. It "fixes" circular references and it is incredibly fast.

davidmarkclements:
Just to clarify, it's not as fast as using JSON.stringify as-is, but as far as I know it's the fastest "safe stringifer" - particularly for large objects

fast-safe-stringify is 2x faster for small objects, and 6x faster for large objects than json-stringify-safe

If you have no concerns around circular reference then to my knowledge, directly using JSON.stringify is as fast as you can get atm.

mcollina:
Also there is a try catch, which is not needed with fast-safe-stringify
Il giorno ven 8 apr 2016 alle 19:28 David Mark Clements <
notifications@github.com> ha scritto:

devinivy:
@mcollina That's actually a pretty good point– the try/catch deoptimizes that internal `_streamify()` function.  If someone has the time, I'd be interested to see if breaking it out into a minimal function makes a difference.

hueniverse:
Thanks for not going to add external deps.

PR #3044, opened 2016-02-03
Enable disabling cache-control headers

csabapalfi:
Sometimes you just don't want Cache-Control headers at all.

One example is old IE versions not loading fonts with a Cache-Control header.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -66,7 +66,7 @@ internals.routeBase = Joi.object({
         expiresAt: Joi.string(),
         privacy: Joi.string().valid('default', 'public', 'private'),
         statuses: Joi.array().items(Joi.number().integer().min(200)).min(1)
-    }),
+    }).allow(false),
     cors: Joi.object({
         origin: Joi.array().min(1),
         maxAge: Joi.number(),
--- a/API.md
+++ b/API.md
@@ -2223,7 +2223,9 @@ following options:
   Ignored if the method is an arrow function.
 
 - `cache` - if the route method is 'GET', the route can be configured to include caching
-  directives in the response using the following options:
+  directives in the response.
+  The default `Cache-Control: no-cache` header can be disabled by setting `cache` to `false`.
+  Caching can be customized using an object with the following options:
     - `privacy` - determines the privacy flag included in client-side caching using the
       'Cache-Control' header. Values are:
         - `'default'` - no privacy flag. This is the default setting.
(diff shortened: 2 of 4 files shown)


csabapalfi:
Is anyone looking at hapi PRs at all? @hueniverse?

mtharrison:
It would feel more consistent with existing api IMO if you could just set `cache` to `false` instead of the `disabled` option. Also I think you should add tests.

csabapalfi:
Thanks for the feedback, will simplify it and add tests!

hueniverse:
I am looking at PRs but this is not a high priority. Since I am not getting paid by anyone to do this, my priorities are pretty limited to security, major bugs, and stuff I need myself. Features like that tend to wait for a release for one of the above reasons (since I'm already spending time preparing it).

That said, most things get merged pretty fast. This one is just very low on my list and has very little interest from other people.

csabapalfi:
Understood and makes sense. Maybe just tagging it (even if the tag says low-priority) would let everyone know about the situation. Actually not touching the PR at all is probably even clearer. Thanks.

hueniverse:
@csabapalfi please make the changes suggested by @mtharrison

csabapalfi:
@hueniverse Done. Thanks for looking into this!

csabapalfi:
An unrelated tests fails with node v4.0. 

Tried reproducing that locally with no luck. :(

(What I did: `nvm install v4.0 && nvm use v4.0 && rm -rf node_module && npm test` - tests pass just fine)

csabapalfi:
An unrelated tests fails with node v4.0. Haven't managed to reproduce that locally :(

lib/schema.js
```diff
@@ -61,12 +61,15 @@ internals.routeBase = Joi.object({
...
-    }),
+    cache: Joi.alternatives(
+        Joi.object({
+            expiresIn: Joi.number(),
+            expiresAt: Joi.string(),
+            privacy: Joi.string().valid('default', 'public', 'private'),
+            statuses: Joi.array().items(Joi.number().integer().min(200)).min(1)
+        }),
+        Joi.boolean().valid(false)
+    ),
```
hueniverse: Instead just add `.allow(false)` to the existing object rule.

csabapalfi: Thanks, that's much simpler. Done.

AdriVanHoudt:
Can I suggest updating the docs as well stating this can be set to false? https://github.com/hapijs/hapi/blob/master/API.md#route-options

csabapalfi:
@AdriVanHoudt :+1: Good point. Updated the docs.

PR #3054, opened 2016-02-22
Upgrade lab and use shuffle option

geek:
* More hapi lint rules 
* Better code coverage detection

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,12 +42,12 @@
     "code": "2.x.x",
     "handlebars": "4.x.x",
     "inert": "3.x.x",
-    "lab": "8.x.x",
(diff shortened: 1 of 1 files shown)


hueniverse:
Does shuffle keep the test id consistent so -i works?

geek:
@hueniverse Not yet... I will work on fixing this so that the number is retained. At the moment, shuffle is valuable for travis runs more than local test runs with `-i`.

hueniverse:
Yeah, I use npm test all the time and then I call lab -i to repeat issues so this will break it for me.

geek:
@hueniverse fixed

PR #3057, opened 2016-02-24
Don't re-initialize the server

gergoerdosi:
I often run into issues in tests when `server.initialize()` is called multiple times. At the moment an error is thrown when it happens:

> Cannot initialize server while it is in initialized state

This PR changes the behavior, when the server is in `initialized` state, it doesn't attempt to run the initialization process again, it just calls the callback. I opened this PR to receive initial feedback, happy to improve it later.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -176,21 +176,21 @@ internals.Server.prototype.initialize = function (callback) {
         return Promises.wrap(this, this.initialize);
     }
 
-    const errorCallback = Hoek.nextTick(callback);
+    const nextTickCallback = Hoek.nextTick(callback);
     if (!this.connections.length) {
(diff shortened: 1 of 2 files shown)


lib/server.js
```diff
@@ -185,6 +185,10 @@ internals.Server.prototype.initialize = function (callback) {
         return errorCallback(new Error('Cannot start server before plugins finished registration'));
     }
 
+    if (this._state === 'initialized') {
+        return callback();
```
hueniverse: Technically it should use the `errorCallback` after renaming it to `nextTickCallback`

gergoerdosi: Renamed `errorCallback` to `nextTickCallback`.

PR #3068, opened 2016-02-29
Fast server shutdown

kanongil:
When a server is handling persistent connections (using http keep-alive), the `timeout` option for `server.stop()` is a bit mislabeled. If just one of the persistent connections don't make any requests within the timeout duration, the server waits for the timeout even though no requests are processing. As such, it is effectively a stop delay option.

This patch tracks the request processing state of the connections, and immediately ends idle connections on `server.stop()`. Additionally, it signals that the server is closing on active requests (by adding `connection: close` to the response), and forcibly ends the connection once the response has been submitted. This means that the `timeout` option works as expected, and the server is much more likely to stop before the timeout is reached).

Note that actively closing these connections means that the closed server check on new requests from #2363, is no longer needed.

The first 3 patches adds additional tests that validate the current behavior, and the final patch actually implements the new behavior along with new and modified tests.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -209,6 +209,16 @@ internals.Connection.prototype._stop = function (options, callback) {
         this._init();
         return callback();
     });
+
+    // Tell idle keep-alive connections to close
+
+    Object.keys(this._connections).forEach((key) => {
+
+        const connection = this._connections[key];
+        if (!connection._isHapiProcessing) {
+            connection.end();
+        }
+    });
 };
 
 
@@ -218,10 +228,17 @@ internals.Connection.prototype._dispatch = function (options) {
 
     return (req, res) => {
 
-        if (!this._started &&
-            !Shot.isInjection(req)) {
+        // Track socket request processing state
 
-            return req.connection.end();
+        if (req.socket) {
+            req.socket._isHapiProcessing = true;
+            res.on('finish', () => {
+
+                req.socket._isHapiProcessing = false;
+                if (!this._started) {
+                    req.socket.end();
+                }
+            });
         }
 
         // Create request
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -285,6 +285,11 @@ internals.transmit = function (response, callback) {
         }
     }
 
+    const isInjection = Shot.isInjection(request.raw.req);
+    if (!isInjection && !request.connection._started) {
+        request.raw.res.setHeader('connection', 'close');
+    }
+
     request.raw.res.writeHead(response.statusCode);
 
     // Write payload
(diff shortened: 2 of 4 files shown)


hueniverse:
This keeps failing on node v4.0.

kanongil:
OK, I will look into it.

kanongil:
My investigation shows that the test fails because wreck never receives the reply. The behavior changes in the v4.1.1 to v4.1.2 update, and I believe my test exposes the bug / security issue fixed in https://github.com/nodejs/node/commit/05040664c2035b8abffb7bd7bb3243af35895130.

Do we really need to fully support 4.0.0 considering https://github.com/hapijs/contrib/issues/54?

kanongil:
Look at that, my rebased version passes :wink: 

Let me know if you have any other concerns.

lib/transmit.js
```diff
@@ -285,6 +285,11 @@ internals.transmit = function (response, callback) {
         }
     }
 
+    const isInjection = Shot.isInjection(request.raw.req);
+    if (!isInjection && !request.connection._started) {
```
hueniverse: I don't like accessing non-public node apis with `_started`.

lib/connection.js
```diff
@@ -218,10 +228,17 @@ internals.Connection.prototype._dispatch = function (options) {
...
+        // Track socket request processing state
+
+        if (req.socket) {
+            req.socket._isHapiProcessing = true;
+            res.on('finish', () => {
 
-            return req.connection.end();
+                req.socket._isHapiProcessing = false;
+                if (!this._started) {
+                    req.socket.end();
```
hueniverse: Why the switch from `req.connection` to `req.socket`?

lib/transmit.js
```diff
@@ -285,6 +285,11 @@ internals.transmit = function (response, callback) {
...
+        request.raw.res.setHeader('connection', 'close');
```
hueniverse: Why not set this header on injection?

lib/transmit.js
kanongil: This is a Hapi connection property lookup to know when `server.stop()` has been called.

lib/transmit.js
kanongil: Since injections apply to a non-started server, it would add this header to every request. Further, injections are not connection oriented and it doesn't make sense to mess with this header.

lib/connection.js
kanongil: Because `socket` is a documented property of an `http.IncomingMessage`.

lib/transmit.js
hueniverse: My bad.

PR #3072, opened 2016-03-03
Test realm pluginOptions in multiple routes using route and reply interface

geek:
@hueniverse there is a question on if `request.server.realm.pluginOptions` should contain the plugin options for the related realm?  At the moment, this value is an empty object.

I documented another way to get this value in a route, which is using reply.realm.


test/plugin.js
```diff
@@ -550,6 +550,39 @@ describe('Plugin', () => {
...
+            server.connection({ labels: ['a', 'b'] });
+
+            const test = function (srv, options, next) {
+
+                expect(options.something).to.be.true();
+                expect(srv.realm.pluginOptions).to.equal(options);
+
+                srv.route({ method: 'GET', path: '/', handler: (request, reply) => {
+
+                    expect(request.server.realm.pluginOptions).to.deep.equal(options);
```
geek: `request.server.realm.pluginOptions` is `{}` here

devinivy:
I would expect `request.server.realm` to be the realm in which the route was registered.  So at a glance this does seem to be a valid issue to me.

On the other hand, is `request.server` the same as `srv.root`?

devinivy:
Nice.  You surfaced some new pieces of the public API to the docs– might as well be a feature!  Do we think the docs listed for `request.server` are specific enough?  It seems to be that it will always be the root server.

geek:
@devinivy I think the current `request.server` are enough.

PR #3074, opened 2016-03-07
Update API.md: added a more precise description for `request.app`

mattii:
Reference: https://github.com/hapijs/discuss/issues/239

`request.app` seems to be the place for request-specific state. This is a proposal for a more precise description.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2134,8 +2134,8 @@ server.route({ method: 'GET', path: '/user', config: user });
 
 Each route can be customized to change the default behavior of the request lifecycle using the
 following options:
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -2134,8 +2134,8 @@ server.route({ method: 'GET', path: '/user', config: user });
 
 Each route can be customized to change the default behavior of the request lifecycle using the
 following options:
-- `app` - application-specific configuration. Should not be used by [plugins](#plugins) which
```
hueniverse: Just replace `configuration` with `request state`

devinivy: I'd call it "request-specific application state", but that's just me!

Issue #2232, opened 2014-12-01
Use parsed payload during strategy validation ?


hems:
In the implementation i'm doing i need to validate user_id / token from a payload, in order to do that i would like to access the parsed content of `request.payload` during my strategy in order to query my login database and forward the user credentials.

But unfortunately during that phase of the Request Cycle, request.payload is undefined.

Any workarounds? Perhaps i should be using som server extension or another module in order to implement that?

example code:

```
server.auth.scheme("payload_auth", function(server, options) {
  return {
    authenticate: function(request, reply) {
      var find, payload, token, user_id;
      payload = request.payload.session;
      user_id = payload.user_id;
      token = payload.token;
      find = {
        _id: user_id,
        token: token
      };
      return db.users.findOne(find, function(error, user) {
        if (error) {
          return console.log(error);
        }
        return reply(null, {
          credentials: {
            user: user
          }
        });
      });
    }
  };
});

server.auth.strategy('my_custom_auth', 'payload_auth');
```

Issue #2066, opened 2014-10-29
Configure auth scheme to require payload validation


mdarnall:
Hi, 

I writing a custom auth scheme for the [Salesforce signedrequest](http://www.salesforce.com/us/developer/docs/platform_connect/Content/canvas_app_unsigning_code_example.htm) authentication. In order for it to verify authentication it needs the body of the `POST` message. 

My initial implementation was to implement the `authenticate` method, but I then realized it is called before `request.payload` is parsed. 

My second implementation moved the verification of the `POST` to a `payload` method of the scheme, while the `authenticate` just always returns valid. I Then have to configure the route to 'require' the payload parameter. 

Is this the right approach? It felt 'off' to always return valid in the `authentication` method and then to do the validation in `payload` while setting the right properties on `request.auth` after. My implementation is [outlined in this gist](https://gist.github.com/mdarnall/22f5eef5ed3ea3081f58)

PR #3075, opened 2016-03-08
Include an option to parse payload before authenticate handler is triggered

mcastany:
This will allow us to have `request.payload` inside the `authenticate` handler when we set the following config

```
{ 
  config : { 
    auth : { 
      strategy : 'foo',
      parsePayloadBeforeAuth : true 
    } 
  } 
}

```

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -283,13 +283,19 @@ internals.Route.prototype.rebuild = function (event) {
         cycle.push(this._extensions.onPreAuth);
     }
 
+    if (this.method !== 'get' && this.settings.auth && this.settings.auth.parsePayloadBeforeAuth) {
+        cycle.push(internals.payload);
+    }
+
     const authenticate = (this.settings.auth !== false);                          // Anything other than 'false' can still require authentication
     if (authenticate) {
         cycle.push(Auth.authenticate);
     }
 
     if (this.method !== 'get') {
-        cycle.push(internals.payload);
+        if (!this.settings.auth || !this.settings.auth.parsePayloadBeforeAuth){
+            cycle.push(internals.payload);
+        }
 
         if (authenticate) {
             cycle.push(Auth.payload);
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -47,7 +47,8 @@ internals.auth = Joi.alternatives([
         payload: [
             Joi.string().valid('required', 'optional'),
             Joi.boolean()
-        ]
+        ],
+        parsePayloadBeforeAuth :Joi.boolean()
     })
         .without('strategy', 'strategies')
         .without('access', ['scope', 'entity'])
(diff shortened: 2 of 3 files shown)


mtharrison:
Related: https://github.com/hapijs/hapi/issues/2066
and: https://github.com/hapijs/hapi/issues/2232

devinivy:
I'm not a huge fan of there being inconsistency in the request lifecycle.  @mcastany was there something in particular that you were hoping to be able to achieve that the payload authentication step can't do?

mcastany:
@devinivy Yes. the scopes from the auth configuration are verified before the payload authentication step. If I move the authentication logic to the payload authentication step, I have the following issues:

1. In the authenticate step, I need to include a hardcoded list of scopes to pass the scopes validation.
2. In the payload authenticate step, when I get the list of scopes for the user, I'm not able to get them validated against the defined list of scopes in the endpoint.

devinivy:
Another solution to this issue would be for hapi to expose the logic to test access (scopes and entities) for the request, not dissimilar to `server.auth.test()`.

hueniverse:
@mcastany  Sorry but I am no going to make this option available. The request lifecycle must never change as this will create express-like middleware hell.

If you can't work with the rules available via the auth API, just write your own extension that runs after payload is parsed.

@devinivy no need for a new API, you can just create an auth scheme that doesn't actually do anything and call `server.auth.test()` with it to get access to the rest of the auth features.

dschenkelman:
@hueniverse related to this: how can we (@mcastany and I) validate a route `scope` that varies depending on the payload? The problem is that the scope validation runs after `auth` (not after `payload`) and the scopes are there regardless of the authentication method (via payload or via header).

Don't know if this makes sense or missing something

hueniverse:
@dschenkelman the user scope changes based on the request payload?

mcastany:
@hueniverse yes, we need the request payload to authenticate the user and then validate the scopes.

With the current request lifecycle, we see the following steps: `authenticate` -> `validate auth scopes` -> `parse payload` -> `payload authenticate`

By moving the authentication logic to the `payload authenticate`, we were able to use the request payload, but now we don't have a way to check the scopes.

Our ideal flow would be: `parse payload` -> `authenticate` -> `validate scope` so if we leave the auth logic in the `payload authenticate` step, is there a way to validate scopes after that ?

Marsup:
Why aren't you using `server.inject` to proxy the query to another route ? Not sure if that bypasses the scopes check though.

hueniverse:
@mcastany ok. Still not going to change the auth and request life cycle flows to support this use case. Sorry. If I come up with a better solution I'll let you know.

dschenkelman:
Great, thanks for letting us know. We are going to review this internally, see what we come up with and post back here.

Issue #2513, opened 2015-04-24
Invalid cookie value


seemsindie:
I have a blank server running with just , and it's working fine. But when i forward the port trough NAT because i have server listening inside VM and when i try to access the API on the host machine i get this:
{"statusCode":400,"error":"Bad Request","message":"Invalid cookie value"}

Any idea why is this happening, and how to fix it?

edimoldovan:
Looks like a validation error, at least I get similar responses when validation fails. Do you have any kind of request validation in there?

seemsindie:
nope, just this:
    server.route({
        method: 'GET',
        path: '/wer',
        handler: function(req, res) {
          res('Hello World!');
        }
      });

edimoldovan:
Just tried your code, works perfectly here. Can you post the rest of your app.js?

MathieuLoutre:
I've seen that before and I'm not sure what's exactly happening (although nothing as complex as port forwarding was needed, just a simple server). The server wasn't even using any cookie related thing explicitly (no cookie auth etc.). Can you try in an incognito window?
I suspect a malformed cookie has been set somehow and hapi can't deal with it.

seemsindie:
@edimoldovan 
var Hapi = require('hapi');

var server = new Hapi.Server();
server.connection({ port: 3000 });
server.route({
method: 'GET',
path: '/',
handler: function(req, res) {
res('Hello World!');
}
});

server.start(function () {
    console.log('Server running at:', server.info.uri);
});

But there is no issue on the host machine, but where the port is forwarded.

@MathieuLoutre Yes, it works in incognito, do you know witch cookie is the issue?

MathieuLoutre:
@seemsindie no clue. But if you find out, I'm interested!

seemsindie:
@MathieuLoutre i have another project on same domain 'localhost', and that project have two cookies, tr and tr_tkn. Can i somehow turn off cookie parsing, or if i can parse what i just want?

MathieuLoutre:
Hum, I'm not 100% sure as I haven't done it myself, but I'd look here: https://github.com/hapijs/hapi/blob/master/API.md#route.config.state

seemsindie:
@MathieuLoutre Yep, i just added 
config: {
      state: {
        parse: false, // parse and store in request.state
        failAction: 'ignore' // may also be 'ignore' or 'log'
      }
    }
to the route and it's working. 
Thanks man!

MathieuLoutre:
[...]

PR #3094, opened 2016-03-21
Change state to be default off

olekenneth:
Because this creates a lot of 400 - Bad Cookie Value, it's best to have this option turned off by default and activate it when needed

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -58,8 +58,8 @@ exports.connection = {
         },
         security: false,                            // Security headers on responses: false -> null, true -> defaults, {} -> override defaults
         state: {
-            parse: true,                            // Parse content of req.headers.cookie
-            failAction: 'error'                     // Action on bad cookie - 'error': return 400, 'log': log and continue, 'ignore': continue
+            parse: false,                            // Parse content of req.headers.cookie
+            failAction: 'ignore'                     // Action on bad cookie - 'error': return 400, 'log': log and continue, 'ignore': continue
         },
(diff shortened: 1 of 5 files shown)


Marsup:
Never happened to me in 3 years of hapi dev, what are you doing to cause this?

geek:
I have seen this happen, but it was always an issue outside of hapi.... I think the defaults should stay as-is in order to help detect client code that sends bad cookie headers.

nlf:
I completely agree. It makes way more sense to have this on and require a user to explicitly disable it.

This would be a breaking change that I would see causing more confusion than anything.

mtharrison:
Yes, seems like a special case where you'd be getting issues (e.g. cookies not set by hapi). So current default makes sense.

olekenneth:
We are using hapi on domains with other webservices, that sets cookies we cannot control. I can only imagine that is a common problem. Do have numbers on how many of the hapi-users actually using the `state`? In the years after SPA and localStorage, I really don't see the point of using cookies. I say that if you really want this stuff, you should be able to turn it on yourself.

AdriVanHoudt:
I don't see a reason why hapi should set these defaults? They are safer (as in less likely to shoot yourself in the foot when not dealing with legacy cookies) and if you want to change these you can easily set these defaults to your connection config see https://github.com/hapijs/hapi/issues/2513#issuecomment-95881509

hueniverse:
I'm not making this change.

PR #3121, opened 2016-04-13
allow array as valid validate config on headers, params, query and payload

### LINKED ISSUE #3119: Support [] as payload validation
regarding https://github.com/hapijs/hapi/blob/af9591356114f64485546cac0aa6b540a010c7da/lib/schema.js#L160
now `{}` works instead of `Joi.object()` so I would expect `[]` to work instead of `Joi.alternatives().try()`
Joi will auto convert it to alternatives so it will work

This would be a trivial PR but still asking first ;)

AdriVanHoudt:
fixes #3119

was not sure where to test, just add a test with arrays under test/validation.js ?

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -116,8 +116,8 @@ internals.routeBase = Joi.object({
         options: Joi.object(),
         ranges: Joi.boolean(),
         sample: Joi.number().min(0).max(100),
-        schema: Joi.alternatives(Joi.object(), Joi.func()).allow(true, false),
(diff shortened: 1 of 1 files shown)


Marsup:
Missing response.schema for consistency.

AdriVanHoudt:
@Marsup also added on status since it also allows a Joi schema it looks like

AdriVanHoudt:
seems like a flakky test?

PR #3124, opened 2016-04-16
Don't force header keys to lowercase

capelio:
While the RFC clearly states that header keys are to be treated as case
insensitive, forcing them lowercase can create confusion between an
API's documentation and what a developer sees when they inspect response
headers.

For the same reasons camelCase matters when it comes to readability and
consistency, it is worthwhile to preserve the case of header keys as
they are defined via response.header(key, value).

My real world use case for this is that I am getting questions on a
semi-regular basis concerning the case differences that exist between
our API docs and our API responses. I'd strongly prefer the case I use
in our documentation to be the same case our API developers see when
they inspect and debug responses.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -111,7 +111,6 @@ internals.Response.prototype.code = function (statusCode) {
 
 internals.Response.prototype.header = function (key, value, options) {
 
-    key = key.toLowerCase();
     if (key === 'vary') {
         return this.vary(value);
     }


hueniverse:
This would require a lot more effort because the code assumes headers are always lowercase for the purpose of merging and replacing headers. To make this work, we would need to keep a side map of the case you originally set, and then override that at the very end when we set the headers. This is not worth the complexity and unpredictability it will add.

capelio:
Ah, I should have looked deeper to see the ramifications of this kind of change. Thanks for taking the time to explain them, makes sense.

PR #3131, opened 2016-04-29
Improve wording of auth strategy mode/default

wprl:
I found this wording confusing, and noticed other people have as well (e.g. https://github.com/dwyl/hapi-auth-jwt2/issues/137 ).  I think the proposed update better reflects when `server.auth.strategy` sets a default strategy for routes.

### DIFF
--- a/API.md
+++ b/API.md
@@ -649,10 +649,10 @@ Registers an authentication strategy where:
 - `name` - the strategy name.
 - `scheme` - the scheme name (must be previously registered using
   [`server.auth.scheme()`](#serverauthschemename-scheme)).
-- `mode` - unless set to `false`, the scheme is automatically assigned as the default strategy for any route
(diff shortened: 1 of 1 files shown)


Marsup:
Not sure it helps since the default is already false.

wprl:
The original wording makes it sound like the strategy will only become the default auth for routes if `required` or `true` is specified as the mode.  The reality is that when anything other than `false` is specified for the mode, the strategy will be set as the default strategy for routes.

geek:
https://github.com/hapijs/hapi/blob/master/lib/auth.js#L69-L71

We definitely can clarify the wording... can you add in that if you do set it to true the auth mode will be set to `required`

wprl:
I misunderstood you initially @Marsup .  Pushed another commit to clarify further based on your guys' suggestions.

PR #3178, opened 2016-05-24
Handle thrown error from res.setHeader() and res.writeHead()

kanongil:
The behavior of `res.writeHead()` is changed in Node@6 (and possibly node@4.4.5) so that it can throw a `RangeError` for certain statusCodes, see https://github.com/nodejs/node/pull/6291. When this throws, hapi will log an internal error without sending a response.

This PR adds a `try/catch` to catch this error and respond with a `500` to the client instead.

Additionally, it will catch thrown errors from `res.setHeader()`, eg. for an empty string header, and respond as above.

### DIFF
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -291,29 +291,9 @@ internals.transmit = function (response, callback) {
 
     // Write headers
 
-    const headers = Object.keys(response.headers);
-    try {
-        for (let i = 0; i < headers.length; ++i) {
-            const header = headers[i];
-            const value = response.headers[header];
-            if (value !== undefined) {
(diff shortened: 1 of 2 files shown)


hueniverse:
Probably another week or so.

lib/transmit.js
```diff
@@ -369,6 +373,34 @@ internals.transmit = function (response, callback) {
...
+            }
+        }
+
+        res.writeHead(response.statusCode);
+    }
+    catch (err) {
+
+        // Reset headers and return err
+
+        res._headers = null;
```
hueniverse: Why?

PR #3179, opened 2016-05-25
Send correct response for HEAD requests

kanongil:
I noticed that a `HEAD` request for some of my routes did not send the `content-length` header, while a normal `GET` request would. I have updated the logic to properly set the header. This is done by not skipping the marshal step.

This also fixes specific cases where the `content-type` header could be different for `GET` and `HEAD` if it is set during or post marshaling.

FYI, the modified and new test both fail on current master as follows:
```
Failed tests:

  141) Connection route() responds to HEAD requests for a GET route:

      Expected undefined to equal specified value

      at /hapijs/hapi/test/connection.js:1690:63

  143) Connection route() returns 500 on HEAD requests for failed responses:

      actual expected

      200500

      Expected 200 to equal specified value

      at /hapijs/hapi/test/connection.js:1749:48
```

### DIFF
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -66,7 +66,7 @@ internals.marshal = function (request, next) {
 
         internals.cache(response);
 
-        if (!response._isPayloadSupported()) {
+        if (!response._isPayloadSupported() && request.method !== 'head') {
 
             // Close unused file streams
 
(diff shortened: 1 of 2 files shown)


kanongil:
Rebased to latest master. Any chance you could look into this?

It's a bit unfortunate that `GET`and `HEAD` for the same resource can return different `content-type` headers.

Also, what happened to the travis integration?

PR #3187, opened 2016-06-05
Add a new 'started' event, emitted when start callbacks are complete

rob-h-w:
Rename test variables started -> start to denote the semantic difference
between the 2 events.

Update the documentation.

This new event is useful to allow users to determine if the server
(including any bespoke initialisation in start's callback) is started.

This change was prompted by the repeated need to implement the following:

```
server.start(err => {
  // init that could fail.

  server.emit('started');
});

server.on('started', () => {
  log('server startup complete');
});
```


mtharrison:
Is this not already covered with a server `onPostStart` extension?

http://hapijs.com/api#serverextevents

mtharrison:
"any bespoke initialisation in start's callback" could be async so I don't think it's satisfactory to emit the event on callback running to completion.

hueniverse:
What's the use case?

rob-h-w:
Agree with @mtharrison, this does not support asynchronous activity in the callback. I think I'll drop this change. Thanks for the feedback!

@hueniverse the use case I ran into this was primarily testing - repeatedly having to synchronize with server startup before injecting requests. I was writing the same code for multiple servers, but that doesn't necessarily justify the code being in the server.

mtharrison:
@rob-h-w I've seen a few patterns for getting an initialized server into tests. Some fall down when there are plugins + caches involved. My favourite approach is having the thing to be tested as a hapi plugin and creating a new server object per test.

I've written an example Gist to show you what I mean: https://gist.github.com/mtharrison/633ee77d1ea0e2a2653403ce85c78f2a

devinivy:
If you're trying to test an initialized server exported and configured elsewhere (possibly composed using glue), I built this tool for the job– https://github.com/devinivy/labbable.  I am a big fan of @mtharrison's approach too if it suits your app.

PR #3196, opened 2016-06-08
Bind Promise's callbacks to current domain if set.

rmedaer:
This prevent crashing when server method returning a Promise with callback option set to false and a route handler throw an error.

How to reproduce ?
```js
const Hapi = require('hapi');

server.connection({
    port: 3000
});


server.method('method', () => Promise.resolve(), { callback: false })

server.route({
    method: 'GET',
    path: '/test',
    config: {
        pre: [
            { method: 'method()' },
        ]
    },
    handler: function(request, reply){
        throw new Error('Oops');
    }
});

server.start(() => {
    console.log('Server listening ...')
})
```

### DIFF
--- a/lib/methods.js
+++ b/lib/methods.js
@@ -107,10 +107,11 @@ internals.Methods.prototype._add = function (name, method, options, realm) {
             };
 
             // Bind Promise's callbacks to current domain if set
-            let namespace = process.domain;
+            const namespace = process.domain;
             if (namespace) {
                 result.then(namespace.bind(onFulfilled), namespace.bind(onRejected));
-            } else {
+            }
+            else {
                 result.then(onFulfilled, onRejected);
             }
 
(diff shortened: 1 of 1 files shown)


rmedaer:
Still have to fix linting and tests ... ;-)

rmedaer:
Ok so I'll need you help !

Because the error is Node's internal between promises and domains I'm not able to write tests with inject method and so on...

Could you help me with this ?


```js 
    it('registers a method returning promise with no callback and a route throwing error', (done) => {

        const add = function () {

            return new Promise((resolve, reject) => {

                return resolve();
            });
        };

        server.connection();
        server.method('add', add, { callback: false });

        server.route({
            method: 'GET',
            path: '/test',
            config: {
                pre: [
                    'add()'
                ],

                    throw new Error('Oops');
                }
            }
        });

        server.inject('/test', (res) => {

            done();
        });
    });
```

hueniverse:
@Marsup Is this still needed?

Marsup:
It looks like the same case as #3202, maybe @rmedaer can confirm it's fixed ?

rmedaer:
@Marsup it was the same issue then #3203 . I'll test and confirm it's fixed ASAP...

Issue #3037, opened 2016-01-29
brotli support


patrickkettner:
It would be great to be able to server assets with [brotli](https://en.wikipedia.org/wiki/Brotli) - especially if I could precompile and serve with `lookupCompressed`. In general it has a much better compression ratio than gzip.

hueniverse:
I don't know what exactly you are asking for.

jamesdixon:
He appears to be asking for Hapi to support brotli compression in addition to gzip as well as allowing inert to lookup pre-compressed assets created with brotli. Maybe something that could be done in a plugin for the time being? [Browser support appears to be nearly non-existent at this point...](http://caniuse.com/#feat=brotli).

patrickkettner:
Sorry for not being clearer. James is correct. Browser support is low, but it is a requirement for WOFF2 web fonts, and as such all browsers will be implementing it in the near future.

hueniverse:
I'll consider a PR.

nrotta:
@patrickkettner I will give it a try. I have an initial version (sort of) working and have already pushed the changes to Inert to my fork. Will push the initial hapi changes to my fork next.

Consider that I'm just starting with Hapi so my code might require lots of changes.

nrotta:
@patrickkettner I finished the work to support brotli on hapi. Looks like it works. Can you please check it out?

You have to use the following from my fork:

hapi (master branch): https://github.com/nrotta/hapijs
subtext (master branch): https://github.com/nrotta/subtext
inert (brotli branch): https://github.com/nrotta/inert

I have a failing test on inert so I cannot send the PRs yet. I will fix it over the next few days.

Thanks

patrickkettner:
I'd love to check it out! did you mean https://github.com/nrotta/hapi ? https://github.com/nrotta/hapijs doesn't exist

I only ask because nrotta/hapi doesn't contain any brotli commits form what I can see.

nrotta:
Yes, you are right. Sorry typo. The correct one is: https://github.com/nrotta/hapi

Let me know how it goes.

patrickkettner:
awesome work! It is not working for me, however. I believe [this line](https://github.com/hapijs/hapi/blob/6d2896b4d387690326f7995a936cdee5e6a6b95f/lib/request.js#L119) has to be modified in order for brotli to be one of the encodings that is possible

nrotta:
[...]

PR #3198, opened 2016-06-09
Added support for brotli compression

nrotta:
Added support for brotli compression. The required changes to subtext and inert have also been done and the PRs submitted for merging.

This PR implements the following: hapijs/hapi#3037

PS: this PR was not sent from the master branch as I already have a PR on that branch waiting to be merged


nrotta:
The CI will fail as it depends on the changes done on Inert and Subtext. Not sure how to set it up.

API.md
```diff
@@ -2310,7 +2310,7 @@ following options:
           clean up the files generated by the framework. This can be done by keeping track
           of which files are used (e.g. using the `request.app` object), and listening to
           the server `'response'` event to perform any needed cleanup.
-    - `parse` - can be `true`, `false`, or `gunzip`; determines if the incoming payload is
+    - `parse` - can be `true`, `false`, `gunzip`, or `br`; determines if the incoming payload is
```
nlf: below where we describe that `true` and `gunzip` will gunzip as appropriate, you should also mention that `br` is for `brotli` support. also why abbreviate it? `brotli` is plenty short IMO

nlf:
CI won't pass until the merges are done and published for your dependencies, you'll also have to update the shrinkwrap here to the correct versions

hueniverse:
This is blocked until I figure out how to move it to to a plugin.

API.md
nrotta: @nlf it seems that this will be added as a plugin but in any case I updated the API doc (not sure if the added text is the best / most comprehensible though).

Regarding the use of `br` (instead of `brotli`), it is the naming scheme that it is being used: https://github.com/google/brotli/issues/299

nlf: fair enough, silly decision on their part IMO, but if it's an established thing then we should probably stick with it

hueniverse:
@nrotta You can now create a new plugin that simply calls `server.decoder()` and `server.encoder()`. Let me know if you need help with the new features.

kanongil:
FYI, I took some time and put together a plugin for this, currently unpublished here: https://github.com/kanongil/brok

While it works, it would be much more useful if the `request` is exposed somehow when the encode function is called. This enables route level options for eg. configurable compression levels and mode.

AdriVanHoudt:
(@kanongil why not `hotli`? ^^)

genediazjr:
@AdriVanHoudt agree
@kanongil any chance to make it `hotli`?

Issue #3102, opened 2016-03-28
Rejected promises are returned as successful status codes


mynameiscoffey:
Not sure if intended or not since there isn't anything either way in the tests or documentation, but when you reply with a Promise from a server route and that promise is rejected, hapi will return a successful status code and serialize the rejection reason as the body of the response.  It looks like resolving OR rejecting any object that is not an instance of Error is treated as a successful response, while resolving OR rejecting any object that IS an instance of Error is treated as an unsuccessful response. 

This feels very counter to how Promises are supposed to be used.  If this is expected, is there a recommended approach to turning rejected promises into an error status?

hueniverse:
Test case?

mynameiscoffey:
```
server.route({
  method: 'GET',
  path: '/',
  handler: (request, reply) => {
    reply(new Promise((res, rej) => { throw "up"; })); // returns 200 OK
  }
});
```

```
server.route({
  method: 'GET',
  path: '/',
  handler: (request, reply) => {
    reply(Promise.reject({ err: 'some issue'})); // returns 200 OK
  }
});
```

```
// idk why you would want to do this...
server.route({
  method: 'GET',
  path: '/',
  handler: (request, reply) => {
    reply(Promise.resolve(new Error("something"))); // returns 500 Server Error
  }
});
```

garthk:
What expected behaviour do you want for these?

0. Rejecting with a non-Error
0. Resolving with an Error

I'd expect 500 for both, with suitable detail in the log's fine print. Hapi's half way there already.

ilyaigpetrov:
Also would be nice if hapi could reply with an error on Promise `unhandledRejection`:
```javascript

const thirdPartyCode = function(next) {

        new Promise( (resolve, reject) => {

            console.log('Throwing!');
            throw new Error('Nobody catches me!');

        } )
            .then( next )
            // .catch( reply ) // Oh no, we forgot to add a catch!
        return 'Please, use callback.';

};

server.route({
    method: 'GET',
    path: '/',
    handler: function(request, reply) {
        return thirdPartyCode( reply ); // Reply never called.
        // return Promise.resolve( thirdPartyCode( reply ) ).catch( reply ); // Always use a con...
    }
});

process.on( 'unhandledRejection', () => console.log('UNHANDLED! Sorry.') );
```

nlf:
yeah we should fix this. i'll write up a test and patch later today.

[...]

PR #3201, opened 2016-06-10
handle more types of promise rejection, for #3102

nlf:
the only thing this doesn't cover from that issue is adding an `unhandledRejection` handler, which i'm still not sure how i feel about anyway


devinivy:
Looks good to me

PR #3202, opened 2016-06-13
Delay continuations when promises may be involved

### LINKED ISSUE #3073: Handlers and Lifecycle hooks lose domain binding
I'm still trying to work out a proper test case, but what happens is basically the following:

I have a whole bunch of plugins that hook into lifecycle hooks. 
- If I throw inside of a `onRequest` hook, everything works as expected and I get the 500 response.
- If I throw inside of a `onPreHandler` hook, the error is swallowed and the request keeps hanging until it times out.
- If I throw inside of a handler, the error is swallowed and the request keeps hanging

This clearly sounds like a domains problem, especially since I also never receive a `uncaughtException` error.
My investigation seems to point to my authenticator function which makes use of a promise and finally calls the `reply.continue` function as the point where things first go wrong, although I haven't been able to clearly track it down.

I've tried manually binding some of the callbacks to `request._protect.domain` but I didn't have much success.

In the end that means that if a developer isn't careful once with manually managing try/catch etc, loads of requests will hang without any indication on that there was an error happening.

Marsup:
Fixes #3073.

It's important with promises that anything callback-style be called on next tick, otherwise any error will be caught into the promise. I might have missed places where it would happen, happy to complete the PR if you can think of any.

### DIFF
--- a/lib/methods.js
+++ b/lib/methods.js
@@ -96,15 +96,15 @@ internals.Methods.prototype._add = function (name, method, options, realm) {
 
             // Promise object
 
-            const onFulfilled = (outcome) => {
+            const onFulfilled = Hoek.nextTick((outcome) => {
 
                 return methodNext(null, outcome);
-            };
+            });
 
-            const onRejected = (err) => {
+            const onRejected = Hoek.nextTick((err) => {
 
                 return methodNext(err);
-            };
+            });
 
             result.then(onFulfilled, onRejected);
         };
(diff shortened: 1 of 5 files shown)


devinivy:
Really glad this doesn't break any tests.

nlf:
do we have a way to test this? blah blah preventing regressions and all that

Marsup:
Adding the original issue's code probably. I don't have time currently but you can complete my PR.

hueniverse:
This has been reverted until it is fully figured out.

PR #3227, opened 2016-06-30
Set route validation bind context

### LINKED ISSUE #3226: Bound context is not available in route validation function:
```this``` in custom validation functions is undefined, so there is no way to benefit from ```server.bind(context)``` or ```server.methods```

```javascript
const Hapi = require('hapi');

server.connection({ port: 8080 });

server.bind({ foo: 'bar' });

server.route({
	method: "GET",
	path: "/{id}",
	config: {
		validate: {
			params: function(value, options, next) {
				console.log(this); // this === undefiend
				next(null, value);
			}
		}
	},
	handler: function(request, reply) {
		reply(this.foo);
	}
});

server.start(() => console.log(`Server running at ${server.info.uri}`));
```

red-0ne:
fixes #3226

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -139,7 +139,8 @@ internals.input = function (source, request, next) {
 
     request._protect.run(postValidate, (exit) => {
 
-        return schema(request[source], localOptions, exit);
+        const bind = request.route.settings.bind || request.route.realm || null;
+        return schema.call(bind, request[source], localOptions, exit);
     });
 };
 
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -139,7 +139,7 @@ internals.input = function (source, request, next) {
 
     request._protect.run(postValidate, (exit) => {
 
-        const bind = request.route.settings.bind || request.route.realm || null;
+        const bind = request.route.settings.bind || request.route.realm;
         return schema.call(bind, request[source], localOptions, exit);
     });
 };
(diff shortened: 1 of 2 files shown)


red-0ne:
How can I fix ```lib/validation.js missing coverage on line(s): 142``` ?

Marsup:
By covering all the possibilities in your tests.

nlf:
so that line looks like this:

```js
const bind = request.route.settings.bind || request.route.realm || null;
```

in order to get full coverage, you need tests where `request.route.settings.bind` are both truthy and falsey, likewise for `request.route.realm`. think of it like making sure you can actually reach that final `|| null`

lib/validation.js
```diff
@@ -139,7 +139,8 @@ internals.input = function (source, request, next) {
 
     request._protect.run(postValidate, (exit) => {
 
-        return schema(request[source], localOptions, exit);
+        const bind = request.route.settings.bind || request.route.realm || null;
+        return schema.call(bind, request[source], localOptions, exit);
```
arb: Do we want `call` on hot code?

Marsup: `call` and `apply` are not slower than bound functions last time I checked, how would you do it ?

arb: Change the whole framework so you always pass a `context` object rather than relying on `this` 😄 

But seriously, yeah there really isn't a way to do that other than `call` here.

Issue #3273, opened 2016-07-29
Add warning when validating headers that they are normalized to lowercase


fourq:
This has been a point of frustration for me several times now while testing. I don't have a solution in mind, but I'm sure a configurable warning that warns that header names/keys are lowercased would save others time, especially while testing. Maybe it's as simple as a documentation update that would prove just as useful?

hueniverse:
We can easily assert headers that are not all lowercase, but the problem is validation rules can be a joi object or a function in which case hapi has no access to this information.

fourq:
How about a documentation addition such as:

[server.inject(options, [callback])](http://hapijs.com/api#serverinjectoptions-callback)

[from]
headers - an object with optional request headers where each key is the header name and the value is the header content. Defaults to no additions to the default Shot headers.

[to]
headers - an object with optional request headers where each key is the header name and the value is the header content. Every key in this object is normalized to lowercase. Defaults to no additions to the default Shot headers.

hueniverse:
That's a bit cryptic. I would just be explicit and say "Header names must be all lowercase in order to match the header names normalized by node"

PR #3274, opened 2016-07-30
Update documentation for server.inject's option argument for headers

fourq:
#3273

Update documentation for server.inject options to note that headers are lowercased.

[from]
headers - an object with optional request headers where each key is the header name and the value is the header content. Defaults to no additions to the default Shot headers.

[to]
headers - an object with optional request headers where each key is the header name and the value is the header content. Header names must be all lowercase in order to match the header names normalized by node. Defaults to no additions to the default Shot headers.


hueniverse:
You put the text in the wrong place.

fourq:
image

bobpaterno:
@hueniverse For future reference, exactly where should this text go?

hueniverse:
You added the comment about the wrong `headers`.

PR #3291, opened 2016-08-17
Ensure that 206 responses are never compressed

kanongil:
Since the `content-range` header relies on an unmodified entity response, we need to leave it alone.

### DIFF
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -211,6 +211,7 @@ internals.transmit = function (response, callback) {
     let compressor = null;
     if (encoding &&
         length !== 0 &&
+        response.statusCode !== 206 &&
         response._isPayloadSupported()) {
 
         delete response.headers['content-length'];
--- a/test/transmit.js
+++ b/test/transmit.js
@@ -725,6 +725,27 @@ describe('transmission', () => {
             });
         });
 
+        it('skips compression for 206 responses', (done) => {
+
+            const server = new Hapi.Server();
+            server.connection();
+
+            const handler = function (request, reply) {
+
+                return reply('test').code(206);
+            };
+
+            server.route({ method: 'GET', path: '/', handler });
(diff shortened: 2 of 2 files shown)


hueniverse:
I don't see anything in the spec that forbids applying content encoding to a 206 response.

@mnot any advice?

kanongil:
@hueniverse 206 + `content-encoding` is indeed perfectly valid, and Inert supports this for pre-compressed files.

However, the current implementation doesn't work for dynamic responses, as the `content-range` header is determined based on the non-encoded representation. Once it is encoded, it is no longer valid (length doesn't match).

Also, I dont believe the compressed stream is actually guaranteed to be idempotent, which it needs to be in order to reliably stitch a complete response together.

hueniverse:
Should the range be calculated based on the encoded or raw payload? I'm too busy to dig up the spec.

mnot:
It's buried a bit. In http://httpwg.org/specs/rfc7233.html#header.content-range - 

> The "Content-Range" header field is sent in a single part 206 (Partial Content) response to indicate the partial range of the **selected representation** enclosed as the message payload, sent in each part of a multipart 206 response to indicate the range enclosed within each body part, and sent in 416 (Range Not Satisfiable) responses to provide information about the selected representation.

Emphasis mine. "Selected representation is defined in http://httpwg.org/specs/rfc7231.html#representations - 

> An origin server might be provided with, or be capable of generating, multiple representations that are each intended to reflect the current state of a target resource. In such cases, some algorithm is used by the origin server to select one of those representations as most applicable to a given request, usually based on content negotiation. This "selected representation" is used to provide the data and metadata for evaluating conditional requests [RFC7232] and constructing the payload for 200 (OK) and 304 (Not Modified) responses to GET (Section 4.3.1).

So, Content-Range is specifying a range of the result of any content negotiation, including encoding.

hueniverse:
Thanks @mnot!

PR #3319, opened 2016-08-26
Allow custom encoders to validate compression options

kanongil:
While implementing support for the new route level `compression` option in a custom encoder registered using `server.encoder()`, I realized that the current implementation either requires me to validate the compression options on each request, or trust that they are correct.

Since neither option is desirable, I created this PR to extend the API with an optional `schema` to allow the validation logic to be applied only once, when the route is created.

Let me know what you think?

Note that this only validates at the route level, and not at the connection level, since the connection already exits when adding the encoder.


hueniverse:
This still fails to validate routes added before the encoder/decoder is added. I need to think about other options.

hueniverse:
I like the idea but this is not the right solution.

Issue #2516, opened 2015-04-28
Assert when server.start() is called before server.register() is done


hulbert:
As far as I can tell, `server.register` doesn't wait for the previous plugin to call the next callback. This is pretty confusing since `next()` seems to imply it's safe to run async code in a plugin that a later plugin would be dependent on.

garthk:
Figured out in IRC @hulbert had not waited for `server.register` to call back before calling `server.start`. Could we have `start` throw an exception if someone tries that?

hulbert:
@garthk talked me through some of this on IRC; as a "new user" I'm hoping to just offer some general feedback about what's confusing:

1. AFAICT doing something like `server.register({ plugin: require('plugin1') }, { plugin: require('plugin2') }, callback)` means that plugin2 will wait on plugin1's `next()` call.

2. Doing the following means plugin2 will *not* wait on plugin1's callback

```
server.register({ plugin: require('plugin1') });
server.register({ plugin: require('plugin2') });
```

3. To register a plugin with a prefix, you can't pass an array to `server.register` because the prefix applies to all the plugins you're registering.

4. Because of this, to do sequential registration of plugins that require different prefixes you essentially need to enter callback hell (or promisify `server.register` which is what I've done).

Now, some of this may not matter since route registration isn't really async as far as I can tell. However, hopefully you can see the confusion with how plugin registration currently works. @garthk mentioned using `server.dependency` but this isn't mentioned in the tutorials so I didn't notice its existence. It also is unclear if it would be recommended to declare dependencies in routing-related plugins which could depend on almost all plugins (or any plugins that behave like middleware, like auth, response manipulation, etc)

Marsup:
I believe most people will use `glue` if they're uncomfortable with what you're calling callback hell.

danielb2:
@garthk   https://github.com/hapijs/hapi/blob/master/lib/plugin.js#L154  hapi already checks to make sure the last argument is a callback

garthk:
The examples implicitly guide new users to `server = new Server`. If we want them to use `glue`, we should tell them that.

hueniverse:
Is this a matter of the `register()` callback being optional?

garthk:
Not quite. The callback isn't optional, at least in 8.4 and 8.5: we assert `typeof callback === 'function'` in `Plugin.register`.

The problem is:

* We assume the user will wait for `server.register` to call back before attempting `server.start`
* The documentation and tutorials do not encourage them to do so
[...]

PR #3333, opened 2016-09-03
Fix flaw in tracking of plugin registration status

jmm:
Per #2516, `server.start()` is supposed to error when plugin registrations haven't completed. It's tracked with a boolean, which means that a later `server.register()` can clobber the value. Example:

```js
// Will work as intended:
server.register(syncPlugin);
server.register(asyncPlugin);
server.start();

// Won't:
server.register(asyncPlugin);
server.register(syncPlugin);
server.start();
```

This PR updates the relevant tests to cover that and changes the status variable to a counter to fix it.

Note that the basic relationship between `server.register()` and `server.start()` still lends itself to race conditions. It's not hard to get into a situation where things appear to be setup ok, but then sometimes some other aysnc prerequisite to calling `server.start()` will finish before plugin registration and `server.start()` will error. Example:

```js
// register some plugins

// When plugin registration wins the race, everything seems to be ok. When
// somethingAsync() does, server.start() will error.
somethingAsync()
.then(function () {
  return server.start();
});
```

You can keep track of it yourself, of course (that's what I'm doing now), but it's kind of a pain. I think it'd be nice to have something like this:

```js
// Promise that settles based on all plugin registrations that have already
// happened (maybe make additional calls to server.register throw after this):
server.ready()
.then(function () {
  return server.start();
})
```

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -244,7 +244,7 @@ internals.Plugin.prototype.register = function (plugins /*, [options], callback 
         registrations.push(registration);
     }
 
-    this.root._registring = true;
+    this.root._registering++;
 
     const each = (item, next) => {
 
@@ -319,7 +319,7 @@ internals.Plugin.prototype.register = function (plugins /*, [options], callback 
 
     Items.serial(registrations, each, (err) => {
 
-        this.root._registring = false;
+        this.root._registering--;
         return Hoek.nextTick(callback)(err);
     });
 };
--- a/lib/server.js
+++ b/lib/server.js
@@ -50,7 +50,7 @@ exports = module.exports = internals.Server = function (options) {
     this._decorations = {};
     this._plugins = {};                                                             // Exposed plugin properties by name
     this._app = {};
-    this._registring = false;                                                       // true while register() is waiting for plugin callbacks
+    this._registering = 0;                                                           // > 0 while register() is waiting for plugin callbacks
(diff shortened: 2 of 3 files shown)


jmm:
I don't think the CI faiure is related to this PR. I was also getting some flaky results locally on master.

Marsup:
I just re-ran CI on master and your PR, it's working on master, more likely your PR is causing that.

hueniverse:
It's your job to serialize `server.register()` calls. They must be called one at a time.

jmm:
@hueniverse You should document that. I don't see anything in the docs that says you have to wait for each plugin registration to finish before starting any other.

@Marsup It's a moot point now for this PR, but I get intermittent test failures locally on master. I just ran it 6 times in a row and had 4 passes and 2 fails, including some of the same errors in the Travis builds:

```
499) Request _execute() handles aborted requests:

    Timed out (3000ms) - Request _execute() handles aborted requests
```

I can also run it on my PR branch and get passes with 100% code coverage.

devinivy:
For what it's worth, I like the spirit of this PR.  At minimum we should update the docs.

jmm:
Thanks @devinivy.

mk-pmb:
> @hueniverse You should document that. I don't see anything in the docs that says you have to wait for each plugin registration to finish before starting any other.

:+1: (or document here that the docs have been updated.)

If I'd read that requirement in the docs, my 1st thought would be to make a promise chain from my list of desired extensions. My 2nd thought would be why I'd have to do that in my code, considering the mission statement of Hapi.

PR #3357, opened 2016-09-29
Invoke reply with promise returned from handler.

JorgenEvens:
This allows a promise to be returned from a handler which will then
automatically be used to invoke `reply`. This way the server responds as
soon as the promise resolves or rejects.

This cleans up the code in handlers that uses promises significantly.

Simple example:

```
function handle (request, reply) {

    const promise = Promise.resolve(request.params.id);

    reply(promise);
};
```

Becomes:

```
function handle (request, reply) {

    return Promise.resolve(request.params.id);
};
```

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -93,7 +93,11 @@ internals.handler = function (request, callback) {
 
     // Execute handler
 
-    request.route.settings.handler.call(bind, request, reply);
+    const promise = request.route.settings.handler.call(bind, request, reply);
+
+    if (typeof promise === 'object' && typeof promise.then === 'function') {
+        reply(promise);
+    }
 };
 
 
--- a/test/handler.js
+++ b/test/handler.js
@@ -130,6 +130,27 @@ describe('handler', () => {
                 done();
             });
         });
+
+        it('calls reply with returned promise', (done) => {
+
+            const response = {};
+
+            const server = new Hapi.Server();
+            server.connection();
+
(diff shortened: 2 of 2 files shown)


vdeturckheim:
Isn't it already working ?

```js
function (request, reply) {

        const prom = Promise.resolve('hello world');
        return reply(prom);
    }
```
works for me

JorgenEvens:
Sure that works, but adding this would allow you to simply return the promise from the handler which is a lot cleaner when your promise is a chain of small `.then` steps.

vdeturckheim:
Oh my bad, I misunderstood the PR. sorry.
Removing my comment :/

hueniverse:
I am not a fan.

First, if we were to support this, we need to apply it consistently across all `reply` interface access. This includes extension points, auth methods, handlers, and prerequisites. That seems pretty intrusive to me in an area of the code that is already pretty delicate. It also adds a wasteful condition after every one of those methods, just in case a promise has been returned.

Second, this is a limited API. Yes, if your handler is simple and just returns a plain response, this will work. But as soon as you want to set custom headers, response code, etc. this API will not work. This means you will still need to use the explicit `reply(promise)` interface.

I appreciate the effort and the idea behind it, but this is incomplete at best. Before you go and put more effort into it, it would be good to hear from others who are likely to use this and see how they currently use the complexity of the `reply` interface.

PR #3358, opened 2016-10-03
fix server not propagating errors on prehandler(promise) + handler error (#3242)

### LINKED ISSUE #3242: Using promises in prehandlers causes exceptions to be swallowed in handlers.
I have the following test case, when you send a request to the server it will hang. It doesn't respond, and nothing is logged to the console (this cost me hours of debugging earlier :/)

``` javascript
const Hapi = require('hapi');

server.connection({ host: 'localhost', port: 8999 });

const pre = (request, reply) => {
  return reply(Promise.resolve('FROM PRE HANDLER'))
};

server.route({
    method: 'GET',
    path:'/', 
    config: {
      pre: [{ method: pre, assign: 'test' }]
    },

        throw new Error('this is an error ' + request.pre.test)
        return reply('hello world ' + request.pre.test);
    }
});

server.start((err) => {
    if (err) { throw err; }
    console.log('Server running at:', server.info.uri);
});
```

@nlf, I saw this which might be related https://github.com/hapijs/hapi/issues/3102 but the fix https://github.com/hapijs/hapi/commit/8babca3ef9271ede738b0377d628e62c94d24f04 doesn't cover this either. I'm trying to debug this further and provide a fix, but I'm not a promises expert, finding it tough to get to the bottom of it so maybe someone might be able to help!

This is with hapi@13.5.0 on Node 6.

turtleDev:
This patch proposes a fix for #3242 ; 

The fix revolves around explicit domain bindings and execution of promise handler from the event loop (using process.nextTick) so that the error is not swallowed up by the promise itself (which was basically leading to a deadlock; the promise never fully resolved, and the callback is never called )

This is technically a bugfix, and I wasn't sure if I'm supposed to increase the patch version number, so I'll leave it up to you guys.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -455,17 +455,22 @@ internals.Response.prototype._prepare = function (next) {
 
     const onDone = (source) => {
 
-        if (source instanceof Error) {
-            return next(Boom.wrap(source));
-        }
+        // do not use Hoek.nextTick because it will not work as intended
 
-        if (source instanceof internals.Response) {
-            return source._processPrepare(next);
-        }
+        process.nextTick(() => {
 
-        this._setSource(source);
-        this._passThrough();
-        this._processPrepare(next);
+            if (source instanceof Error) {
+                return next(Boom.wrap(source));
+            }
+
+            if (source instanceof internals.Response) {
+                return source._processPrepare(next);
+            }
+
+            this._setSource(source);
+            this._passThrough();
+            this._processPrepare(next);
+        });
     };
 
     const onError = (source) => {
--- a/lib/response.js
+++ b/lib/response.js
@@ -453,25 +453,20 @@ internals.Response.prototype._prepare = function (next) {
         return this._processPrepare(next);
     }
 
-    const onDone = (source) => {
+    const onDone = Hoek.nextTick((source) => {
 
-        // do not use Hoek.nextTick because it will not work as intended
+        if (source instanceof Error) {
+            return next(Boom.wrap(source));
+        }
 
-        process.nextTick(() => {
+        if (source instanceof internals.Response) {
+            return source._processPrepare(next);
+        }
 
-            if (source instanceof Error) {
-                return next(Boom.wrap(source));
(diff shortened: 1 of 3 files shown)


johnbrett:
Can confirm this fixes #3242 - Thank you for this @turtleDev!

turtleDev:
@johnbrett it was a pleasure :smile:

Marsup:
Looks very similar to the previous PR, doesn't exit the server anymore @johnbrett ?

lib/response.js
```diff
@@ -455,17 +455,22 @@ internals.Response.prototype._prepare = function (next) {
 
     const onDone = (source) => {
 
-        if (source instanceof Error) {
-            return next(Boom.wrap(source));
-        }
+        // do not use Hoek.nextTick because it will not work as intended
```
nlf: hoek's nextTick drops context so `this` would become invalid in this case

nlf: that's a good point, what didn't work for you here @turtleDev?

johnbrett:
Yep - very similar @Marsup. It seems to be be that not using `Hoek.nextTick` is the only difference, I don't really understand why it makes the difference though :/

turtleDev:
@johnbrett `Hoek.nextTick` will work too, see https://github.com/hapijs/hapi/pull/3358#discussion_r81996518

lib/response.js
Marsup: See https://github.com/hapijs/hapi/pull/3202/files#diff-d86a59ede7d999db4b7bc43cb25a1c11R399, it doesn't have to be weird :)

turtleDev: ahh yes, I realized this after I updated my comment. Thanks for pointing that out.

turtleDev:
@johnbrett @Marsup Hello guys! Any updates on when this PR will be reviewed and /or merged?

Marsup:
I'm 👍 on this, we're not the ones to convince, @hueniverse is.

Issue #3378, opened 2016-11-12
Server fails to reuse Boom object


wwalser:
I am re-creating this issue which was opened https://github.com/hapijs/boom/issues/140 which is more likely an issue inside Hapi. I've seen the same behavior on an older vertion of Hapi and Boom (see original issue for my comment).

------- pasted issue -------

I recently started a new project and since I have used Hapi before and was happy with it, I used it again.

I have everything modularized, nice and pretty. The project includes a handlers folder with files that each contains handlers for the same resource. For example, in the same file, I would have a handler to get a resource by its ID and another one that would update a resource by its ID. Since I want handlers of the same resource to return the same errors (in the previous example, an example would be a Boom.notFound('Resource was not found');), I store them in a const and use the const where needed.

In my previous project, this gave me no issue, but it does now. The issue is that the server becomes unresponsive after the second call. Trying to find the source of the problem, I came up with the following code that reproduces the issue:

```javascript
'use strict';

const Boom = require('boom');
const Hapi = require('hapi');

const ERROR_TEST = Boom.forbidden('Error Test');

const server = new Hapi.Server();
server.connection({ port: 3000 });

server.route({
    method: 'GET',
    path: '/',
    handler: function (request, reply) {
        reply(ERROR_TEST);
    }
});

server.start((err) => {

    if (err) {
        throw err;
    }
    console.log(`Server running at: ${server.info.uri}`);
});
```

With this code, I call the server and I get an error response, as expected. If I call it again, there's no response from the server. If I call it a third time, there's no response from the server. All subsequent calls return no response.

If I replace the const for its value, everything works, so I have been changing my code accordingly, but I know that in the previous project the server worked without the replacement. This indicates me that there was a bug introduced in later versions of either hapi or more likely, boom.

The problem is not solely with Boom.forbidden(). I have tried the same code with Boom.badRequest(), Boom.unauthorized(), Boom.paymentRequired(), and Boom.notFound(). I'm guessing that the other errors will behave similarly.

## Old Project

node version: v6.0.0
hapi version: ^13.5.0
boom version: ^3.2.1
os: Debian Jessie
Inside a Docker Container

## New Project

node version: v6.0.0
hapi version: ^15.0.3
boom version: ^4.0.0
os: Debian Jessie
Inside a Docker Container
Test Code in this issue

node version: v6.0.0
hapi version: ^15.2.0
boom version: ^4.2.0
os: OS X El Capitán
Outside a Docker Container

[...]

PR #3380, opened 2016-11-16
Ensure Boom objects can be reused - Fix for issue #3378

wwalser:
Fix for #3378 Let me know when it's all clear and I'll squash.

Cheers,
Wes

### DIFF
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -110,11 +110,11 @@ internals.marshal = function (request, next) {
 
 internals.fail = function (request, boom, callback) {
 
-    const error = boom.output;
+    const error = Hoek.clone(boom.output);
     const response = new Response(error.payload, request);
     response._error = boom;
     response.code(error.statusCode);
-    response.headers = Hoek.clone(error.headers);
+    response.headers = error.headers;
     request.response = response;                            // Not using request._setResponse() to avoid double log
 
     internals.marshal(request, (err) => {
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -114,7 +114,7 @@ internals.fail = function (request, boom, callback) {
     const response = new Response(error.payload, request);
     response._error = boom;
     response.code(error.statusCode);
-    response.headers = error.headers;
+    response.headers = Hoek.clone(error.headers);
     request.response = response;                            // Not using request._setResponse() to avoid double log
 
     internals.marshal(request, (err) => {
--- a/test/transmit.js
+++ b/test/transmit.js
@@ -1400,6 +1400,49 @@ describe('transmission', () => {
             });
         });
 
+        it('boom object reused does not affect encoding header.', (done) => {
+
+            const error = Boom.badRequest();
+            const data = JSON.stringify({
+                statusCode: error.output.statusCode,
+                error: error.message
+            });
+
+            const server = new Hapi.Server();
+            server.connection();
+
+            const handler = function (request, reply) {
+
+                return reply(error);
+            };
+
+            server.route({ method: 'GET', path: '/', handler });
+            server.start((err) => {
+
+                expect(err).to.not.exist();
+
+                const uri = 'http://localhost:' + server.info.port;
+
+                Zlib.gzip(new Buffer(data), (err, zipped) => {
+
+                    expect(err).to.not.exist();
+
+                    Wreck.get(uri, { headers: { 'accept-encoding': 'gzip' } }, (err, res, body) => {
+
+                        expect(err).to.not.exist();
+                        expect(body.toString()).to.equal(zipped.toString());
+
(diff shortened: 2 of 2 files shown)


wwalser:
Added an additional regression test based on comments in the issue's discussion thread, see #3378.

AdriVanHoudt:
I think this is correct unless @hueniverse knows of a better fix.
Only thing I worry about might be perf with all the cloning

kanongil:
I'm confused – why should this be supported? Boom objects are not designed for reuse.

Marsup:
Same for me, I feel that it's a bad pattern. Pre-generating errors won't improve your performance by any means (if that's the intention), and you'll lose a meaningful way of debugging, that is stacktraces. I'd favor error factories for that.

AdriVanHoudt:
I can agree on that, I wouldn't do it either but the way hapi handles it now does not feel correct either

devinivy:
I agree with both sides– ideally I'd like to see hapi deal with this better, but booms are _actual_ `Error`s, which shouldn't be reused.

arb:
`clone` also isn't cheap and should only be used as a last ditch.

wwalser:
I was concerned as well but having read Boom more carefully I think the `.output` objects are so shallow and simple that it shouldn't be a problem.

Here is an example of common content of `.output` at the time that we clone it based on Boom's implementation of `.wrap` and `.initialize`:
```javascript
    error.output = {
        statusCode: 500,
        payload: {
            message: 'An internal server error occurred'
        },
        headers: {}
    };
```

AdriVanHoudt:
What is the reason the headers are copied? For errors like unauthorized right?
If so another solution could be to handle the special content-encoding case.

I do agree that hapi should not modify (boom) errors you give it.

devinivy:
I'm not first and foremost concerned with performance here.  The reason I'm not into supporting reuse is that boom is an `Error`, and thus holds information like `boomErr.stack` that is absolutely particular to moment that the error occurs.  `Error`s in general aren't good candidates for reuse.  If you want reusable const-style errors, consider adding a reply decoration that accepts a reusable error code.

At the same time, I do see how hapi hanging upon error reuse is very confusing and isn't ideal.

wwalser:
> What is the reason the headers are copied?

Headers are added by two of Boom's errors, `unauthorized` and `methodNotAllowed`.

> At the same time, I do see how hapi hanging upon error reuse is very confusing and isn't ideal.

Hapi doesn't actually hang. It responds with headers that don't match the content causing clients to display no output because it can't decode it properly. [See my comment in the original thread](https://github.com/hapijs/hapi/issues/3378#issuecomment-260802252).

AdriVanHoudt:
@wwalser I meant that by 'hanging up' ^^ 

I think we all agree that the use case is an anti pattern that you should not be doing but is this a bug then or not?
It feels like hapi should play nice though :/

wwalser:
A few thoughts re: is this a bug?

1. Hapi's behavior of respecting an error's `content-encoding` and assuming the API user has already encoded their content thus returning a content body that can't be decoded is not a bug. However, because of how confusing this case can be, I would love to have an error, warning or _some_ indication that this is what is happening. Multiple people opened an issue believing this to be a bug and it took 3 people reading through the call path to uncover what was happening and why.
2. Reusing Boom's is and anti pattern. I feel compelled to point out that it is only an anti-pattern because the `Error` is instantiated upon _creation_ and not on _use_. It seems to me that this fact could in fact be the anti-pattern though I'm not sure. It is indeed very easy to use `const userNotFound = ()=>Boom.unauthorized('…', '…');` instead of `const userNotFound = Boom.unauthorized('…', '…')`.
3. The fact that mutation occurs still bothers me but I respect the fact that I lean toward immutability and purity whereas most of the JS community does not.

I feel that prescribing what should be done would be to step out of bounds as someone who has just shown up. I defer to those who have cared for and crafted Hapi for much longer than myself. Thanks for looking into the issue and PR :).

AdriVanHoudt:
1: agree
2: this is just how js errors work?
3: I would also prefer that hapi did not modify the error
4: As a hapi user you are very much entitled to an opinion, that is why we have this PR system :P But I also don't feel like I'm experienced enough to call this one :D

wwalser:
> 2: this is just how js errors work?
I only meant that there's no need for Boom to create it's internal Error until the Boom is actually being handled by Hapi. If it instead created the Error at the time of use instead of instantiation the stack traces would be correct regardless of the way that the Booms are created, referenced, copied etc.

AdriVanHoudt:
Well since Boom errors are actual errors but with more info I think that would be weird but I see where you come from

hueniverse:
1. You should not reuse errors.
2. hapi should not mess around with user responses, errors or otherwise.

This is not the right solution, but the tests are useful.

PR #3401, opened 2016-12-01
Attempt to fix #3399 where it crashes on route prerequisites when no domain is present

### LINKED ISSUE #3399: Regression on 16, lib/handlers.js
Hi everyone.

I'm experiencing a regression on 16. One of the routes on my hapi application won't respond. This route has a prerequisite without assignment and the error is triggered while still inside the pre.

The line that's causing problem is the 65th of lib/handler.js. The error says that domain is undefined, thus it cannot read it's bind property.

To fix the error, the only thing I did was change ^16.0.0 in my package.json to ^15.2.0, delete the node_modules and reinstall. Now everything works as it always did.

Sadly I can't give you access to my code and I don'T have the time at the moment to create a server which exhibits the bug. I'll try to do one soon.

Thank you.

sirgallifrey:
I wrote a failing test witch match the description on #3399 and added a check to see if domain exists before binding. I also changed `domain = process.domain` to `domain = request.domain` to prevent the test domain from being used.

I'm sorry in advance if this fix is no good, I just recently started to studying the hapi internals

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -54,7 +54,7 @@ internals.prerequisites = function (request, callback) {
         }, nextSet);
     };
 
-    const domain = process.domain;      // Save a reference to the current domain
+    const domain = request.domain;      // Save a reference to the current domain
 
     Items.serial(request._route._prerequisites, each, (err) => {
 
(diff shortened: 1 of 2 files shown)


hueniverse:
Don't apologize in advance (or after the fact for that matter) for making a mistake. I don't know if you have, I'm just saying in general. I appreciate the effort. I'll review this week.

simlevesque:
I'll be able to test this PR on my machine environement later today.

AdriVanHoudt:
@sirgallifrey yeah no need to say sorry, we appreciate every contribution very much!

simlevesque:
This PR solves the bug ! Not much else to say, it used not to work, now it does.

Thank you very much everyone.

PR #3406, opened 2016-12-13
Match response.send() docs to implementation

kanongil:


hueniverse:
I don't mind this change but resume doesn't imply transmit so it wasn't incorrect before.

Issue #3408, opened 2016-12-16
Safeguarding A Decorated Request Object


gordysc:
When dealing with a HapiJS server w/ multiple connections, there are no friendly safeguards to prevent plugins from decorating the request object multiple times.  For example:

```
exports.register = (server, options, next) => {
  server.decorate('request', 'decoratedFunctionName', .... );
}
```

As stated in HapiJS's documentation:

> Note that decorations apply to the entire server and all its connections regardless of current selection.

Since we are registering the plugin across multiple connections, the only way to get the above to work is using a try/catch block.  It'd be super useful if there was a mechanism to help check for an existing decoration.  Perhaps something as simple as a method that'd return an array of existing decorations on the request object, like so:

`server.inspect('request')`
and return:
`['decoratedFunctionName', ...]`

Related issues:
https://github.com/hapijs/discuss/issues/415
https://github.com/hapijs/hapi-auth-cookie/issues/119

AdriVanHoudt:
@gordysc you can access them via `server._decorations` but that is not a "public" api. Just exposing that would solve your case no?

gordysc:
@AdriVanHoudt Yes, that's exactly what I'd love to have.  I imagine others would benefit from this being public as well.

AdriVanHoudt:
I totally agree, if @hueniverse agrees a PR to make the decorations public is welcome!

gordysc:
I'll gladly submit the PR if he's on board.

AdriVanHoudt:
He might be more willing when the PR is there already 😉

gordysc:
We can close this issue out and I'll track progress in the referenced PR.

AdriVanHoudt:
@gordysc if you add `closes #3408` to your commit message github will auto link the commit and PR to this issue and when the PR gets merged the issue gets closed automatically :D

gordysc:
@AdriVanHoudt good to know! You learn something new everyday :)

closes #3409

PR #3409, opened 2016-12-17
Adding server.inspect to list decorated framework interfaces

gordysc:
Addresses enhancement request: https://github.com/hapijs/hapi/issues/3408

### DIFF
--- a/API.md
+++ b/API.md
@@ -5,6 +5,7 @@
     - [Server properties](#server-properties)
         - [`server.app`](#serverapp)
         - [`server.connections`](#serverconnections)
+        - [`server.decorations`](#serverdecorations)
         - [`server.info`](#serverinfo)
         - [`server.load`](#serverload)
         - [`server.listener`](#serverlistener)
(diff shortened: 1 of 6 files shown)


gordysc:
[coverage.pdf](https://github.com/hapijs/hapi/files/658558/coverage.pdf)

hueniverse:
I prefer this was much simpler. `server.decorations` as a property. Since you cannot un-decorate, keeping such a public property up to date should be trivial.

gordysc:
Agreed, will revisit this approach.

gordysc:
@hueniverse just for clarification, are you suggesting:
```
server.decorations
request.decorations
reply.decorations
```

Or
```
server.decorations = {
  server: <decorations>,
  request: <decorations>,
  reply: <decorations>
}
```

devinivy:
The latter makes more sense to me.

AdriVanHoudt:
@devinivy any particular reason why?

devinivy:
Yes– it's hard to get a hold of a request object or reply interface simply to check if they're decorated with a specific method.  Decorations are added at plugin-registration time, so I think they should be able to be introspected at plugin registration time as well.

AdriVanHoudt:
Ok I agree. Why not both? if you want to check on request you'd have to go `request.server.decorations.request[]` :P

hueniverse:
Just `server.decorations`.

devinivy:
@AdriVanHoudt I would omit that for the express purpose of keeping the features of the API independent of each other.  Plus, if you already have a `request` object you can just check for the decorated method directly on it!

gordysc:
@hueniverse is this approach more inline?

hueniverse:
I'll review it in a week or two.

PR #3418, opened 2017-01-05
support a header to override request.id

brettporter:
I'd like to pass a custom `request.id` value via a request header such as `X-Request-Id`, that enables request tracing across a series of internal service calls, instead of using the generated ID of each.

There are plugins to enable this, such as https://www.npmjs.com/package/hapi-request-id, however this is dependent on the order of the plugins if a different `onRequest` extension needs to use the `request.id`. These also don't take effect in time to be used by the initial `received` logging statement at the end of the `Request` constructor.

I've created a patch that covers this capability. I'm interested to hear if this makes sense, and if so whether there might be a better place than `connection.settings` to configure it. Another possibility I considered was whether to supply a function to generate the ID rather than using a header, but was unclear where that might belong.

Thanks!

### DIFF
--- a/API.md
+++ b/API.md
@@ -155,6 +155,9 @@ Creates a new `Server` object where:
            which is used to store configuration values and `connection.plugins` which is meant for
            storing run-time state.
 
+        - `requestIdHeader` - the name of a header that can be used to override the value of
+          `request.id` for a request
+
         - <a name="connection.config.router"></a>`router` - controls how incoming request URIs are
           matched against the routing table:
             - `isCaseSensitive` - determines whether the paths '/example' and '/EXAMPLE' are
--- a/API.md
+++ b/API.md
@@ -155,9 +155,6 @@ Creates a new `Server` object where:
            which is used to store configuration values and `connection.plugins` which is meant for
            storing run-time state.
 
-        - `requestIdHeader` - the name of a header that can be used to override the value of
-          `request.id` for a request
-
         - <a name="connection.config.router"></a>`router` - controls how incoming request URIs are
           matched against the routing table:
             - `isCaseSensitive` - determines whether the paths '/example' and '/EXAMPLE' are
@@ -936,6 +933,17 @@ exports.register.attributes = {
     name: 'example',
     connections: false
 };
+```
+
+### `server.requestIdGenerator(method)`
+
+Registers a custom generator for request IDs where:
+- `method` - the function that is invoked using the signature `function(request)` and returns an ID, or `null` if the default should be used
+
+```js
+server.requestIdGenerator((request) => {
+    return request.headers['x-request-id'];
+});
 ```
 
 ### `server.decoder(encoding, decoder)`
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -38,6 +38,8 @@ exports = module.exports = internals.Connection = function (server, options) {
     this.settings = options;                                                        // options cloned in server.connection()
     this.server = server;
 
+    this.generateRequestId = this._defaultRequestIdGenerator;
+
     // Normalize settings
 
     this.settings.labels = Hoek.unique(this.settings.labels || []);                 // Remove duplicates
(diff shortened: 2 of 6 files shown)


nlf:
from a security standpoint, i don't like this. allowing a user to override the request id is not only giving someone the opportunity to render your logs useless by sending the same request id with every single request, but is a potential attack vector (i.e. what happens if they send a request id in the header that's 2 million characters long? what if they use some whacky utf-8 characters to try to confuse the http parser?)

what's your use case that you're wanting different request ids?

brettporter:
Thanks for the quick responses!

The use case is to do simple distributed tracing of requests, for example:
* Client calls public facing service A, which generates a request ID as normal (it doesn't use the header)
* A calls internal service B, setting `X-Request-Id` to it's own `request.id`
* B uses the supplied request ID
* Both services generate `request.log` statements that log the same request ID, which can be used for searching events across all the services from a single client request.

@mtharrison the reason for changing the existing `request.id` is because it's already used in necessary places. `request.log` above passed through `good` or another logging library will log that ID. Ideally we do not want to duplicate the logic that checks for the header, then falls back to the `request.id`, across multiple places where it is used. We also log the `request-internal` `received` event, which is hard to intercept since it's part of the request constructor. 

Is there another alternative I might have missed?

@nlf excellent point. Would generating an ID conventionally if the supplied header didn't validate against certain criteria suffice? Say, `/^[A-Za-z0-9\-:]{20,200}$/`.

mtharrison:
> Is there another alternative I might have missed?

Personally, I'd probably want to keep the per-request ids visible in my logs. Otherwise by overwriting them you're somewhat obscuring the fact that this is a distributed transaction and making it look like it's all happening in a single process.

Another approach could be to just associate the different request hapi request ids with each other (either in the actual logs themselves) or by recording the relationship in a db or redis or alike.

Then the heavy lifting of filtering log entries when you're browsing your logs could be left up to which ever tools you're using for that purpose.

hueniverse:
The only thing I am willing to consider is a request id generation function config. You provide your own function that gets the raw node request object and based on that generate an id. The id generation would probably be blocking but I need to look at the actual code to see if we need to limit it.

brettporter:
Certainly happy to look at that @hueniverse. I'm looking at options, but do you have any suggestions about where that config would best live?

hueniverse:
In the routes config, top level `id` at the same level `payload` and `validate` are.

brettporter:
I'd prefer to use the routes config, however it appears that the `request.id` is set prior to routing in the request constructor - `this.route` refers to the `notFound` stub route for settings initially when the ID is generated. I can't see a way to use that without moving the `request.id` code.

I initially opted for a server method which I've pushed here. This is used across all the connections, which looks more like decorate, encoder, etc. It does seem rather heavy handed though.

Would something along these lines work, or perhaps a connection setting again, or is there another approach to the routes config?

Thanks!

AdriVanHoudt:
Most other distributed systems use a custom header that they pass along.
You can easily write a transform stream for good that adds the header to your logs btw.

API.md
```diff
@@ -935,6 +935,17 @@ exports.register.attributes = {
 };
 ```
 
+### `server.requestIdGenerator(method)`
+
+Registers a custom generator for request IDs where:
+- `method` - the function that is invoked using the signature `function(request)` and returns an ID, or `null` if the default should be used
+
+```js
+server.requestIdGenerator((request) => {
```
hueniverse: Why a method? Why not just a route config option?

brettporter: Hi @hueniverse - my additional push crossed paths with your message suggesting this. I attempted to address this in this comment. Would appreciate any thoughts on how I could make the route config work, as I'd also prefer that.

hueniverse: Probably not getting to review this again for another week or two.

hueniverse: So why not put it in the connection config then?

hueniverse:
Idle for too long.

PR #3420, opened 2017-01-07
[update] license to 2017

josephlin55555:


hueniverse:
Technically, this should only happen once there is a commit in 2017.

AdriVanHoudt:
@hueniverse there is one now ;) https://github.com/hapijs/hapi/commit/a1fbae43a0a4199c00d034f0860d5f7d4f37eea9

Issue #3430, opened 2017-01-26
Allow setUrl to overwrite host


rubennorte:
## Context

At the moment, when re-routing a request internally (`onRequest`) only the new path is being set. `request.info.host` is unmodified by this call and preserves the original value. This prevents using the `vhost` route config option for re-routed requests.

My use case is to implement URLs with a dynamic component in their subdomains (e.g. `<dynamic>.en.example.com/foo`), re-routing them to static domains and dynamic paths (e.g.: `en.example.com/<dynamic>/foo`) because the call router does not currently support it. 

My current fix is to manually reset `request.info.host` and `request.info.hostname` with the values in the parsed URL passed to `request.setUrl`.

## Current behaviour

Current URL: http://dyn.en.example.com

```
request.setUrl('http://en.example.com/dyn');
// request.path => '/dyn'
// request.info.host => 'dyn.en.example.com'
// request.info.hostname => 'dyn.en.example.com'
```

## Expected behaviour

Current URL: http://dyn.en.example.com

```
request.setUrl('http://en.example.com/dyn');
// request.path => '/dyn'
// request.info.host => 'en.example.com'
// request.info.hostname => 'en.example.com'
```

rubennorte:
I also noticed that `request.url` omits the information about the host and the protocol. Would you accept a PR changing that for both original requests and re-routed requests (using `setUrl`)?

## Current behaviour

Current URL: http://dyn.en.example.com/path?foo=bar

```javascript
request.url = {
  auth: null
  hash: null
  host: null
  hostname: null
  href: "/path?foo=bar"
  path: "/path?foo=bar"
  pathname: "/foo"
  port: null
  protocol: null
  query: { foo: "bar" }
  search: "?foo=bar"
  slashes: null
}
```

## Expected behaviour

Current URL: http://dyn.en.example.com/path?foo=bar

```javascript
request.url = {
  auth: null
  hash: null
  host: "dyn.en.example.com"
  hostname: "dyn.en.example.com"
  href: "http://dyn.en.example.com/path?foo=bar"
  path: "/path?foo=bar"
  pathname: "/foo"
  port: 80
  protocol: "http:"
  query: { foo: "bar" }
  search: "?foo=bar"
  slashes: null
}
```

hueniverse:
Sure.

PR #3438, opened 2017-02-15
3430 causes setUrl to update host, hostname, port and protocol fields…

dpenny52:
… in request object

Referencing this issue: https://github.com/hapijs/hapi/issues/3430


hueniverse:
I prefer the solution in #3441. Thanks.

Issue #3431, opened 2017-01-30
when failAction is log or ignored, the error is assigned to request.preResponse but not request.pre


wbcustc:
```js
// Route config
{ 
    ...
    pre: [
         { method: 'serializeToS3()', assign: 's3Error', failAction: 'log' },
    ],
}
```
In the code above, the s3Error is assigned to request.preResponse but not request.pre, I wonder that why in the doc, it just states that the error is assigned but not mention where it is assigned to.

Solution:
Update the hapi doc to explicitly told that the error is assigned to request.preResponse 
* *hapi version*: 16.0.3

sirgallifrey:
is this just a documentation thing or does this line needs fixing?

[lib/handler.js#L280](https://github.com/hapijs/hapi/blob/master/lib/handler.js#L280)

should it be something like this?

```javascript
request.pre[pre.assign] = response instanceof Error ? response : response.source;
```

I mean, it doesn't seems that the combination of `log` and `assign` are leaving the response out of request.pre on purpose

wbcustc:
I finally moved out from assigning the `Error` object. I felt like it's not a good practice to do that.
So I chose to use `onPreResponse` extension to handle all the errors.

But I think for documentation, we definitely want to mention the error assigning behavior.
Also, I think your change is good too.

sirgallifrey:
IMO using assign should always set the response on `request.pre`, it is more predictable that way. But of couse, if this is not a bug and this behavior is intended we should put it on the docs definitely!

Squawk09:
@sirgallifrey  So the response.source is always assigned to request.pre and the actual response is always assigned to request.preResponses? 

What is the benefit of request.preResponsives in the first place? 

Sorry if this is a silly question, I am still learning all of this.

sirgallifrey:
@Squawk09 You don't have to say sorry for asking a question!

> So the response.source is always assigned to request.pre and the actual response is always assigned to request.preResponses?

Yes, we should have the response.source on `request.pre` and the actual response object on `request.preResponse` but as we seen in the issue this is not working well with errors.

> What is the benefit of request.preResponsives in the first place?

With `request.pre` you only get the payload, with suits most use cases

```javascript
//prerequisite 
reply({ something: 'banana' });

// later on handler

console.log(request.pre[nameIAssignedIt]);  // {something:'banana'}
```

[...]

PR #3439, opened 2017-02-15
Issue 3431

Squawk09:
I'm pretty sure this is what you are wanting added to the guide.  

Thanks1


sirgallifrey:
ref: #3431

sirgallifrey:
Hello @Squawk09, thanks for the collaboration! I've expressed a concern on the issue. In my opinion it's a bug and not a documentation problem, but I might be wrong.

hueniverse:
This is incorrect.

PR #3441, opened 2017-02-21
Fix several request.url property issues on setUrl()

kanongil:
This PR fixes this following issues in the `request.setUrl()` method:

 * The passed url object is not cloned.
 * Trailing slash stripping should be done after normalization (important if hapijs/call#29 is implemented).
 * Always update pathname derived url properties when path is modified.

This should make for a safer and more consistent `request.setUrl()` experience.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -198,19 +198,32 @@ Hoek.inherits(internals.Request, Podium);
 
 internals.Request.prototype._setUrl = function (url, stripTrailingSlash) {
 
-    this.url = (typeof url === 'string' ? Url.parse(url, true) : url);
-    this.query = this.url.query;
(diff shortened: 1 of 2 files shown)


kanongil:
I added another commit to fix the issues in #3430.

hueniverse:
Why are you preserving the port number when the new host doesn't have one?

kanongil:
I'm not sure. It must have seemed sensible at the time, but using the default port for the scheme also makes sense. This way it is more similar to a redirect.

hueniverse:
If you are going to provide a new full URI, it should override everything including the port.

kanongil:
Ok, I have rebased and revised the info update.

PR #3443, opened 2017-02-24
Customize validation error code

defrag:
This feature would allow set up the default validation error code, which i find to be beneficial for some users (like myself). Im not sure if I put the option in the right place, since
i have been just using Hapi.js from yesterday. If you think the improvement would be beneficial, feel free to guide me where would be the best place to put the option.

Thanks

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -70,7 +70,8 @@ exports.connection = {
             server: false                           // Determines how long to wait for server request processing. Disabled by default
         },
         validate: {
-            options: {}                             // Joi validation options
+            options: {},                            // Joi validation options
+            errorStatusCode: 400                    // Default status code returned when validation failed
         }
     }
 };
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -175,7 +175,8 @@ internals.routeBase = Joi.object({
             Joi.func()
         ],
         errorFields: Joi.object(),
-        options: Joi.object()
+        options: Joi.object(),
+        errorStatusCode: Joi.number()
     })
 });
 
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -80,7 +80,12 @@ internals.input = function (source, request, next) {
 
         // Prepare error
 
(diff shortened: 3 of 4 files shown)


sirgallifrey:
There are other ways to achieve this, here is an example:

```javascript
const Hapi = require('hapi');
const Joi = require('joi');
const Boom = require('boom');

server.connection({
    routes: {
        validate: {
            failAction: function (request, reply, source, error) {

                if (source === 'query') {

                    return reply(Boom.badData(error.message));
                }
            }
        }
    } 
});

server.route({
    method: 'GET',
    path: '/',

        return reply('ok');
    },
    config: {
        validate: {
            query: {
                a: Joi.number()
            }
        }
    }
});

server.inject('/?a=abc', (res) => {

    console.log(res.statusCode);
    console.log(res.payload);
});
```

defrag:
@sirgallifrey Yes that will work, but it will miss all the structured error output that is composed in the postValidate action. If you think this is something that is worth customizing for end users, i'll gladly work on it. If not you can just close the PR and i can just override in my project :)

sirgallifrey:
> If you think this is something that is worth customizing for end users, i'll gladly work on it. If not you can just close the PR and i can just override in my project :)

@defrag Don't get me wrong, I'm not the one who will accept or reject your PR, I'm just a regular Hapi user trying yo offer a different perspective and share of ideas :)

And by the way, thanks for contributing! :smile: 

> Yes that will work, but it will miss all the structured error output that is composed in the postValidate action. 

I'm not sure of what is the `postValidate`, but I'm assuming you are talking about the resulting joi error properties, if that is the case changing my example like this will result in an error object just like the original, except that it will have a different statusCode and in its payload, the error string will be different (based on the satusCode), the Joi message will remain the same.

```javascript
failAction: function (request, reply, source, error) {

                if (source === 'query') {

                    return reply(Boom.create(422, error.message, error.data));
                }
            }
```

hueniverse:
The example from @sirgallifrey is the way to address this.

PR #3446, opened 2017-02-28
remove extra closure

cjihrig:
Closes https://github.com/hapijs/discuss/issues/433

### DIFF
--- a/lib/protect.js
+++ b/lib/protect.js
@@ -25,10 +25,7 @@ exports = module.exports = internals.Protect = function (request) {
     Domain = Domain || require('domain');
 
     this.domain = Domain.create();
-    this.domain.on('error', (err) => {
-
-        return this._onError(err);
-    });
+    this.domain.on('error', this._onError.bind(this));
 };
 
 


Marsup:
Care to elaborate why this would fix the situation ? This seems equivalent.

cjihrig:
The error handler was capturing the things that were reported as leaking in https://github.com/hapijs/discuss/issues/433. By removing the arrow function, the closure goes away, and the memory associated with those things can be released.

Marsup:
JS is so stupid sometimes... Good catch !

AdriVanHoudt:
So this was an actual leak?

sjuchno:
It only affects you if you are creating and removing many servers. If you are creating a new server per test then those were never released from memory. There is no issue in production.

AdriVanHoudt:
@sjuchno makes sense thanks!

arb:
Can someone explain this in more detail? Is it because `request` had a reference to `protect` and this extra closure wouldn't let the ref to the `request` object get GCed?

cjihrig:
I made a mistake. I made the change to `this.domain.on('error', this._onError)` and verified in the heap snapshot that the leak was gone. However, when I ran the tests, there were failures, and I had to add the `bind(this)`. This means that the `domain` still has a reference to all of the hapi objects, and the leak is still there.

The problem is that there are 193 `domain` objects left around. Most are from lab, but some from `protect.js` are getting out. From the heap snapshots, it looks like it happens through a list of `Immediate` objects (`setImmediate()` in Node core makes them) that connect to lab domains. Still need to do more digging.

cjihrig:
It looks like lab is adding a bunch of `Immediate`s to its domain [here](https://github.com/hapijs/lab/blob/master/lib/runner.js#L600-L605). Some of those `Immediate`s have `_idlePrev` pointers to other `Immediate` objects that were created by hapi, whose `domain` pointer references the domain created in `protect.js`. lab never removes the `Immediate`s that it adds, so they stay [here](https://github.com/nodejs/node/blob/master/lib/domain.js#L193), along with the references to all of the hapi stuff. Adding a couple `activeDomain.remove(immed);` [here](https://github.com/hapijs/lab/blob/master/lib/runner.js#L592-L596) seems to let the hapi domain references go.

@geek what do you think?

geek:
@cjihrig sounds like a good change to me... we didn't bother with the removal likely because the process is so short lived with lab

hueniverse:
Should this be revered in hapi?

cjihrig:
There's no harm in writing it like this. It still saves on an unnecessary closure.

PR #3448, opened 2017-03-01
Bind request continuation methods to current domain

kanongil:
Fix for https://github.com/hapijs/discuss/issues/445.

This ensures that any built-in request continuation methods are called using the correct domain. This is needed since promises lose the domain binding when eg. `reply()` is called inside a `.then()`.

### DIFF
--- a/lib/reply.js
+++ b/lib/reply.js
@@ -64,13 +64,32 @@ internals.Reply.prototype.decorate = function (property, method) {
 
 internals.Reply.prototype.interface = function (request, realm, options, next) {        // next(err || response, data);
 
-    const reply = (err, response, data) => {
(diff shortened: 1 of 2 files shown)


kanongil:
Note that this doesn't fix it completely for extended methods, like inert's `reply.file()`, though it still helps, since it eventually calls `reply.response()`.

devinivy:
Is there a specific reason the same fix for decorations is omitted (as you've mentioned)?

kanongil:
Mostly I only thought about it after I submitted the PR. Secondly, there is no way to know if the decoration is a continuation method, though we could just always bind the method to the domain to be safe.

Let me know if you want me to update the PR.

PR #3449, opened 2017-03-02
Change 2011-2016 to 2011-present 

Yahkob:
avoid future trivial copyright change PRs by setting static year to present


hueniverse:
Nah. I like editing these by hand.

Issue #3290, opened 2016-08-16
500 statusCode, next connection emits clientError HPE_CB_headers_complete


lostthetrail:
When running Node 6 and the latest HAPI.js, a second connection will hang and emit HPE_CB_headers_complete if the previous connection results in a 500 statusCode.

Here is a reproduction repo. Instructions are in the readme.
https://github.com/lostthetrail/hapi-node6-hang

I have been attempting to dig into the code to see if I can determine/solve the problem, but I am coming up a bit short and would like some help determining the problem.

lostthetrail:
Related to: https://github.com/hapijs/discuss/issues/315

devinivy:
I mentioned in that discussion above that I couldn't reproduce the hang (though I did see the `HPE_CB_headers_complete`).  Could it be client-dependent?  It would be easier to diagnose if the hang itself were demonstrable in node (by scripting requests to the server).  Or A cURL could help too.

lostthetrail:
It happens on all browsers, but not cURL for me.

lostthetrail:
This appears to be only happening on my machine. Going to close it out.

devinivy:
@lostthetrail I'm now experiencing the hang (so far I think it's related to this same problem) when using HTTPS but not HTTP.  Does that ring a bell?

PR #3461, opened 2017-03-21
explicitly destroy sockets on clientError

nlf:
node versions >= 6 no longer destroy sockets by default when a clientError listener is attached, we have to do so ourselves


devinivy:
Good catch.  Is that well-documented anywhere?  I wonder if it will help fix server hangs, i.e. in #3290.

nlf:
kind of? https://nodejs.org/api/http.html#http_event_clienterror click the little "history" arrow

nlf:
and no, the destroy() method returns early if the socket is already destroyed

PR #3472, opened 2017-04-11
Send connection close when there is unconsumed payload

kanongil:
With `payload.output === 'stream'`, there is no guarantee or requirement that the user consumes the incoming payload. When a response is sent with an unconsumed payload, the connection needs to be closed to signal that the payload has been discarded. This is best done by adding the `connection: close` header to the response.

Note that node.js clients might not actually be able to handle such a response right now, due to https://github.com/nodejs/node/issues/12339.


Marsup:
The test on this PR is flaky. I get a 400 `Payload content length greater than maximum allowed: 11534336` about half the time. Looking at the test I don't see how that's possible, any idea ?

PR #3475, opened 2017-04-13
Do not use raw request in api docs example

AdriVanHoudt:
This can cause weird issues when you rewrite the url in plugins and then rewrite it again in example to the raw url.


wswoodruff:
Seems legit, don't think this will bring any breaking changes to Hapi

PR #3479, opened 2017-04-18
Updates to 16.1.1 API docs

AJamesPhillips:
Some of the changes are obviously minor but some of them are significant and I'm not certain about all of them.  I've tried to parse the docs, example code, source code and tests but there may still be errors with these suggested edits.  Please tell me if so and I'll happily update them.

### DIFF
--- a/API.md
+++ b/API.md
@@ -88,6 +88,9 @@
     - [`reply.continue([result])`](#replycontinueresult)
     - [`reply.close([options])`](#replycloseoptions)
     - [`reply.redirect(uri)`](#replyredirecturi)
+    - [`reply.response(result)`](#replyresponseresult)
+    - [`reply.state(name, value, [options])`](#replystatenamevalueoptions)
+    - [`reply.unstate(name, [options])`](#replyunstatenameoptions)
 
 ## Server
 
@@ -113,16 +116,7 @@ Creates a new `Server` object where:
       cache configuration only defines the storage container itself. `cache` can be assigned:
         - a prototype function (usually obtained by calling `require()` on a **catbox** strategy
           such as `require('catbox-redis')`). A new **catbox** [client](https://github.com/hapijs/catbox#client) will be created internally using this function.
-        - a configuration object with the following options:
-            - `engine` - a prototype function or **catbox** engine object.
-            - `name` - an identifier used later when provisioning or configuring caching for
-              [server methods](#servermethodname-method-options) or [plugins](#plugins). Each cache
-              name must be unique. A single item may omit the `name` option which defines the
-              default cache. If every cache includes a `name`, a default memory cache is provisioned
-              as well.
-            - `shared` - if `true`, allows multiple cache users to share the same segment (e.g.
-              multiple methods using the same cache storage container). Default to `false`.
-            - other options passed to the **catbox** strategy used.
+        - a configuration object, same as the server `cache` configuration options.
         - an array of the above object for configuring multiple cache instances, each with a unique
           name. When an array of objects is provided, multiple cache connections are established
           and each array item (except one) must include a `name`.
@@ -227,7 +221,7 @@ accessible. Initialized with an empty object.
 
 ```js
 const Hapi = require('hapi');
-server = new Hapi.Server();
+const server = new Hapi.Server();
 server.app.key = 'value';
 
 const handler = function (request, reply) {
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -1382,6 +1382,7 @@ for performing injections, with some additional options and response properties:
         - `close` - if `true`, emits a `'close'` event after payload transmission (if any).
           Defaults to `false`.
         - `end` - if `false`, does not end the stream. Defaults to `true`.
+        - `split` - indicates whether the request payload will be split into chunks. Defaults to undefined, meaning payload will not be chunked.
```
AdriVanHoudt: \`undefined\`

API.md
```diff
@@ -3015,7 +3017,7 @@ Each request object includes the following properties:
   different response. Contains `null` when no response has been set (e.g. when a request terminates
   prematurely when the client disconnects).
 - `preResponses` - same as `pre` but represented as the response object created by the pre method.
-- `query` - an object containing the query parameters.
+- `query` - an object containing the parsed [query parameters](https://nodejs.org/docs/latest/api/url.html#url_urlobject_query).
```
AdriVanHoudt: by default hapi uses `querystring` and `url` but you can set these yourself (for example when you parse using `qs`) so maybe link to it in a different way?

API.md
```diff
@@ -460,7 +454,7 @@ When the server contains exactly one connection, `registrations` is an object wh
 registered plugin name and value contains:
 - `version` - the plugin version.
 - `name` - the plugin name.
-- `options` - options used to register the plugin.
+- `options` - options object passed to the plugin during registration.
```
AdriVanHoudt: If you pass `{ once: true }` to the options does that get passed to the plugin? and does it end up in the options here?

AJamesPhillips: @AdriVanHoudt thanks for the speedy response.  Good question(s).  I will try to answer.  Will take me a few days to get back to you on your suggestions but I will let you know.

API.md
AJamesPhillips: Ah right, is this the ternary conditional on `typeof url === 'string'` here: https://github.com/hapijs/hapi/blob/770cc7bad15122f796d938738b7c05b66d2f4b7f/lib/request.js#L200-L201 ?

AdriVanHoudt: Not really, you can set the query object yourself (see https://github.com/hapijs/hapi/blob/master/API.md#requestseturlurl-striptrailingslash) which means it is not UrlObject.query per se. By default it will be though.

AJamesPhillips: It's pretty wordy now but hopefully this captures the behaviour accurately?
 https://github.com/hapijs/hapi/pull/3479/commits/fe6732224021ba755eef05138615665f4f33289f

Happy to change this again if you think there's improvement to be had.

API.md
AJamesPhillips: @AdriVanHoudt it does the options are part of the object with the `register` key ["an object with the following: ..."](https://hapijs.com/api#serverregisterplugins-options-callback).

    var server = new hapi.Server();
    server.connection({port: 3000})

    var pluginFunction = function(server, optionsPassedToPlugin, nextFn) {
        console.log('pluginFunction called optionsPassedToPlugin:', optionsPassedToPlugin);
        nextFn();
    }
    pluginFunction.attributes = {
        name: 'pluginFunction',
    }

    var options = {
        once: true, // ignored in terms of actual `once` option
        something: 'else',
    };
    var callback = function(err) {
        console.log('callback called')
    };

    server.register({
        register: pluginFunction,
        options: options,
    }, callback)

    console.log(server.registrations)

    // Output:
    //
    // pluginFunction called optionsPassedToPlugin: { once: true, something: 'else' }
    // { pluginFunction:
    //    { version: '0.0.0',
    //      name: 'pluginFunction',
    //      options: { once: true, something: 'else' },
    //      attributes:
    //       { name: 'pluginFunction',
    //         pkg: [Object],
    //         multiple: false,
    //         connections: true } } }
    // callback called

Did I understand your question correctly?

AdriVanHoudt: Hmm I thought passing `once` would work, calling plugin master @devinivy for his view :D

API.md
AdriVanHoudt: I would use `might` instead of `may` but that is nitpicking, the rest seems good! ^^

AJamesPhillips: (Apologies I really should have left rebasing and force push to the end of this PR)

API.md
```diff
@@ -113,16 +116,7 @@ Creates a new `Server` object where:
...
-            - `engine` - a prototype function or **catbox** engine object.
-            - `name` - an identifier used later when provisioning or configuring caching for
-              [server methods](#servermethodname-method-options) or [plugins](#plugins). Each cache
-              name must be unique. A single item may omit the `name` option which defines the
-              default cache. If every cache includes a `name`, a default memory cache is provisioned
-              as well.
-            - `shared` - if `true`, allows multiple cache users to share the same segment (e.g.
-              multiple methods using the same cache storage container). Default to `false`.
-            - other options passed to the **catbox** strategy used.
+        - a configuration object, same as the server `cache` configuration options.
```
hueniverse: Why not be explicit?

AJamesPhillips: I thought these were referring to the same things.  I did not understand this correctly and have reverted.

PR #3484, opened 2017-04-30
Add test for base64json cookie setting and parsing

AJamesPhillips:
I don't know if this kind of PR is appreciated?  I saw the example of setting a cookie using an object in the [helpful tutorial docs](https://hapijs.com/tutorials/cookies?lang=en_US) and thought it would be worth while to have it in the hapi tests?


dwmkerr:
I've got to admit this seems like a very sensible addition, it definitely gets my vote!

Issue #3429, opened 2017-01-23
Better support for async handlers


rubennorte:
Given that async functions are now standard it'd be great to have better support for them in Hapi. Specifically, using them as handlers:

```javascript
server.route({
  method: 'GET',
  path: '/',
  handler: async function (request, reply) {
    const data = await getData();
    const other = await getOther(data);
    reply(other);
  }
});
```

We can use them now as I described but any error in the handler would become an unhandled promise rejection. Catching them in the framework instead of forcing the programmers wrap the code inside a try/catch would be nice. This is what I'd like to avoid:

```javascript
server.route({
  method: 'GET',
  path: '/',
  handler: async function (request, reply) {
    try {
      const data = await getData();
      const other = await getOther(data);
      reply(other);
    } catch (error) {
      reply(error);
    }
  }
});
```

Would you accept a PR on this? Would you also use the result of the async function to call `reply` or only catch the errors?

Thanks for considering this.

nlf:
i thought async function support was still behind the `--harmony` flag? if that's the case, anything like this is unlikely to be accepted until the flag is no longer required.

cjihrig:
> i thought async function support was still behind the --harmony flag?

It is.

rubennorte:
Async functions will land in Node 8 without flags (they're available without them in V8 5.5, the version that's currently being used in Chrome stable).

I don't see a reason why this cannot be implemented now to be used with generators or transpiled code. Even with regular functions that return promises would be helpful.

hueniverse:
This would have to be done by me because I need to study this more.

kanongil:
As far as I know, the only thing this request requires from hapi, is to support returning a `Promise` from a handler by calling `reply()` when they fail. This is somewhat similar to how a `throw` is already handled.

rubennorte:
I reviewed the code and I think @kanongil is right. We just need to wrap the result of the handler as a promise and call reply with any errors it might throw.

@hueniverse good. I just want Hapi to support them no matter who implements it 😄

jeff-kilbride:
I'm trying to achieve the same basic thing, but I also wanted centralized error handling for all of my handlers without having to manually call some sort of utility error handling function. I'm new to `Hapi` and just wondering if I'm going about this the right way. I'm using what I have gleaned from looking at `hapi-plugin-co` to wrap all handlers using `onPreStart`. Also, I'm using `haute-couture` to wire my plugins. Here's what I have:

[...]

Issue #2771, opened 2015-09-17
Promises on prerequisites makes handler to swallow exceptions


dmacosta:
I've this issue when using `Promises` on a prerequisite handler and throwing an error from a handler.

```javascript
var Promise = require('promise');


exports.register = function (server, options, next) {

    var pre = function (request, reply) {

        return reply(Promise.resolve('value'));
    };

    server.route({
        method: 'GET',
        path: '/bar',
        config: {
            pre: [
                { method: pre, assign: 'pre' }
            ],
            handler: function (request, reply) {

                // throws here will get swallowed

                throw new Error('will timeout');
                return reply();
            }
        }
    });

    return next();
};

exports.register.attributes = {
    pkg: require('./package.json')
};
``` 

Replying with a literal value on the `pre()` does it make it work. Using Promises on the handler will too, but it sounds more like a workaround to me.

Here's the dependencies I'm using:

```
code@1.5.0 node_modules/code
└── hoek@2.16.2

promise@7.0.4 node_modules/promise
└── asap@2.0.3

hapi@9.3.1 node_modules/hapi
├── ammo@1.0.1
├── cryptiles@2.0.4
├── topo@1.0.3
├── peekaboo@1.0.0
├── items@1.1.0
├── kilt@1.1.1
├── catbox-memory@1.1.2
├── accept@1.1.0
├── call@2.0.2
├── statehood@2.1.1
├── boom@2.8.0
├── iron@2.1.3
├── shot@1.6.0
├── qs@4.0.0
├── hoek@2.14.0
├── catbox@6.0.0
├── mimos@2.0.2 (mime-db@1.18.0)
├── subtext@2.0.0 (content@1.0.2, wreck@6.1.0, pez@1.0.0)
├── joi@6.6.1 (isemail@1.1.1, moment@2.10.6)
└── heavy@3.0.0 (joi@5.1.0)

lab@5.17.0 node_modules/lab
├── diff@1.4.0
├── items@1.1.1
├── json-stringify-safe@5.0.1
├── bossy@1.0.3
├── eslint-config-hapi@2.0.1
├── hoek@2.16.2
├── espree@2.2.5
├── mkdirp@0.5.1 (minimist@0.0.8)
├── source-map-support@0.3.2 (source-map@0.1.32)
├── jslint@0.9.3 (exit@0.1.2, nopt@3.0.4, readable-stream@1.0.33, glob@4.5.3)
[...]

Issue #2097, opened 2014-11-05
Support Promises in reply() interface


yonjah:
As with every node app logic relays heavily on async calls.
Today you can only use pure callbacks with Hapi which can make you end up with messy code.
Promises would be a much nicer way for handling async calls, so instead of doing this -

``` javascript
server.pack.register({
        plugin: plugin,
    }, function callbackA(){
        doSomeMoreAsyncStuff(function callbackB() {
            //callbackB code
        });
});
```

We could do some thing like this -

``` javascript
server.pack.register({
    plugin: plugin,
}).then(function callbackA(){
    return doSomeMoreAsyncStuff();
}).then(function callbackB() {
    //callbackB code
});
```

In most cases code can be changed so it will still support both regular Callbacks and Promises style.

SimonDegraeve:
@hueniverse 
> Nope. Sorry. You can wrap hapi yourself but I am not going to change the framework from callback to promises. And while I am not dismissing those who find promises to be better, the above example does not demonstrate any benefit.

I have on use case. I have an array of plugins like this
```
var plugins = [
    Auth,       
    {
        register: Database,
        options: {}
    },    
    Routes,      
    ...
]
```

So I would like to register all of them, in my main plugin like this
```
server.register(plugins, function (err) {
    if (err) {
        console.error('Failed to load a plugin:', err);
    }
});
```

But I want to set a `route prefix` for the `Routes` plugin and because this option is only available in the register function and not the plugin options, I end up doing something like this
```
var plugins = [
    [Auth],        
    [{
        register: Database,
        options: {}
    }],   
    [Routes, {
      routes: { prefix: '/v11 }
    }],     
    ...
]

plugins.forEach(function(plugin) {
    server.register(plugin[0], plugin[1], function (err) {
        if (err) {
            console.error('Failed to load a plugin:', err);
        }
    });
});
```
But then I have to use an external dependency like [async](https://github.com/caolan/async) to make it works serially (some plugins depends on each other).
If the register function is `Promise-able` it would e easier to control the flow, without dependencies. 

Actually I use ES6 with the [6to5](https://github.com/6to5/6to5) transpiler. It looks like this
```
export async function register(server, options, next) {
    ...

    // async / await is ES7 wrapper for generator/promise

    try {

    // UGLY: Transform callback to promise
    var registerFn = server.register;
    server.register = (...args) =>
      new Promise((resolve, reject) =>
[...]

PR #3486, opened 2017-05-18
Provide cover for unhandled rejections

dwmkerr:
This small change allows an extra degree of protection for those who might be using promises in their handlers.

## Problem Statement

Developers who are using promises in their handlers will often write handlers which look like this:

```js
function handler(request, reply) {
  doSomething
    .then((interimResult) => {
      return somethingElse(interimResult);
    })
    .then((result) => {
      reply(result.whatever);
    });
```

The challenge is that this opens up lots of opportunities for exceptions at different points to lead to uncaught promise rejections. For example:

```js
function handler(request, reply) {
  doSomething
    .then((interimResult) => {
      throw new Error('whoops...');
    })
    .then((result) => {
      reply(result.whatever);
    });
```

will lead to a timeout of the request.

## Suggested Change

Rather than attempting some kind of large scale change which adds more support for promises (which is not needed, as `reply` can take a promise), I propose allowing the handler to return a promise. If the promise is returned, then hapi can register a `catch` and `reply` with any error. This means that if the developer goofs up and forgets to catch (all to easy with the promise spec), hapi at least will attempt to deal with the error in the same way it would as if an exception was thrown.

This PR adds the support, updates the docs and tests.

## Questions

**Do we need this?**

It is possible to rewrite the dodgy code as:

```js
function handler(request, reply) {
  reply(doSomething
    .then((interimResult) => {
      return somethingElse(interimResult);
    })
    .then((result) => {
      return result.whatever;
    }));
```

However this tends to feel a little unnatural, as the natural reading order is more 'do work, reply with the result'. Fairly subjective but I've seen quite a few devs on teams default to doing it the way described in the problem statement, rather than this. It also makes doing things like changing the status code conditionally based on the result near impossible. (CMIIW!)

Of course, people could just avoid promises (which might not be a bad idea given some of the issues around how they can swallow exceptions), but at least an appreciable proportion of people are using them heavily.

**Should we ever have 'partial promise' support**

Really, if you are returning a promise in the handler it might make more sense to simply not even have the `reply` function used, and change the spec to say `if you return a promise, hapi with reply with the result or if there is an error, wrap it in Boom and then reply with that`.

This would perhaps be cleaner in some cases:

```js
function handler(req, reply) {
  return doSomething().then(r = doSomethingElse(r));
}
```

But is a larger structural change. Also, it still means we cannot easily change status codes, add headers etc.

I *think* this is a happier middle group - hapi's API stays the same, but it has a 'safety net' for promises.

## That's It!

I'd love to get any feedback on this as a potential feature. It would certainly help myself and my colleagues on some of the projects we're on at the moment, I believe it allows us to provide a safety net for potentially nasty issues (promise black holes are a pain) but am happy to take any input for changes!

## Related Issues

- https://github.com/hapijs/hapi/issues/2097
- https://github.com/hapijs/hapi/issues/2771
- https://github.com/hapijs/hapi/issues/3242
- https://github.com/hapijs/hapi/issues/3429

### DIFF
--- a/API.md
+++ b/API.md
@@ -2871,23 +2871,25 @@ const handler = function (request, reply) {
 };
 ```
 
-If the handler returns a Promise then Hapi will register a `catch` handler on the promise object to catch unhandled promise rejections. The handler will `reply` with the error:
+If the handler returns a Promise then Hapi will register a `catch` handler on the promise object to catch unhandled promise rejections. The handler will `reply` with the rejected value, wrapped in a [`Boom`](https://github.com/hapijs/boom) error:
 
 ```js
 const handler = function (request, reply) {
 
-  const badPromise = () => {
+    const badPromise = () => {
 
-    new Promise(() => {
+        new Promise((resolve, reject) => {
 
-      setTimeout(() => throw new Error(), 1000);
+            // Hapi catches this...
+            throw new Error();
+
+            // ...and this...
+            return reject(new Error());
+        }
     }
-  }
 
-  // You *should* catch the rejection yourself! But hapi will
-  // return the error for you...
-  return badPromise()
-    .then((result) => { reply(result); });
+    // ...if you don't provide a 'catch'. The rejection will be wrapped in a Boom error.
+    return badPromise().then(reply);
 }
 ```
 
--- a/API.md
+++ b/API.md
@@ -2871,6 +2871,25 @@ const handler = function (request, reply) {
 };
 ```
 
+If the handler returns a Promise then Hapi will register a `catch` handler on the promise object to catch unhandled promise rejections. The handler will `reply` with the error:
+
+```js
+const handler = function (request, reply) {
+
+  const badPromise = () => {
+
+    new Promise(() => {
+
+      setTimeout(() => throw new Error(), 1000);
+    }
+  }
+
+  // You *should* catch the rejection yourself!
+  return badPromise()
+    .then((result) => { reply(result); })''
+}
+```
+
 ### Route prerequisites
 
 It is often necessary to perform prerequisite actions before the handler is called (e.g. load
--- a/API.md
+++ b/API.md
@@ -2884,11 +2884,14 @@ const handler = function (request, reply) {
     }
   }
 
-  // You *should* catch the rejection yourself!
+  // You *should* catch the rejection yourself! But hapi will
+  // return the error for you...
   return badPromise()
-    .then((result) => { reply(result); })''
+    .then((result) => { reply(result); });
 }
 ```
+
+This provides a safety net for unhandled promise rejections.
 
 ### Route prerequisites
 
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -5,6 +5,7 @@
 const Hoek = require('hoek');
 const Items = require('items');
 const Methods = require('./methods');
+const Promises = require('./promises');
 const Response = require('./response');
 
 
@@ -96,7 +97,12 @@ internals.handler = function (request, callback) {
 
     // Execute handler
 
-    request.route.settings.handler.call(bind, request, reply);
+    const handlerResult = request.route.settings.handler.call(bind, request, reply);
+
+    // Give hapi a fighting chance to deal with uncaught rejections.
+    if (Promises.isThennable(handlerResult)) {
+        handlerResult.catch(reply);
+    }
 };
 
 
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -3,6 +3,7 @@
 // Load modules
 
 const Hoek = require('hoek');
+const Boom = require('boom');
 const Items = require('items');
 const Methods = require('./methods');
 const Promises = require('./promises');
@@ -101,7 +102,18 @@ internals.handler = function (request, callback) {
 
     // Give hapi a fighting chance to deal with uncaught rejections.
     if (Promises.isThennable(handlerResult)) {
-        handlerResult.catch(reply);
+
+        handlerResult.then(null, (error) => {
+
+            //  Unhandled rejections are always Boom-ified, just like uncaught exceptions.
+            if (error instanceof Error) {
+                return reply(Boom.wrap(error));
+            }
+
+            const err = new Error('Rejected promise');
+            err.data = error;
+            return reply(Boom.wrap(err));
+        });
     }
 };
 
--- a/test/handler.js
+++ b/test/handler.js
@@ -170,7 +170,39 @@ describe('handler', () => {
             });
         });
 
-        it('catches unhandled promise rejections when a promise is returned', (done) => {
+        it('catches asynchronous exceptions in promise constructors via the standard \'protect\' mechanism', (done) => {
+
+            const server = new Hapi.Server({ debug: false });
+            server.connection();
+
+            const asyncOperation = function () {
+
+                return new Promise((resolve, reject) => {
+
+                    setTimeout(() => {
+
+                        throw new Error('This should be caught...');
+                    }, 100);
+                });
+            };
+
+            const handler = function (request, reply) {
+
+                return asyncOperation()
+                  .then(reply);
+            };
+
+            server.route({ method: 'GET', path: '/', handler });
+
+            server.inject('/', (res) => {
+
+                expect(res.statusCode).to.equal(500);
+                expect(res.result.error).to.equal('Internal Server Error');
+                done();
+            });
+        });
+
+        it('catches synchronous exceptions which lead to unhandled promise rejections when then promise is returned', (done) => {
 
             const server = new Hapi.Server();
             server.connection();
@@ -187,6 +219,46 @@ describe('handler', () => {
 
                 return asyncOperation()
                   .then(reply);
+            };
+
+            server.route({ method: 'GET', path: '/', handler });
+
+            server.inject('/', (res) => {
+
+                expect(res.statusCode).to.equal(500);
+                expect(res.result.error).to.equal('Internal Server Error');
+                done();
+            });
+        });
+
+        it('catches unhandled promise rejections when a promise is returned', (done) => {
+
+            const server = new Hapi.Server();
+            server.connection();
+
+            const handler = function (request, reply) {
+
+                return Promise.reject(new Error('This should be caught.'));
+            };
+
+            server.route({ method: 'GET', path: '/', handler });
+
+            server.inject('/', (res) => {
+
+                expect(res.statusCode).to.equal(500);
+                expect(res.result.error).to.equal('Internal Server Error');
+                done();
+            });
+        });
+
+        it('wraps unhandled promise rejections in an error if needed', (done) => {
+
+            const server = new Hapi.Server();
+            server.connection();
+
+            const handler = function (request, reply) {
+
+                return Promise.reject('this should be wrapped in an error');
             };
 
             server.route({ method: 'GET', path: '/', handler });
--- a/test/handler.js
+++ b/test/handler.js
@@ -169,6 +169,35 @@ describe('handler', () => {
                 done();
             });
         });
+
+        it('catches unhandled promise rejections when a promise is returned', (done) => {
+
+            const server = new Hapi.Server();
+            server.connection();
+
+            const asyncOperation = function () {
+
+                return new Promise((resolve, reject) => {
+
+                    throw new Error('This should be rejected...');
+                });
+            };
+
+            const handler = function (request, reply) {
+
+                return asyncOperation()
+                  .then(reply);
+            };
+
+            server.route({ method: 'GET', path: '/', handler });
+
+            server.inject('/', (res) => {
+
+                expect(res.statusCode).to.equal(500);
+                expect(res.result.error).to.equal('Internal Server Error');
+                done();
+            });
+        });
     });
 
     describe('register()', () => {
--- a/lib/promises.js
+++ b/lib/promises.js
@@ -24,3 +24,9 @@ exports.wrap = function (bind, method, args) {
         method.apply(bind, args ? args.concat(callback) : [callback]);
     });
 };
+
+exports.isThennable = function (candidate) {
+
+    return candidate && (typeof candidate.then === 'function') ? true : false;
+
+};
--- a/test/promises.js
+++ b/test/promises.js
@@ -1,0 +1,43 @@
+'use strict';
+
+// Load modules
+
+const Code = require('code');
+const Lab = require('lab');
+const Promises = require('../lib/promises');
+
+// Declare internals
+
+const internals = {};
+
+
+// Test shortcuts
+
+const lab = exports.lab = Lab.script();
+const describe = lab.describe;
+const it = lab.it;
+const expect = Code.expect;
+
+
+describe('promise', () => {
+
+    it('recognises a promise as thennable', (done) => {
+
+        const promise = new Promise(() => {});
+        expect(Promises.isThennable(promise)).to.equal(true);
+
+        done();
+
+    });
+
+    it('recognises invalid input as not thennable', (done) => {
+
+        expect(Promises.isThennable(undefined)).to.equal(false);
+        expect(Promises.isThennable(null)).to.equal(false);
+        expect(Promises.isThennable({ })).to.equal(false);
+        expect(Promises.isThennable({ then: true })).to.equal(false);
+
+        done();
+
+    });
+});


devinivy:
Hey, just wanted to share– there are two things that can help you think about this.

1. You can reply with a pre-generated response– this way, you can turn promise results into responses with status codes, cookies, headers, etc.  See [`request.generateResponse()`](https://github.com/hapijs/hapi/blob/master/API.md#requestgenerateresponsesource-options).
2. I have opted to handle general-purpose promise rejections with a helper that you can use, [`Toys.handler()`](https://github.com/devinivy/toys/blob/master/API.md#toyshandlerasynchandler).  That's designed for use with async/await, but really works with any promise-returning function.

My impression is that there is not any intention of porting this feature into hapi land, but we shall see :)

dwmkerr:
Thanks @devinivy, I like the tips! I'll look into both as an interim solution. I'd like to stay as close to hapi semantics as I can and avoid wrapping it with other code, but definitely have to try and find a way to reduce the likelihood of those unhandled rejections occurring.

I'm a fan of promises in how they can make some code a little easier to reason about, but the extra effort required to make sure people are getting all of the error conditions right is a nightmare!

AdriVanHoudt:
Can you not add a `process.on('unhandledRejection', (err, promise) => {})` to catch that behaviour?

dwmkerr:
Hi @AdriVanHoudt,

The challenge with this is that there is no way to then actually call `reply` and successfully continue with the request pipeline. What I am aiming for is a way for an unhandled rejection to go through the same workflow as an unhandled exception. (i.e. with this approach I can still use `onPreReply` etc etc to transform the error if needed)

AdriVanHoudt:
Ok but normally you should catch this in testing no?
Btw I think the proposed PR is not bad, less ways to shoot yourself in the foot is 👌

lib/handler.js
```diff
@@ -96,7 +97,12 @@ internals.handler = function (request, callback) {
 
     // Execute handler
 
-    request.route.settings.handler.call(bind, request, reply);
+    const handlerResult = request.route.settings.handler.call(bind, request, reply);
+
+    // Give hapi a fighting chance to deal with uncaught rejections.
+    if (Promises.isThennable(handlerResult)) {
+        handlerResult.catch(reply);
```
kanongil: It's probably better to use `.then(null, reply)` instead of `catch()`, since we haven't tested that the `catch()` method exists.

API.md
```diff
@@ -2871,6 +2871,28 @@ const handler = function (request, reply) {
...
+      setTimeout(() => throw new Error(), 1000);
+    }
+  }
+
+  // You *should* catch the rejection yourself! But hapi will
+  // return the error for you...
+  return badPromise()
+    .then((result) => { reply(result); });
+}
+```
```
kanongil: Bad indentation.

API.md
```diff
@@ -2871,6 +2871,28 @@ const handler = function (request, reply) {
...
+If the handler returns a Promise then Hapi will register a `catch` handler on the promise object to catch unhandled promise rejections. The handler will `reply` with the error:
+
+```js
+const handler = function (request, reply) {
+
+  const badPromise = () => {
+
+    new Promise(() => {
+
+      setTimeout(() => throw new Error(), 1000);
```
kanongil: The `setTimeout` seems redundant, and confuses the example.

API.md
```diff
@@ -2871,6 +2871,28 @@ const handler = function (request, reply) {
...
+    }
+  }
+
+  // You *should* catch the rejection yourself! But hapi will
+  // return the error for you...
```
kanongil: Why? You should only need to catch it, if you want to handle it in a special way.

lib/handler.js
kanongil: It also needs to handle rejections where the object is not an `instanceof Error`, like this code:
https://github.com/hapijs/hapi/blob/b4aeaf77cf3f74fa12b2a514c01748768d974e4a/lib/response.js#L471-L480.

dwmkerr: Hi @kanongil yes it needs to use `.then`, goofy mistake on my side. For the second point that's a great spot, am checking on this now.

dwmkerr:
@AdriVanHoudt agreed, tests should cover these scenarios, but I think it still adds a layer of safety as even for experienced developers it can be easy to miss an edge case with a promise

API.md
AJamesPhillips: I guess it's to show that errors thrown from asynchronous functions inside Promises are caught and handled correctly.  By extension, synchronous errors will be caught and handled correctly too.  (?)  Happy to be corrected.

test/handler.js
```diff
@@ -169,6 +169,35 @@ describe('handler', () => {
...
+        it('catches unhandled promise rejections when a promise is returned', (done) => {
+
+            const server = new Hapi.Server();
+            server.connection();
+
+            const asyncOperation = function () {
+
+                return new Promise((resolve, reject) => {
+
+                    throw new Error('This should be rejected...');
```
AJamesPhillips: I'd add a test for your  `setTimeout(() => throw new Error(), 1000);` too from above.

AdriVanHoudt:
@dwmkerr 👌 as @kanongil pointed out this actually matches the behaviour with callbacks so that is even an extra for me

API.md
dwmkerr: @kanongil @AJamesPhillips yes that's correct, it's to make it clear and explicit that this method will work even if exception is thrown in a genuinely async context (a `nextTick` could be used if the fact that it is a timeout might be misleading people into thinking it is a timing related issue)

API.md
dwmkerr: I'll make the comment clear, what I meant was that if you are orchestrating a promise chain, you should pretty much always use a catch (unless you return the promise to something else which can catch it). In this case as hapi would be able to catch it, the comment is misleading

API.md
dwmkerr: On it

test/handler.js
dwmkerr: Good call! Adding now

API.md
TimBeyer: Maybe there's something I'm missing here, but unless there's some magic going on behind the scenes with [domains](https://nodejs.org/api/domain.html) this should result in resolving with `undefined` and triggering an `uncaughtException` on the process.

dwmkerr: Let me make sure the test case is exactly right for this to confirm everything, working on the changes now

dwmkerr: @TimBeyer thanks for the heads up, you are quite right. In this case the following is hit:

https://github.com/hapijs/hapi/blob/master/lib/protect.js#L55

And so this case is handled. This feature will only solve for unhandled rejections and exceptions thrown *synchronously* in a promise constructor, I'm updating the tests to correctly reflect this.

lib/handler.js
dwmkerr: fixed, also added support for wrapping the object in an error needed to maintain parity with the unhandled exceptions. would welcome advice on whether you think the logic to wrap an error in a boom object is general enough to extract into its own function to avoid duplication of logic @kanongil. It's a few line and might be clearer left inline, but am happy to extract it (perhaps into `promise.js` as a `boomReject` function or something)

API.md
dwmkerr: Added a test case showing that the `setTimeout` example is already handled by the `protect` mechanism, to keep things clearer. Also attempted to make the api documentation sharper.

test/handler.js
dwmkerr: Added the test, it's there now :)

dwmkerr:
Hi @AdriVanHoudt yes it now goes through the same process as with callbacks, transforming the error by wrapping it in Boom as needed, so the behaviour of both is very close now 😄

lib/handler.js
```diff
@@ -96,7 +98,23 @@ internals.handler = function (request, callback) {
...
-    request.route.settings.handler.call(bind, request, reply);
+    const handlerResult = request.route.settings.handler.call(bind, request, reply);
+
+    // Give hapi a fighting chance to deal with uncaught rejections.
+    if (Promises.isThennable(handlerResult)) {
+
+        handlerResult.then(null, (error) => {
+
+            //  Unhandled rejections are always Boom-ified, just like uncaught exceptions.
+            if (error instanceof Error) {
```
hueniverse: Is it common practice to reject a promise with a non-Error object?!

hueniverse:
Should this be the default behavior? Do people who use promises consider this a breaking change?

lib/handler.js
nlf: i sure hope not..

devinivy: You certainly aren't supposed to, but out in the wild it happens.  As to whether hapi should deal with the rejection of a non-error, I can't say.  I would probably treat it the same way hapi deals with a non-error thrown in the handler.

corbinu:
Just one user here but I do use promises with hapi. I would like this as the default behavior but would consider it a breaking change :/

nlf:
i'm curious why you would consider this a breaking change @corbinu. are you using a handler somewhere in the request lifecycle to intercept promise rejections somehow?

that's the only situation i can think of off the top of my head that would cause this patch to lead to undesirable results. i'm definitely all ears if i'm missing something, though.

corbinu:
I am not but assumed somebody might be you guys know your users better than I do though :)

nlf:
just checking, thought maybe you had a specific use case in mind. i personally can't think of one, and unless someone else can IMO i'm not sure i'd call this a breaking change.

to me this is very similar to having handlers run through domains, the intent isn't to alter behavior, but to make sure that a response is sent even if an unexpected failure occurs. some sort of error handling plugin is the only thing i can think of this change would interfere with, and an error handling plugin that only catches rejected promises seems like the sort of thing a user would be happy to get rid of in favor of native support.

TL;DR - unless someone has a strong objection or legitimate use case for the existing behavior, i don't believe this is a breaking change and i do believe it's a good idea to enable by default

devinivy:
I agree that this isn't a breaking change.  If you're already catching your own rejections, that should still work just fine.

lib/handler.js
atrauzzi: Yeah, can happen.  Have seen strings thrown.

dwmkerr: @hueniverse it is definitely non-conventional to `reject` a non-error object. If done deliberately it would definitely be an anti-pattern. However, there are cases where it will slip in by mistake, here's the most common scenario:

```
function doSomethingAsynchronous(input) {
  return new Promise((resolve, reject) => {
    callbackToLibrary(input, (err, result) => {
      if (err) return reject(err);  // uh-oh....
      return resolve(result);
    });
  });
}
```
In this case, if the library you are using does something goofy like callback with a non-error object, the promise will reject with the non-error object.

I think that with the handling of thrown exceptions in domains we have the same behaviour (because `Boom.wrap` covers whatever is thrown, ensuring that even if what is thrown is not an error, it gets wrapped in one).

dwmkerr:
@nlf that's essentially the idea, that we get a similar opportunity to cover unhandled rejections in the same way we can with uncaught exceptions in domains. In [Node 8 domains will be aware of native unhandled rejections](https://nodejs.org/en/blog/release/v8.0.0/) interestingly (which should make it easier to be resilient in this situation, even if the user doesn't `return` the promise there's a chance to catch it).

@devinivy yep, if you do:

```function handler(request, reply) {
  return doSomething().
    then((result) => reply(result))
    .catch((err) => console.log("Don't touch my error..."));
}
```

Then the extra `catch` I added will not be called, so if you already catch your own rejections, there is no change to behaviour.

rubennorte:
> In Node 8 domains will be aware of native unhandled rejections interestingly

Not exactly. It just executes the functions passed to `then` and `catch` with the proper domain, which wasn't happening before and forced users to use another Promise implementation with support for it.

jeff-kilbride:
I was using generators with a different framework and am now using promises and async / await with Hapi. I would love to see better built-in support for unhandled rejections. I am currently using `onPreStart` in a plugin to walk through the handlers and manually attach a `catch` to any that return a promise:
```
module.exports = [{
    type: 'onPreStart',
    method: function (server, next) {

        // Iterate over all connections.
        const connections = server.table();
        for (const conn of connections) {

            // Iterate over all routes in each connection.
            for (const route of conn.table) {
                if (typeof route.settings.handler !== 'function')
                    continue;

                // Store the original handler function.
                const h = route.settings.handler;
                route.settings.handler = (request, reply) => {

                    // Call the original handler.
                    const p = h(request, reply);

                    // If the handler returns a promise, attach a `catch()` method.
                    p && p.catch && p.catch(reply);
                };
            }
        }

        next();
    }
}];
```
This isn't the best implementation, but it works. However, with this PR, it looks like this won't be necessary anymore. Correct? It looks like you're doing something similar, but at the framework level -- which is great.  👍 

Just wondering if I'm missing anything or if I'll be able to get rid of the above code once this lands.

dwmkerr:
@rubennorte thanks for the clarification!

dwmkerr:
@jeff-kilbride Correct, you will no longer need the code shown, however there is one subtle caveat. With your code, if a handler does this:

```js
function handler(request, reply) {
  return new Promise((reply, reject) => {
    throw "The cake is a lie";
  });
}
```

Your current work-around will lead to `reply("The cake is a lie")`. With the implementation I have added, Hapi will check whether the rejected value is an error, if it is not it will wrap it in a Boom object, meaning you essentially have: `reply(Boom.wrap("The cake is a lie")`.

In short:

*Before*: `reject('the cake is a lie')` -> HTTP 200, with content 'the cake is a lie'
*After*: `reject('the cake is a lie`) -> HTTP 500, with hapi error content

This is to maintain parity with how hapi handles uncaught exceptions (wrapping them in errors).

jeff-kilbride:
@dwmkerr  I upgraded to hapi 16.4.3 today with these latest changes and removed my `onPreStart` extension (from previous comment), but it's not working as I expected. Here's a simple example of the way I write handlers:

```js
Class FooHandler {

    static async myHandler(request, reply) {
        const res = await someAsyncMethod();
        reply(res.data);
    }
}

//  In separate route config file...
const Handler = require('FooHandler');
return [
    {
        method: 'GET',
        path: '/',
        handler: Handler.myHandler
    }
];
```
Typically, `someAsyncMethod` is an http request to my API backend using `axios` and returns the `axios` promise. 

With my extension code, when an error/rejection occurs in `someAsyncMethod` (e.g. a 4xx or 5xx http error), it is caught and fed into the `reply` method. I then handle the error in an `onPreResponse` extension, reformatting it if necessary, before calling `reply.continue()`. With the upgrade to 16.4.3, when an error/rejection occurs in `someAsyncMethod`, the handler just dies. My `onPreResponse` extension is never called and there is no reply received by the front end. Here is an example of the error in the logs:
```
(node:180) Error: Request failed with status code 401
     at createError (/app/node_modules/axios/lib/core/createError.js:15:15)
     at settle (/app/node_modules/axios/lib/core/settle.js:18:12)
     at IncomingMessage.handleStreamEnd (/app/node_modules/axios/lib/adapters/http.js:191:11)
     at emitNone (events.js:91:20)
     at IncomingMessage.emit (events.js:185:7)
     at endReadableNT (_stream_readable.js:974:12)
     at _combinedTickCallback (internal/process/next_tick.js:80:11)
     at process._tickDomainCallback (internal/process/next_tick.js:128:9)
 Debug: request, closed, error
```
I am not explicitly returning a promise from the handler and am wondering if that may be the problem. In my extension, I wrap the original handler and capture the implicit promise created by `async`. I don't think you are handling `async` methods the same way.

Fortunately, my `onPreStart` extension still works with the upgrade, but I would eventually like to remove it. I would much prefer if the framework handled this stuff.

cur3n4:
@jeff-kilbride I am doing something similar and it is working fine. The only difference is that my handler method is not static.
I had a similar issue where the hapi library being used was the one under glue and not the main one.

jeff-kilbride:
@cur3n4  Thanks for the heads up! It looks like `glue` could be the issue. I added the `server.version` variable to my startup string and got:
```
Hapi server version 16.3.0 started at http://7b72193a91a0:4000
```
I have `hapi` as a dependency in my package.json file:
```
    "hapi": "^16.4.3",
```
How do I get `glue` to use the correct version? Were you able to fix this?

cur3n4:
In my package.json I have glue ^4.1.0. That did it for me.

jeff-kilbride:
I found it. I needed to rebuild my docker containers after upgrading hapi.  🙄 

Looks like everything is working as it should, now, without my extension!  🎉 

@cur3n4  Thanks for the push in the right direction! I'm deleting my previous post, so it doesn't cause any confusion...

dwmkerr:
@cur3n4 thanks for the pointers, I've been busy so haven't been able to get to this until now. @jeff-kilbride delighted it works for you and you can chop out some extra code! Doing the same in my project now 😄

krazik:
So I have Hapi 16.5.2 when is this going to be included?

jeff-kilbride:
It's been included since 16.4.0.

PR #3506, opened 2017-05-31
Add json escape option

g-k:
refs: https://github.com/hapijs/hoek/pull/204

Adds the `json.escape` option to call `hoek.escapeJSON` on the JSON output. It's disabled by default.

There's probably a better place and name for this option.


API.md
```diff
@@ -2521,10 +2521,11 @@ following options:
   for internal routes that should not be accessible to the outside world. Defaults to `false`.
 
 - `json` - optional arguments passed to `JSON.stringify()` when converting an object or
-  error response to a string payload. Supports the following:
+  error response to a string payload or escaping it after stringification. Supports the following:
     - `replacer` - the replacer function or array. Defaults to no action.
     - `space` - number of spaces to indent nested object keys. Defaults to no indentation.
     - `suffix` - string suffix added after conversion to JSON string. Defaults to no suffix.
+    - `escape` - calls [`Hoek.jsonEscape`](https://github.com/hapijs/hoek/blob/master/API.md#escapejsonstring) after conversion to JSON string. Defaults to not escaping.
```
AdriVanHoudt: I would go with "defaults to `false`."

package.json
```diff
@@ -26,7 +26,7 @@
     "catbox-memory": "2.x.x",
     "cryptiles": "3.x.x",
     "heavy": "4.x.x",
-    "hoek": "4.x.x",
+    "hoek": "git+https://git@github.com/hapijs/hoek.git",
```
AdriVanHoudt: This should be a legit version, maybe ask @nlf to do a release of hoek?

test/response.js
```diff
@@ -800,6 +800,26 @@ describe('Response', () => {
         });
     });
 
+    describe('escape()', () => {
```
AdriVanHoudt: Add a test with a success case as well for `.escape()`

AdriVanHoudt:
Also CI is failing since, I think, hapi uses a shrinkwrap that causes it to install hoek@4 and not the latest version you told it to in the package.json. A new version of hoek would solve this.

hueniverse:
@g-k you still have some open comments above (beside the missing hoek version).

PR #3507, opened 2017-06-02
Node 8 fixes

### LINKED ISSUE #3510: Update hapijs/shot to 3.4.2 from 3.4.0

kanongil:
Node 8 changes the `stream.Readable` implementation a bit. Some of the tests expects the `.destroy()` method to be absent, which I fixed by always clearing it.

The more serious issue, is `server.inject()`, which can return incorrect headers. This is not a test issue, but rather a bug, which is fixed by updating to the latest `shot` release.

Closes #3510

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -63,7 +63,7 @@
       "version": "1.2.5"
     },
     "shot": {
-      "version": "3.4.0"
+      "version": "3.4.2"
     },
     "statehood": {
       "version": "5.0.2"
--- a/test/transmit.js
+++ b/test/transmit.js
@@ -1785,6 +1785,8 @@ describe('transmission', () => {
 
                 const stream = new Stream.Readable();
 
+                stream.destroy = undefined;    // Node 8 streams comes with a destroy method – disable for this test
+
                 stream._read = function (size) {
 
                     const chunk = new Array(size).join('x');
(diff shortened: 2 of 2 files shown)


cjihrig:
Node core has a testing utility named [Canary in the Gold Mine (CITGM)](https://github.com/nodejs/citgm) that is used to detect npm ecosystem breakage. It runs the regression tests of many popular modules, but [hapi is no longer one of them](https://github.com/nodejs/citgm/issues/8). @hueniverse, you might want to work with the CITGM team to get hapi tested again.

Sorry to hijack this PR.

DavidTPate:
@kanongil For my own curiosity can you point me in the direction of the tests that are expecting the `destroy` method to not be present? I'm guessing that you're referring to [response](https://github.com/hapijs/hapi/blob/c4597b1ddb86ae1f9fd99ddc14aabcf37157df95/lib/response.js#L627-L629).

If we have a part causing these tests to fail with Node 8 instead of just removing the `destroy()` method for streams I would expect us to fix the underlying code to prevent compatibility issues with Node 8.

kanongil:
@DavidTPate

AdriVanHoudt:
it looks like hapi has the same errors on node 6 https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-nobuild/150/#showFailuresLink

Issue #3512, opened 2017-06-05
Remove npm-shrinkwrap


hueniverse:
npm 5 completely broke the solution we were using until now. It forces changes on the existing file so no manual management is possible. The new format then breaks older versions of npm. I am not sure what's the full replacement but for now I'll just lock the package.json versions for the direct dependencies which is not perfect but better than nothing.

If anyone has suggestions, let me know.

devinivy:
Anyone know more about this?  You can't author/manage shrinkwraps manually anymore?

zkat:
So, a couple of things:

1. The new format _should_ be backwards-compatible. Is there something that's breaking? D:
2. we've been moving away from a "pick your own shrinkwrap" model for a while. One of those changes was adding devDeps. Another was no longer allowing partial shrinkwraps. I'm not sure what else would be breaking you here, aside from moving to the new format: the only difference besides the overall format change is that we `--save` by default, which has been updating shrinkwraps already since `npm@3`.

devinivy:
@zkat thank you very much for the speedy response!

I imagine part of our problem is that we use a partial shrinkwrap (if I remember correctly).  But I'm not sure that's necessary for us.  I still don't fully understand the issue @hueniverse is keying-in on here.  I'd guess that the "forced changes" to the shrinkwrap could be ignored using a `git checkout` after an `npm install`.

zkat:
to be clear: unless you're installing new modules or removing them, `npm-shrinkwrap.json` should _not_ change, unless something weird's going on. In which case we should fix it. /cc @iarna to make sure I got this right.

We really haven't supported partial shrinkwraps for a while, and last time I asked, I thought hapi had stopped doing that (it was one of the reasons we weren't making the change yet!)

iarna:
I concur with everything @zkat said here. I'm pretty sure we patched the partial shrinkwrap out of hapi before the npm@4 launch (I opened in issue here in advance of that, since we knew that change was going to break you all.)

We definitely aren't going to be supporting manually editing your shrinkwraps going forward, but if you want to mutate it in a consistent way the `postshrinkwrap` lifecycle exists to support that.

iarna:
@hueniverse If you'd like, I'd be happy to PR a `postshrinkwrap` that gives makes all npm's give you shrinkwraps comparable to your old one.

hueniverse:
[...]

PR #3513, opened 2017-06-06
Restore shrinkwrap

iarna:
Regarding: #3512

This restores the shrinkwrap incrementally so that the differences between the existing shrinkwrap and the new one are more obvious. It's also put together such that if you want to skip the integrity or resolved fields you can just skip those commits.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "code": "4.x.x",
     "handlebars": "4.x.x",
     "inert": "4.x.x",
-    "json": "9.x.x",
     "lab": "13.x.x",
     "vision": "4.x.x",
     "wreck": "12.x.x"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,8 +49,7 @@
   "scripts": {
     "test": "lab -a code -t 100 -L -m 3000",
     "test-tap": "lab -a code -r tap -o tests.tap -m 3000",
-    "test-cov-html": "lab -a code -r html -o coverage.html -m 3000",
-    "postshrinkwrap": "json -I -f npm-shrinkwrap.json -e 'f(this) ; function f (p) { var deps = p.dependencies || {}; Object.keys(deps).forEach(function (n) { delete deps[n].resolved; f(deps[n]) }) }'"
```
devinivy: This commit seems to have clobbered your nifty postshrinkwrap script!

iarna: I did! Because the final commit makes the `npm-shrinkwrap.json` be entirely unedited, and thus the `postshrinkwrap` script no longer has any purpose. The idea here is to illustrate some options on what one might do with the shrinkwrap over the course of the commits. One can crop this commit log anywhere after devdeps are added and it'll produce the kind of shrinkwrap in that commit.

devinivy: Ah, I see!  Very cool.  I was originally confused because the `json` package was added to devDeps but only used in the postshrinkwrap, which had disappeared.

iarna: oh oops, lemme remove that from the final commit…

iarna: There =)

devinivy: 👌 👌  :)

PR #3518, opened 2017-06-08
Add docs on sameSite having some browser support, need for CSRF token…

AJamesPhillips:
…, option behaviour

Do people think this is a) useful to add and b) correct (I sure hope it is but one always needs to triple check with security).

### DIFF
--- a/API.md
+++ b/API.md
@@ -1898,10 +1898,10 @@ across multiple requests. Registers a cookie definitions where:
       deleted when the browser is closed).
     - `isSecure` - sets the 'Secure' flag. Defaults to `true`.
     - `isHttpOnly` - sets the 'HttpOnly' flag. Defaults to `true`.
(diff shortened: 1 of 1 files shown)


AdriVanHoudt:
useful yes, correct, not sure

AJamesPhillips:
The only other thing is that the other options are not explained in detail so perhaps this one should not be either and people should just do their own research on this issue... though I'd like to think that hapi aims to be the most secure of all web applications which includes the education of its new (and existing?) users.

AdriVanHoudt:
I think it is better -> linking to the official docs should be required though ^^

hueniverse:
I actually rather the docs be less useful here. This is a complicated security option and I don't want people to learn about it from the hapi docs. A simple link to the spec is enough. I don't want to explain the options or give guidance.

PR #3520, opened 2017-06-09
Fixed validation of server's Joi options

etrnljg:
I encountered an error when trying to pass this to my server:
```js
// ...
{
  modify: true,
  options: {
    stripUnknown: {
      objects: true,
      arrays: false
    }
  }
}
```
...telling me `stripUnknown` didn't `meet requirement of having peer modify set to true`.

It happened the validation made in Hapi was asserting stripUnknown was matching the validation of `modify`: a boolean. However, the documentation specifies `stripUnknwon` can also take an object.

Just to make a remark about my commit; I'm kind of disappointed to provide the validations for the inner values of the object [here](https://github.com/hapijs/hapi/compare/master...Brigad:validationOptions/stripUnknown?expand=1#diff-f412f4722a52cf22ed3d0a9a694b837dR144) because I feel like this is not the job it is intended to. Especially since I'm not doing it [there](https://github.com/hapijs/hapi/compare/master...Brigad:validationOptions/stripUnknown?expand=1#diff-f412f4722a52cf22ed3d0a9a694b837dR139) as it is resolved later by Joi itself.
I would  prefer to stick with a simple:
```js
// ...
otherwise: Joi.boolean().only(false)
```
Any preference?

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -132,7 +132,22 @@ internals.routeBase = Joi.object({
         status: Joi.object().pattern(/\d\d\d/, Joi.alternatives(Joi.object(), Joi.array(), Joi.func()).allow(true, false))
     })
         .without('modify', 'sample')
-        .assert('options.stripUnknown', Joi.ref('modify'), 'meet requirement of having peer modify set to true'),
+        .assert('options.stripUnknown', Joi.alternatives().when('modify', {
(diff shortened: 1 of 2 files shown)


hueniverse:
This was too complicated. I kept it simple.

PR #3531, opened 2017-06-28
Return 413 when payload too large.

aheinz-fe:
From https://github.com/hapijs/discuss/issues/516


aheinz-fe:
See also https://github.com/hapijs/subtext/pull/48

PR #3538, opened 2017-07-13
Add support for Referrer-Policy security header

victorkohl:
This PR adds support for Referrer-Policy security header.
The allowed values were taken from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy).

The default value was set to `origin` because it's a middle ground between sending the full URL and not sending the referrer at all, but it's up to debate.


hueniverse:
This feels premature given how new this is and how little support there is for it outside of Firefox.

@nlf? @evilpacket?

Issue #3509, opened 2017-06-02
Return 413 when payload is too big


twrobel3:
In order to comply with [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.5.11) we would like our application to response with status code 413 - Payload Too Large - when the maximum payload size, as defined in the route by `config.payload.maxBytes` is exceeded.

However, Hapi has this implemented by responding with a 400 - Bad Request.  In addition, the payload parsing failAction is required to be one of `error`, `log`, or `ignore`.  Unlike validation failures, there is no opportunity to provide a custom failAction.

Currently, we are working around this limitation by checking the response code and content during the onPreRequest event, but that is rather hacky and not ideal.

* *node version*: 7.4.0
* *hapi version*: 16.1.0
* *os*: OSX
* *any other relevant information*:

hueniverse:
I don't mind this feature, but the real fix to your specific issue is to return the right status code. I'll take over this issue for that.

hueniverse:
subtext already returns 413, but wreck and pez need to change.

PR #3539, opened 2017-07-14
Add payload fail action custom handler

jeff-kilbride:
Hi All,

This PR adds a custom handler implementation to the route payload failAction, much like the route validate failAction. It addresses #3509.

I have also added tests for the three string failAction types, along with a test for the new functionality.

I'm consistently getting an error in one test locally that was present before I made these changes:

```
Failed tests:

  757) transmission transmit() handles stream errors on the response after the response has been piped (inject):

      Timed out (3000ms) - transmission transmit() handles stream errors on the response after the response has been piped (inject)
```

Just a head's up.

Thanks!

### DIFF
--- a/API.md
+++ b/API.md
@@ -2591,6 +2591,11 @@ following options:
         - `'error'` - return a Bad Request (400) error response. This is the default value.
         - `'log'` - report the error but continue processing the request.
         - `'ignore'` - take no action and continue processing the request.
+        - a custom error handler function with the signature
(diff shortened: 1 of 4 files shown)


jeff-kilbride:
I'm using node v6.11.0 locally and it looks like the error I'm getting is also showing up in `travis` in v4 and v6. It also looks like v4 handles the `payload error` logged to `request-internal` a bit differently.

PR #3560, opened 2017-08-03
fix response.message(httpMessage)

amincheloh:
res.statusMessage should assign before calling res.writeHead() explicitly. see https://nodejs.org/api/http.html#http_response_statusmessage


kanongil:
Looks legit – thanks.

Can you add a test as well?

amincheloh:
@kanongil added

PR #3576, opened 2017-09-16
chore(improvement): no need to loop arguments

incubus8:
This PR is minor improvement to current Hapi code in `plugin.js`. If the intent is to pass all `arguments` from select. We can just process flattened on `arguments` variable without having redeclaring and put it into temporary variables. CMIIW.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -101,12 +101,7 @@ internals.Plugin.prototype._single = function () {
 
 internals.Plugin.prototype.select = function (/* labels */) {
 
-    let labels = [];
-    for (let i = 0; i < arguments.length; ++i) {
-        labels.push(arguments[i]);
-    }
-
-    labels = Hoek.flatten(labels);
+    const labels = Hoek.flatten(arguments);
     return this._select(labels);
 };
 


lib/plugin.js
```diff
@@ -101,12 +101,7 @@ internals.Plugin.prototype._single = function () {
 
 internals.Plugin.prototype.select = function (/* labels */) {
 
-    let labels = [];
-    for (let i = 0; i < arguments.length; ++i) {
-        labels.push(arguments[i]);
-    }
-
-    labels = Hoek.flatten(labels);
+    const labels = Hoek.flatten(arguments);
```
cjihrig: In older versions of Node, writing it the way you're proposing came with a decent performance hit. I still don't like the idea of using `arguments` unless necessary. I'd go with a rest parameter instead.

incubus8: Hi @cjihrig, many thanks for your feedback. It's my first PR for Hapi. I'd like to know more why is removing for-loop is slower than having for-loop for `arguments`. Would you mind to elaborate it a bit more? In my opinion, using `arguments` variable directly is resulting the same result with previous implementation - `labels` variable.  I do agree with you to use rest parameter. Since Hapi requires Node version >= 4.8.0, we can't use rest parameter.

cjihrig: You can read about "optimization killers" [here](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers). A lot of that document is specific to older versions of Node. The upcoming v17 release will drop support for Node < 8, at which point using a rest parameter should be fine.

incubus8: @cjihrig thanks for the heads up 👍 I'll close this PR.

PR #3578, opened 2017-09-17
Use rest parameter for optimized code

incubus8:
As @cjihrig suggested, this is **breaking changes** toward Hapi 17.0.0 release by using Node 8.3. 

Hapi will be in 17.0.0, toward this release we can use Node version 8 above. In Node 8.3, we can use rest parameter to destruct arguments and avoid workaround solution to optimized arguments. 

Implementation reference: https://github.com/davidmarkclements/v8-perf
PR reference: https://github.com/hapijs/hapi/pull/3576


hueniverse:
Thanks but this is already taken care of in v17.

PR #3579, opened 2017-09-19
v17 rc 4

hueniverse:
Feel free to leave comments. This is very early and not meant to be merged.


corbinu:
Does this mean going forward that hapi will support only the lastest node? Or is this planned to land after 8 goes LTS and hapi will support only the latest LTS. I am curious only for devops planning.

AdriVanHoudt:
@corbinu afaik the plan is to support the latest LTS

hueniverse:
We support as many LTS versions as possible. This will drop support for node 4 and 6, but we will continue to support hapi v16 as an LTS for those versions. But in general, hapi is a framework that should really be kept up to date. It should be part of your budget when picking it.

corbinu:
@hueniverse IMHO it really needs to support at least one LTS though as the Node Foundation does not recommend latest for production deployments. Next one drops next month so shouldn't be an issue though.

hueniverse:
v17 will not be ready under 10 days. We are fully aware of the schedule...

corbinu:
@hueniverse I wasn't saying you wern't but your first response made it sound very much like @AdriVanHoudt statement was not the official policy.

AdriVanHoudt:
@corbinu my statements are never the official policy unless I directly quote @hueniverse ;)

mcollina:
👍  on the courage to go fully on async/await and drop support for Node 4 and 6.

Issue #3423, opened 2017-01-10
Remove "Vary: Origin" for wildcard CORS origins


rubennorte:
At the moment, enabling CORS in a route using `cors: true` or `cors: { origin: ['*'] }` sets the following response headers:

```
Vary: Origin
Access-Control-Allow-Origin: <origin-request-header-received>
...
```

This breaks caches unnecessarily as it could be done for all requests the same way:

```
Access-Control-Allow-Origin: *
```

I've found this issue because I have dynamic subdomains in my site and CloudFront has been caching different versions of each static file for each origin, when it's not necessary.

Would you accept a PR changing this behaviour only for routes configured that way? I'm pretty sure it requires just a small change in the code.

rubennorte:
I tried modifying the header in a `onPreResponse` extension but the code that handles the CORS options is executed after all extension points, so it's not possible.

As a workaround I've implemented the CORS headers myself but that way I loose all the features this options provides (OPTIONS method, validation, etc.).

kanongil:
I think a CDN-friendly mode would be a nice addition to Hapi, as it is something that I have had to work-around in the past.

I don't think it should be signalled as you suggest, rather it needs to be a distinct option, since it changes the way the CORS headers are validated and transmitted. Eg. it needs to _always_ add the CORS headers to responses to safely remove the `Vary: Origin` header. Even for non-CORS requests.

Note that Hapi used to work this way until it was completely reworked in one of the releases.

rubennorte:
> I don't think it should be signalled as you suggest, rather it needs to be a distinct option, since it changes the way the CORS headers are validated and transmitted. Eg. it needs to always add the CORS headers to responses to safely remove the Vary: Origin header. Even for non-CORS requests.

I understand it would be necessary to always set the CORS headers in this particular case, but is that a problem? I rather have some additional headers and be able to cache my responses.

IMO setting a new configuration option is adding unnecessary complexity.

kanongil:
> IMO setting a new configuration option is adding unnecessary complexity.

In general, Hapi prefers explicit options, rather than implicit as you suggest. It is also kind of required since there are valid use-cases for the config you mention, where you expect the existing behaviour. 

Finally, re-intepreting the meaning of the config is a breaking change.

rubennorte:
[...]

PR #3598, opened 2017-10-08
Implement wildcardIgnoresOrigin CORS option

rubennorte:
This PR allows the implementation of CDN-friendly CORS routes.

When `wildcardIgnoresOrigin` is set to `true` and the route allows any origin using the wildcard value (`['*']`), the `Origin` header of the request is ignored and `Access-Control-Allow-Origin: *` is send.

Example:

```javascript
server.route({
  method: 'GET',
  path: '/',
  handler: (request, reply) => reply('ok'),
  config: {
    cors: {
      origin: ['*'],
      wildcardIgnoresOrigin: true
    }
  }
});
```

Request:
```
GET / HTTP/1.1
...
```

Response:
```
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET
Access-Control-Allow-Headers: Accept, Authorization, Content-Type, If-None-Match
Access-Control-Expose-Headers: WWW-Authenticate, Server-Authorization
Access-Control-Max-Age: 86400
...

ok
```

Request:
```
OPTIONS / HTTP/1.1
Origin: https://example.com
Access-Control-Request-Method: GET
```

Response:
```
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET
Access-Control-Allow-Headers: Accept, Authorization, Content-Type, If-None-Match
Access-Control-Expose-Headers: WWW-Authenticate, Server-Authorization
Access-Control-Max-Age: 86400
...
```

Implements #3423


lib/cors.js
```diff
@@ -110,6 +109,12 @@ internals.handler = function (request, h) {
 
     // Validate Origin header
 
+    const origin = request.headers.origin;
+
+    if (!origin && !settings._ignoreOrigin) {
+        throw Boom.notFound('CORS error: Missing Origin header');
```
hueniverse: Isn't this a violation of the CORS spec?

kanongil: > A resource that is publicly accessible, with no access control checks, can always safely return an `Access-Control-Allow-Origin` header whose value is "*".

Eg. it can send an invalid `origin: *` header, when the response would also set `access-control-allow-credentials: true`.

This can be guarded for in the Joi schema, by not allowing `wildcardIgnoresOrigin` to be used with `credentials`.

rubennorte: I'm not sure that the spec forbids allowing credentials with a wildcard origin (I don't see anything in the spec pointing to it), but I think adding that restriction is a good security measure. The main use case for this flag is to improve caching of static files, and that still works if we add the restriction.

Should I add it then?

PR #3600, opened 2017-10-11
Prerequisite documentation

BigWillie:
Last night, I attended a talk given by https://github.com/codesleuth called "Being Asynchronous" which lead me to question how prerequisites work in Hapi. I had been told that they were Asynchronous, but the documentation says: 

https://hapijs.com/api#route-prerequisites

> To force the return value as the response and skip any other prerequisites and the handler, use the reply().takeover() method.

This appears to be a mistake. As I understand it you cannot skip functions run in parallel.

Prerequisites are not parallel. They are kicked off asynchronously. However, those prerequisite methods might have parallel operations within them. The Handler gets the reply once all those prerequisites are done. I would like to update the documentation to state this.

Many thanks,
Rob

(my first pull request on an open source project! :D)


AdriVanHoudt:
Hi @BigWillie 

It looks like you opened a PR to merge the master branch into the v16 branch. From your explanation I would guess you would want to do a PR to update the documentation? If so you will have to do a commit on your fork from hapi and then make a PR to merge your branch (with the updated docs) into the one from hapi.

BigWillie:
@AdriVanHoudt yep - I totally messed this up... So have forked it - and made a PR... dumb mistake. Sorry about this...

BigWillie:
Sorry I messed up... please ignore this PR and see my other one :) thanks!

AdriVanHoudt:
Np 👍

PR #3601, opened 2017-10-11
Updated documentation on prerequisites. They are kicked off asynchron…

BigWillie:
Last night, I attended a talk given by https://github.com/codesleuth called "Being Asynchronous" which lead me to question how prerequisites work in Hapi. I had been told that they were Asynchronous, but the documentation says:

https://hapijs.com/api#route-prerequisites

To force the return value as the response and skip any other prerequisites and the handler, use the reply().takeover() method.
This appears to be a mistake. As I understand it you cannot skip functions run in parallel.

Prerequisites are not parallel. They are kicked off asynchronously. However, those prerequisite methods might have parallel operations within them. The Handler gets the reply once all those prerequisites are done. I would like to update the documentation to state this.

Many thanks,
Rob

(my first pull request on an open source project! :D)

### DIFF
--- a/API.md
+++ b/API.md
@@ -2977,7 +2977,11 @@ Note that prerequisites do not follow the same rules of the normal
 [reply interface](#reply-interface). In all other cases, calling `reply()` with or without a value
 will use the result as the response sent back to the client. In a prerequisite method, calling
 `reply()` will assign the returned value to the provided `assign` key. If the returned value is an
-error, the `failAction` setting determines the behavior. To force the return value as the response
+error, the `failAction` setting determines the behavior.
+
+Prerequisites are not parallel. They are kicked off asynchronously. However, prerequisite methods can have parallel operations within them. The Handler gets the reply once all prerequisite metods are done.
+
+To force the return value as the response
 and skip any other prerequisites and the handler, use the `reply().takeover()` method.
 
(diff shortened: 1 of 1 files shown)


mtharrison:
Prerequisites which are defined in the config within in an array are executed in parallel. 

```js
server.route({
    method: 'GET',
    path: '/',
    config: {
        pre: [
            [
                // m1 and m2 executed in parallel
                { method: pre1, assign: 'm1' },
                { method: pre2, assign: 'm2' }
            ],
            { method: pre3, assign: 'm3' },
        ],

            return reply(request.pre.m3 + '\n');
        }
    }
});
```
If you call `takeover()` in one of pre1 or pre2, the value will be used as the response. Obviously as you noted you cannot skip the already executing functions, however any values they pass to `reply()` will be ignored and the handler and pre3 will not execute.

hueniverse:
Thanks but this is not exactly right. It will get rewritten anyway as part of the v17 work.

Codesleuth:
Any news on this? It's technically incorrect to say ...

>those methods are called in parallel

... as these are issued as microtasks that will be processed sequentially, in guaranteed order. Any further actions within each promise, such as HTTP or file I/O, can then be considered asynchronous, but the JavaScript code is itself not executed in parallel.

hueniverse:
It's as parallel as node can get. I don't see the issue here.

PR #3602, opened 2017-10-11
exclude connection header in response passThrough

jordanrogers:
When a hapi.js server is acting as a proxy or gateway, and a handler calls the reply/responder interface with a stream that is an IncomingMessage from an upstream, I believe the default behavior of the passThrough mode should be to not copy the connection header.

The semantics of the Connection header are that it represents per-hop connection options, and that a proxy or gateway MUST remove any connection options before forwarding a message (in this case a response to its client).  Basically connection options (usually keep-alive or close) should be negotiated between the server/client and between the server/upstream independent of each other.

The semantics of the Connection header are actually a little more complicated than that, because the header is supposed to be able to container tokens representing *other* headers that are per-hop. But I do not have that use case myself, so I did not undertake to implement that.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -464,7 +464,10 @@ exports = module.exports = internals.Response = class {
 
                     for (let i = 0; i < headerKeys.length; ++i) {
                         const key = headerKeys[i];
-                        this.header(key.toLowerCase(), Hoek.clone(this.source.headers[key]));     // Clone arrays
+                        const lower = key.toLowerCase();
+                        if (lower !== 'connection'){                                      // Do not copy per-hop connection options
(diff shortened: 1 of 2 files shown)


kanongil:
If you want to be thorough, there are a several other headers that need to go as well, like `Keep-Alive` & `Transfer-Encoding`.

These are listed nicely in the deprecated https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.5.1. Additionally, you need to remove any headers listed in the value of the connection header according to the current RFC: https://tools.ietf.org/html/rfc7230#section-6.1.

PR #3660, opened 2017-10-24
Restore parallel cookie values for state

kanongil:
The v17 upgrade removed the parallel computation.


hueniverse:
Is this really needed? How crazy are your autoValue functions?

kanongil:
Don't know really, I just noticed the regression. Feel free to drop this PR.

hueniverse:
I'll just document this in the release notes.

PR #3663, opened 2017-10-27
Loosen sample and modify peer validator in the routeBase schema

### LINKED ISSUE #3567: response validate options for 'modify' and 'sample' is too strict

Describe your issue here, include as much detail as neccessary to reproduce the issue
or implement a missing functionality.

```js
// set connection-level configurations
server.connection({
  ...,
  route: {
    ...,
    response: {
      modify: true
    }
  }
})

// set route-level configurations
server.route([{
  ...,
  configs: {
    response: {
      modify: false,
      sample: 0,
      schema: Joi.object({ ... })
    }
  }
}])
```

   
#### What was the result you received?
```
[1] "modify" conflict with forbidden peer "sample"
```

#### What did you expect?
When we set `modify` false, we can use `sample` option.
However, the validator forbids on the `key peers`, not the value.
Could it be relieved by replacing them with `Joi.alternative().when()`? 

#### Context

* *node version*: 8
* *hapi version*: 16.5.2

rokoroku:
It fixes #3567

- Make `routeBase` schema allows `sample` props when `modify` is set to false
- Make `options.stripUnknown` assertion allows `false` value when `modify` is set to false

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -163,13 +163,12 @@ internals.routeBase = Joi.object({
         modify: Joi.boolean(),
         options: Joi.object().default(),
         ranges: Joi.boolean().default(true),
-        sample: Joi.number().min(0).max(100),
+        sample: Joi.number().min(0).max(100).when('modify', { is: true, then: Joi.forbidden() }),
         schema: Joi.alternatives(Joi.object(), Joi.array(), Joi.func()).allow(true, false),
         status: Joi.object().pattern(/\d\d\d/, Joi.alternatives(Joi.object(), Joi.array(), Joi.func()).allow(true, false))
     })
         .default()
-        .without('modify', 'sample')
-        .assert('options.stripUnknown', Joi.when('modify', { is: true, otherwise: Joi.forbidden() }), 'meet requirement of having peer modify set to true'),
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -163,13 +163,12 @@ internals.routeBase = Joi.object({
...
         ranges: Joi.boolean().default(true),
-        sample: Joi.number().min(0).max(100),
+        sample: Joi.number().min(0).max(100).when('modify', { is: true, then: Joi.forbidden() }),
         schema: Joi.alternatives(Joi.object(), Joi.array(), Joi.func()).allow(true, false),
         status: Joi.object().pattern(/\d\d\d/, Joi.alternatives(Joi.object(), Joi.array(), Joi.func()).allow(true, false))
     })
         .default()
-        .without('modify', 'sample')
-        .assert('options.stripUnknown', Joi.when('modify', { is: true, otherwise: Joi.forbidden() }), 'meet requirement of having peer modify set to true'),
+        .assert('options.stripUnknown', Joi.when('modify', { is: true, otherwise: [Joi.boolean().valid(false), Joi.forbidden()] }), 'meet requirement of having peer modify set to true'),
```
hueniverse: Do you need to keep the forbidden as alternative?

rokoroku: I just keep the existing one while allowing "false" value.

Do you want me to remove the forbidden as an alternative here?

hueniverse: Actually, this is correct.

hueniverse: It's not needed because the default is optional.

PR #3666, opened 2017-10-31
Fixed 'server.once is not a function' error in the doc

midnightcodr:


API.md
```diff
@@ -1686,9 +1686,9 @@ const Hapi = require('hapi');
 const server = Hapi.server({ port: 80 });
 
 server.event('test');
-server.once('test', (update) => console.log(update));
-await server.emit('test', 'hello');
-await server.emit('test', 'hello');       // Ignored
+server.events.once('test', (update) => console.log(update));
+await server.events.emit('test', 'hello');
+await server.events.emit('test', 'hello');       // Ignored
```
hueniverse: `server.emit()` is a valid function. Please revert.

midnightcodr: @hueniverse I've created this gist https://gist.github.com/midnightcodr/c99884ea277d07973969b479e281e39f to show how `server.emit is not a function` error is triggered. Can you take a look? Maybe I am not coding correctly. Thanks.

hueniverse: My bad. I remembered documenting it but that was for the full version.

PR #3684, opened 2017-11-19
remove await server.start(); in docs

johnbrett:
I've gotten messages for new users trying to run examples from docs where this is failing, it needs to be removed as it doesn't work and is causing confusion.

### DIFF
--- a/API.md
+++ b/API.md
@@ -267,7 +267,7 @@ internals.start()
 
         console.error(err);
         process.exit(1);
-    })
+    });
 ```
 
 ### <a name="server.options" /> Server options
@@ -2471,7 +2471,12 @@ internals.start = async () => {
 }
 
 internals.start()
-  .then((server) => console.log('Server started at: ' + server.info.uri));
+  .then((server) => console.log('Server started at: ' + server.info.uri))
+  .catch(err => {
+
+        console.error(err);
+        process.exit(1);
+  });
 ```
 
 ### <a name="server.state()" /> `server.state(name, [options])`
--- a/API.md
+++ b/API.md
@@ -1459,7 +1459,7 @@ server.route({
     handler: { test: { msg: 'test' } }
 });
 
-await server.start();
+server.start();
 ```
 
 The `method` function can have a `defaults` object or function property. If the property is set to
@@ -1841,7 +1841,7 @@ server.ext({
 });
 
 server.route({ method: 'GET', path: '/test', handler: () => 'ok' });
-await server.start();
+server.start();
 
 // All requests will get routed to '/test'
 ```
@@ -1866,7 +1866,7 @@ server.ext('onRequest', function (request, h) {
 });
 
 server.route({ method: 'GET', path: '/test', handler: () => 'ok' });
-await server.start();
+server.start();
 
 // All requests will get routed to '/test'
 ```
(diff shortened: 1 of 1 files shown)


hueniverse:
The solution is to wrap it in a function, not to remove it.

corbinu:
@johnbrett I believe this would replicate the behavior most expect:
```
(async () => {
    return await server.start();
}).catch((err) => {
    console.log(err);
    process.exit(1);
});
```

mtharrison:
We should try to establish something consistent for this. We've got around 5 different styles emerging now.

devinivy:
Agree!  Worth nothing that unhandled rejections are deprecated, and that in the future node will crash the process just like an uncaught exception– but I don't know the timeline for that.

corbinu:
@mtharrison agreed if there is a standard format obviously that is better
@devinivy I believe that has already occurs in 8 but don't quote me on that. That is why my version logs the error and then exits.

hueniverse:
```js
internals.start = async function () {

    // Create instance, await start, etc.
};

internals.start();
```

corbinu:
@hueniverse I am sorry but a whole hardheartedly disagree if it was up to me the rejection would be better handled by node. However for somebody first coming to node hapi should not be an unexpected/painful experience so the extra line is beyond worth the chance of driving somebody way from node forever.

johnbrett:
I modified the first snippet in the API page, as well as the `server.start()` example to include a full example where rejections are handled.

csrl:
Why bother with the wrapper in the examples? Seems overly verbose.

```javascript
const Hapi = require('hapi');
const server = Hapi.server({port: 8080});

server.start().then(() => {

    console.log('Server started at: ' + server.info.uri)
}).catch(err => {

    console.error(err);
    process.exit(1);
});
```

Although it would be nice if server.start() returned itself, so we could just use something like:

```javascript
const Hapi = require('hapi');

Hapi.server({port: 8080}).start().then(server => {

    console.log('Server started at: ' + server.info.uri)
}).catch(err => {

    console.error(err);
    process.exit(1);
});
```

(or modified for appropriate hapi styling...)

Obviously for those already having an async function wrapper, calling  `await server.start()` is cleanest, but for the example code, I don't see why exampling the abstraction everywhere is necessary.

Also, the v17 documentation has ` await server.start([callback])`  but it does not actually have an optional callback parameter.

hueniverse:
@corbinu I don't want people starting with node coming here for help. I am ok being unhelpful to those are are not sufficiently proficient with the core language and its new features. hapi is never going to be the community you go to for starting with node.

I actually think the docs are fine as-is. I think the full examples/tutorials could be more cut/paste ready but wrapping every code example in a function just to make `async` work seems absurd to me. The snippets are here to demonstrate a single function. Most won't work cause they lack other required server setup.

hueniverse:
I don't want any `then()` or `catch()` in the docs. It's an anti-pattern when using async/await.

hueniverse:
Going to solve this differently. Thanks!

csrl:
So based on that styling feedback, a minimal v17 tutorial example would be:

```javascript
const Hapi = require('hapi');
const internals = {};

internals.start = async function () {
    const server = Hapi.server({port: 8080});
    try {
        await server.start();
    }
    catch (e) {
        console.error(err);
        process.exit(1);
    }
    console.log(`Server started at: ${server.info.uri}`);
};

internals.start();
```

hueniverse:
@csrl yep. That's the right style for tutorials. For this document, I am just going to wrap all await with async example method.

PR #3703, opened 2017-12-01
Incorrect function signature in plugin examples?

thebinarypenguin:
The docs say the register function should have the signature `async function(server, options)` but the examples have the signature `function(server, options)`


Marsup:
You don't need async if there's nothing async.

devinivy:
And the reason for that is that "regular" functions are `await`-able, since you can always await a plain value.

```js
const fn = () => 42;

const result = await fn();
//    result = await 42;
//    result = 42;
```

Issue #3701, opened 2017-11-29
Enable registered encoders to have higher priority than built-in


kanongil:
See https://github.com/hapijs/discuss/issues/589 for details.

hueniverse:
Any reason not to just give higher priority to the last user-added encoder?

kanongil:
@hueniverse That will solve the immediate issue quite well.

This solution I explored in the discussion looks further, but I agree that it is unlikely to be an issue in practice (at least for now).

hueniverse:
Closed by #3717

PR #3717, opened 2017-12-20
Prefer user added encoding over built-in

benib:
This implements #3701 and resolves https://github.com/hapijs/discuss/issues/589

Please let me know if I missed anything. And thanks a lot for the great framework ❤️


kanongil:
Thanks, can you add a test as well?

benib:
@kanongil i changed the existing test to test if the added encoder is preferred over built in ones by sending `accept-encoding: gzip, deflate, test` instead of `accept-encoding: test`.

Would you prefer to have a separate test instead?

kanongil:
Should be fine. The important part is that it would fail with the old code.

hueniverse:
I consider this a bug fix since the only reason to add a custom encoder is to give it higher priority. But also marked it as breaking as it does change existing behavior.

PR #3722, opened 2018-01-04
Update license year

Yahkob:


hueniverse:
I do this when I make an actual change.

Yahkob:
One day I'll get that license change commit 😢

Issue #3661, opened 2017-10-25
Intentions for security fixes for pre-v17  hapi


midknight41:
The new **v17 looks fantastic** and I've certainly keen to migrate existing APIs to the new version as soon as is practically possible. Thanks for your hard work!

I'm interested if you have a position of how you would respond to a security issue on the v16 line *after* you launch v17? I'm concerned that if we haven't finished upgrading to the latest version of hapi and node.js that we'll be in a bit of a tight spot.

Thanks for your time.

Marsup:
Maintenance will be carried as long as node is LTS.

midknight41:
I'm really pleased to hear that. Thanks for your help!

hueniverse:
Yeah, we will continue to support v16 for critical bugs and security issues for at least a year. Not specific dates yet, but this is largely based on resources. Companies who care about it should consider investing some resources in it - directly by letting employees contribute time, or indirectly by making a financial contribution. We will have more specific info about it when v17 is officially out.

PR #3730, opened 2018-01-27
Prefer user added encoding over built-in

cvbuelow:
Adding [bug fix in v17](https://github.com/hapijs/hapi/pull/3717) to v16


cvbuelow:
@hueniverse any chance of getting this merged? Would be great to have this fix in v16...

kanongil:
Thanks for taking the time to do this. However, this is not a security fix, and I dare say, not a critical bugfix, so don't expect it to be merged to the LTS branch. See https://github.com/hapijs/hapi/issues/3661#issuecomment-339437548 for the stated v16 support policy. 

If you need to continue on v16, I suggest you either maintain your own custom version, or monkeypatch the runtime with a fix.

PR #3731, opened 2018-01-27
Simplify code to get coverage to 100%

cvbuelow:


hueniverse:
You don't make things slower for coverage.

PR #3741, opened 2018-02-18
Replace new Buffer()

idandagan1:
Hi,
This is my first PR to hapi.js, after I posted in the discuss project ([Link](https://github.com/hapijs/discuss/issues/631)) 

This PR is related to the issue with using `new Buffer()`. As you probably know, `new Buffer()` has been deprecated due to security issues and replaced by `Buffer.from()`.

The changes:
- replaced all new Buffer(string) with Buffer.from(string)
- replaced all new Buffer(size) with Buffer.alloc(size)

I followed the "contributing guidelines" including tests.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -44,7 +44,7 @@ const internals = {
         'start',
         'stop'
     ],
-    badRequestResponse: new Buffer('HTTP/1.1 400 Bad Request\r\n\r\n', 'ascii')
+    badRequestResponse: Buffer.from('HTTP/1.1 400 Bad Request\r\n\r\n', 'ascii')
 };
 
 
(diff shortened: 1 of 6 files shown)


devinivy:
It appears that the failures here are primarily due to the fact that the npm shrinkwrap needs updates (bounce to v1.2.0 so the tests can use `background()` and hoek to v5.0.3 to pass the nsp checks).  I can confirm that a fresh install without the shrinkwrap passes all tests on node v8.9.4.

idandagan1:
@devinivy should I run full update for the shrinkwrap file? or just for those you mentioned? (bounce and hoek)

PR #3746, opened 2018-02-21
Updates Hoek dep to fix NSP vulnerability.

Alex-Krautmann:
Patches Hoek to 4.2.1 which fixes https://nodesecurity.io/advisories/566

I also fixed some missing coverage.


WesTyler:
The `^` syntax in the package.json means the patch will already be included in fresh installs.

Alex-Krautmann:
Good point.

PR #3766, opened 2018-03-24
Add PATCH to `created` status, better error msg

bioball:
The `created` status code helper should also be valid at least for `PATCH` calls. 

BTW: Should this assertion even exist? I'd be in favor of getting rid of it and allow developers break the rules as they see fit.


hueniverse:
You can still manually do what this does and violate HTTP rules on your own...

bioball:
Cool. Thanks!

PR #3775, opened 2018-04-06
Add support for Referrer-Policy header

geek:


test/headers.js
```diff
@@ -438,6 +438,41 @@ describe('Headers', () => {
             expect(res.headers['x-download-options']).to.equal('noopen');
             expect(res.headers['x-content-type-options']).to.equal('nosniff');
         });
+
+        it('does not return the referrer-policy header whe security.referrer is false', async () => {
```
devinivy: It might be good to also ensure that the default behaves like `false`, since it would not likely be picked-up by another test.

geek: https://github.com/hapijs/hapi/pull/3775/files#diff-fecfa4a97c0ff7750b8c1e1549f7f3faR306 like this?

geek: @devinivy added another test for the default and fixed that `whe` typo.

API.md
```diff
@@ -3394,6 +3394,18 @@ following options:
 - `noSniff` - boolean controlling the 'X-Content-Type-Options' header. Defaults to `true` setting
   the header to its only and default option, `'nosniff'`.
 
+- `referrer` - controls the ['Referrer-Policy'](https://www.w3.org/TR/referrer-policy/) header value:
+    - `false` - the 'Referrer-Policy' header will not be sent with responses. This is the default value.
+    - `''` - empty string indicating that the Referrer-Policy will be defined elsewhere.
+    - `'no-referrer'` - never include the referrer header.
```
johnbrett: @geek this is question more than feedback on PR - what is the difference here between `false` and `'no-referrer'`?

hueniverse: The description is misleading...

API.md
```diff
@@ -3394,6 +3394,18 @@ following options:
 - `noSniff` - boolean controlling the 'X-Content-Type-Options' header. Defaults to `true` setting
   the header to its only and default option, `'nosniff'`.
 
+- `referrer` - controls the ['Referrer-Policy'](https://www.w3.org/TR/referrer-policy/) header value:
+    - `false` - the 'Referrer-Policy' header will not be sent with responses. This is the default value.
+    - `''` - empty string indicating that the Referrer-Policy will be defined elsewhere.
```
hueniverse: "elsewhere"?

API.md
```diff
@@ -3394,6 +3394,18 @@ following options:
...
+    - `'no-referrer'` - never include the referrer header.
+    - `'no-referrer-when-downgrade'` - never include the referrer when navigating from HTTPS to HTTP.
+    - `'same-origin'` - only include the referrer on the current site origin.
+    - `'origin'` - include the referrer but strip off path information so that the value is the current origin only.
+    - `'strict-origin'` - same as `'origin'` but the referrer will be omitted when going from HTTPS to HTTP.
+    - `'origin-when-cross-origin'` - the referrer includes the full path for same-origin requests but only the origin components of the URL are included for cross origin requests.
+    - `'strict-origin-when-cross-origin'` - same as `'origin-when-cross-origin'` but the referrer will be omitted when going from HTTPS to HTTP.
+    - `'unsafe-url'` - the referrer will always be included with the full URL.
+
```
hueniverse: Basically, the descriptions are confusing because they are about the HTTP policy, not the hapi server actions. Needs to rewrite them to read "Informs the client to...".

API.md
hueniverse: Also, can this be `"empty"` instead?

geek: Yes, we can change this, but it won't be consistent with the referrer policy options, which is to send a literal empty string. Up to you, do you want me to change this?

hueniverse: It is just so odd to have an empty string as value. We can start with it as is and see if people are confused by it.

geek: "elsewhere" is the language used by w3, which really means in html meta-headers. I'll make this clearer

API.md
geek: makes sense, I'll update

PR #3776, opened 2018-04-09
correct h.continue in api.md 

dragonwong:
should `return h.continue()` but not `h.continue`


dragonwong:
@devinivy Oh yes, I just found I am using an older version. Thanks for your explanation!

devinivy:
No problem!

PR #3779, opened 2018-04-19
Allow undefined handler and prehandler response values

txase:
There isn't anything inherently wrong with a handler or pre-handler that
doesn't need to modify the response. As an example, a pre-handler may
simply enforce RBAC privileges:

```JS
const checkWriteAccess = request => {
  if (request.params.admin) {
    throw Boom.unauthorized();
  }
}
```

(This is admittedly bad security design, but imagine a proper RBAC
mechanism)

Likewise, a handler that doesn't return a value may be used for a route
that returns an empty 204 response.

### DIFF
--- a/lib/toolkit.js
+++ b/lib/toolkit.js
@@ -48,10 +48,6 @@ exports = module.exports = internals.Manager = class {
 
         // Process response
 
-        if (response === undefined) {
-            response = Boom.badImplementation(`${method.name} method did not return a value, a promise, or throw an error`);
-        }
-
         if (options.continue &&
             response === this.continue) {
 
--- a/test/toolkit.js
+++ b/test/toolkit.js
@@ -10,7 +10,6 @@ const Handlebars = require('handlebars');
 const Hapi = require('..');
 const Inert = require('inert');
 const Lab = require('lab');
-const Teamwork = require('teamwork');
 const Vision = require('vision');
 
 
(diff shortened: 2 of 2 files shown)


devinivy:
Requiring an explicit return value of `h.continue` ensures that the developer didn't make a mistake (e.g. miss a case), causing an unintentionally empty response.  I imagine this is especially useful for folks migrating from hapi v16 with callback usage to hapi v17 with async/await.  At the same time, I understand the desire to cut loose from the safeguard—I suggest writing a little helper that wraps handlers/prereqs/extensions and implicitly returns `null` or `h.continue` for you.

```js
const checkWriteAccess = allowUndefined(request => {
  if (request.params.admin) {
    throw Boom.unauthorized();
  }
})
```

txase:
I don't expect to change any minds. I did take @devinivy's suggestion, though I tweaked it by processing all of our route configs before registering the routes, transforming `undefined` responses into `h.continue` responses.

The reason I feel this is a bad design goal is that it changes a fundamental expectation of how code should behave. JavaScript, as a dynamically typed language, has the ability to write functions that do not return any value. This is one small part of why many people prefer the language. But here, Hapi is forcing the developer to write functions like they are statically typed and must return a specific value in all cases.

As I lead an engineering team, comprised of individuals who all spend time in JS, but not all that time in the API, this Hapi interface adds extra unnecessary complexity to API development. Anyone working on the API now needs to know of this special behavior that Hapi requires. I understand the idea of trying to protect devs from making mistakes, but if that requires writing code differently than for any other project then I think the design has overshot the goal.

I'm fine if this PR is closed and locked. Hapi has the ability to decide what its goals are and stick to them without bike shedding. My suggestion would be to rethink this one, though.

hueniverse:
An `undefined` return value is almost always a developer error. If you use async/await and do not return a value, you are almost certainly going to have a bug that's hard to otherwise notice.

Issue #3718, opened 2017-12-21
Request decorations leak across server instances in same process


hueniverse:

SimonSchick:
I'm running into this issue when running unit tests, particularly when I register plugins multiple times with different server instances.

The error I run into most of the time is `Toolkit decoration already defined: NAME`.

What are your thoughts on adding an `ignoreExisting` option?

SimonSchick:
ty

PR #3795, opened 2018-05-30
do not attach request decorations to the prototype

cjihrig:
This commit prevents request decorations from being attached
to the `Request` prototype. This was leading to decorations being
leaked between servers in the same process.

Fixes: https://github.com/hapijs/hapi/issues/3718

Note: I haven't benchmarked this change.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -79,9 +79,14 @@ exports = module.exports = internals.Request = class {
     static generate(server, req, res, options) {
 
         const request = new internals.Request(server, req, res, options);
+        const proto = Object.getPrototypeOf(request);
+
+        // Give request access to all decorations without adding them to
+        // the Request prototype.
+        Object.setPrototypeOf(request, server._core._decorations.request);
+        Object.setPrototypeOf(server._core._decorations.request, proto);
 
         // Decorate
-
         if (server._core._decorations.requestApply) {
             const properties = Object.keys(server._core._decorations.requestApply);
             for (let i = 0; i < properties.length; ++i) {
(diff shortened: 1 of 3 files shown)


lib/request.js
```diff
@@ -79,9 +79,14 @@ exports = module.exports = internals.Request = class {
     static generate(server, req, res, options) {
 
         const request = new internals.Request(server, req, res, options);
+        const proto = Object.getPrototypeOf(request);
 
-        // Decorate
+        // Give request access to all decorations without adding them to
+        // the Request prototype.
+        Object.setPrototypeOf(request, server._core._decorations.request);
+        Object.setPrototypeOf(server._core._decorations.request, proto);
```
devinivy: Are there upsides to this versus the approach taken with `requestApply`?

cjihrig: You mean looping over the decorations? There shouldn't be. Depending on the number of decorators, this might be faster since you're just adjusting some pointers vs. looping. I mostly went with this approach since it seemed more similar to the existing implementation (in case code was relying on decorations being on the prototype).

devinivy: Ah, that makes a lot of sense.  I'm a little concerned by the performance warning MDN gives about [`setPrototypeOf()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf) (which I can't personally confirm or deny for current-day V8).  But also, setting the prototype of `server._core._decorations.request` to the same prototype on each request seems like it may be avoidable.  Here's one idea for an alternative implementation that might have less question marks re: performance.  https://github.com/hapijs/hapi/compare/master...devinivy:3718-alternative

cjihrig: I like that approach better. I'm going to close this.

devinivy: I made a PR taking that approach in #3796—feel free to suggest improvements.  I nabbed the test you had written here—if you'd like a proper "commit" credit for that, I'm happy to take the time.

PR #3812, opened 2018-08-11
Add support for Content-Security-Policy header directives

georgewatts:
Add support for content-security-policy header with unit testing and documentation.

These are necessary for browsers to verify origins and block the loading of undesirable content on webpages.


georgewatts:
Thanks for the swift reply @nlf!

Our problem is that we find the CSP header getting scrubbed from the response when handling errors, after playing around with the Hapi code we found this to work. 

Was there any particular reason for you guys keeping the CSP stuff separate?

Issue #3813, opened 2018-08-15
redirecting in onRequest throws Cannot read property 'isOriginMatch' of null


jgallen23:
We have a https redirect plugin that onRequest, if not https, then redirect to https (https://github.com/bendrucker/hapi-require-https).  As of 17.5.3, this causes the server to crash.  I pinned our server back to 17.5.2 and there was no issue.

Here is example stack trace
```
(node:24804) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'isOriginMatch' of null
    at exports.headers (/home/nicolas/dev/yellow/mistertracer/mistertracer-server/node_modules/hapi/lib/cors.js:171:28)
    at Object.internals.marshal (/home/nicolas/dev/yellow/mistertracer/mistertracer-server/node_modules/hapi/lib/transmit.js:46:24)
    at Object.exports.send (/home/nicolas/dev/yellow/mistertracer/mistertracer-server/node_modules/hapi/lib/transmit.js:31:25)
    at Request._reply (/home/nicolas/dev/yellow/mistertracer/mistertracer-server/node_modules/hapi/lib/request.js:349:24)
    at <anonymous>
(node:24804) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:24804) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

more context: https://github.com/hapijs/discuss/issues/704

mariecl:
I am also experiencing this issue.

In my case, one route handler calls the handler of another route under certain conditions. When those conditions are met, the "redirect" happens and I get the same `TypeError`.

Issue #3792, opened 2018-05-17
Global cors and security options not respected in 404


erfanio:
#### What are you trying to achieve or the steps to reproduce?
From API docs:
> If set to `'ignore'`, any incoming Origin header is ignored (present or not) and the 'Access-Control-Allow-Origin' header is set to `'*'`

I have this in my hapi server configs:
```js
...
    routes: {
      cors: { origin: 'ignore' },
...
```

This works fine most of the time
![image](https://user-images.githubusercontent.com/9994172/40184453-ab74d2a4-5a33-11e8-82dd-11d405938d8a.png)

But when a route doesn't exist, the 404 response doesn't include the cors headers.
![image](https://user-images.githubusercontent.com/9994172/40184553-e4a3678e-5a33-11e8-9700-3a1a558216f3.png)

#### What did you expect?
I expect to have `access-control-allow-origin: *` on the 404

#### Context
* *node version*: 10
* *hapi version*: 17.4
* *os*: Linux

vladikoff:
I was about to file this bug, but saw yours.

This can be reproduced with this: 

```
git clone https://github.com/vladikoff/issue-hapi-17-404-headers.git
cd issue-hapi-17-404-headers
node index.js
```

Navigate to http://localhost:3000

### Existing route:

![image](https://user-images.githubusercontent.com/128755/41607041-1c4dd0a0-73b3-11e8-8b45-ba88c61b057f.png)


### 404 route: http://localhost:3000/404something

Response headers are missing security route options: 

![image](https://user-images.githubusercontent.com/128755/41607106-37b50fc0-73b3-11e8-9317-c0fbfa17e2c6.png)

This seems like a regression in Hapi 17. In hapi 16 this worked fine, we have testa covering this behaviour.

vladikoff:
Sample code:

```js
'use strict';

const Hapi = require('hapi');

const server = Hapi.server({
  port: 3000,
  host: 'localhost',
  routes: {
    cors: true,
    payload: {
      maxBytes: 16384
    },
    security: {
      hsts: {
        maxAge: 15552000,
        includeSubdomains: true
      },
      xframe: true,
      xss: true,
      noOpen: false,
      noSniff: true
    },
  }
});

server.route({
  method: 'GET',
  path: '/',
  handler: (request, h) => {

    return 'Hello, world!';
  }
});

server.route({
  method: 'GET',
  path: '/hello',
  handler: (request, h) => {

[...]

PR #3816, opened 2018-08-21
Ensure request.info.cors is set for special routes

devinivy:
Enclosed is a fix for #3813.

When an error occurs in `onRequest` hapi would throw because it expected `request.info.cors` to exist, but the route (and its CORS info) naturally hadn't been looked-up yet.  The bug was introduced in the fix for #3792, which made server CORS defaults apply to the special "notFound" route, which is the effective route during `onRequest`.

The fix here is to set `request.info.cors` prior to `onRequest` based upon the server's default CORS settings.  An alternative fix would be to compute `isOriginMatch` when attempting to set CORS headers specifically in the case of a special "notFound", rather than attempting to read it off of `request.info.cors`.

### DIFF
--- a/API.md
+++ b/API.md
@@ -4621,9 +4621,10 @@ Request information:
 - `cors` - if CORS is enabled for the route, contains the following:
     - `isOriginMatch` - `true` if the request 'Origin' header matches the configured CORS
       restrictions. Set to `false` if no 'Origin' header is found or if it does not match.
-      Note that this is only available after the `'onRequest'` extension point as CORS is
-      configured per-route and no routing decisions are made at that point in the request
-      lifecycle.
+      Note that this set based upon the server's CORS defaults
+      ([`server.options.routes`](#server.options.routes)) until after the `'onRequest'` extension
+      point, as CORS is configured per-route and no routing decisions are made at that point in
+      the request lifecycle.
 
 - `host` - content of the HTTP 'Host' header (e.g. 'example.com:8080').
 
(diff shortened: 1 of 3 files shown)


kanongil:
Another option is just not add cors headers for errors this early. The linked issue does not expect or require CORS headers.

Also, we already do this for [errors that occur before `onRequest`](https://github.com/hapijs/hapi/blob/master/lib/core.js#L504).

devinivy:
That is true.  I don't have a strong opinion on the matter—primarily was hoping to tease out the cause of the error.  The handling of client errors is an interesting precedent.  In my opinion it's more important that we treat CORS in whatever way that makes the most sense to us, rather than try to stick with the precedent set by client error handling.

kanongil:
From my POV, it makes sense to not apply CORS headers in this case. It is hard to reason about what the values should be before the host and method has been resolved in `onRequest`. It seems reasonable to omit them for requests that never reach the "Route lookup" step.

kanongil:
Actually, this matches the JSONP handling in `onRequest`:

> JSONP configuration is ignored for any response returned from the extension point since no route is matched yet and the JSONP configuration is unavailable.

hueniverse:
I think the JSONP approach is wrong. When I wrote it, there were no server defaults and because I hate JSONP so much, I have not looked at it since I wrote it, other than dragging it along. But the right approach is to use the server defaults for the bad request and not found routes, as well as for anything that is returned in onRequest. It's the intuitive approach.

hueniverse:
That said, it should not set `info.cors` with what is potentially incorrect information. So I would keep 'info.cors' undefined but still return the headers.

PR #3822, opened 2018-09-11
Use WHATWG URL for request.url

kanongil:
I had a look into what it would take to update hapi to use the [WHATWG URL API](https://nodejs.org/api/url.html#url_the_whatwg_url_api) since `Url.parse()` [is being deprecated](https://github.com/nodejs/node/pull/22715), and node 8 already supports the replacement API.

As far as I can tell, the only directly affected api is the `request.url` property, and the `setUrl()` method.

PR details:

 * All usage of old style **URL** objects, and `Url.parse()` has been replaced with the new **URL** API. There is no legacy support.
 * Added a `route.queryParser` option to retain ability to customize the query parsing. The return value is validated so that `request.query` will always be an object.
 * `request.url` property is no longer relative, since the **URL** constructor won't create relative URLs.
 * `setUrl` throws a `badRequest` error if the URL parsing fails.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -157,6 +157,7 @@ internals.routeBase = Joi.object({
     })
         .default(),
     plugins: Joi.object(),
+    queryParser: Joi.func().allow(null).default(null),
     response: Joi.object({
         emptyStatusCode: Joi.valid(200, 204).default(200),
         failAction: internals.failAction,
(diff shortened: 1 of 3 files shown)


hueniverse:
I changed it to be more consistent and less flexible.

PR #3872, opened 2018-11-07
add linting

phapdinh:
This will add linting to your project because I saw some unused variables in there.


hueniverse:
We already have linting. Our rules are just different.

PR #3879, opened 2018-11-17
Set cookie autoValue only when no value received

### LINKED ISSUE #3831: State autoValue function overwrites existing cookie value
#### Are you sure this is an issue with the hapi core module or are you just looking for some help?
Yeah—either an issue with hapi core, or a doc issue within API.md.

#### Is this a security related issue?
Nah.

I’ve been configuring a state `autoValue`.

My goal is this: if the cookie does not exist, it should be set with an automatically generated value. If it does exist, the value should be left as-is.

The documentation in [API.md](https://github.com/hapijs/hapi/blob/7ffd301a3e0499df4c0989bd3aca7498bf314ee3/API.md#-serverstatename-options) describes this option:

> `autoValue` - if present _**and the cookie was not received from the client**_ or explicitly set by the route handler, the cookie is automatically added to the response with the provided value. The value can be a function with signature `async function(request)`

#### What was the result you received?
Using this sample script:

```js
const Hapi = require('hapi');

const server = Hapi.server({
    port: 3000,
    host: 'localhost'
});

server.state('test-cookie', {
    isSecure: false,
    autoValue: async () => `${new Date().getTime()}`,
});

server.route({
    method: 'GET',
    path: '/',
    handler: () => 'Test Cookie',
});

const init = async () => {
    await server.start();
    console.log(`Server running at: ${server.info.uri}`);
};

init();
```

1. First request made without a cookie: a new cookie is set, with value from `autoValue`
2. Second request made with cookie: cookie is overwritten, with value from `autoValue`

screen shot 2018-10-17 at 12 29 13 pm

screen shot 2018-10-17 at 12 29 22 pm

#### What did you expect?
1. First request made without a cookie: a new cookie is set, with value from `autoValue`
2. Second request made with cookie: cookie value is left as-is

#### Context

* *node version*: 10.12.0
* *hapi version*: 17.6.0
* *os*: Mac OSX High Sierra (10.13.6)

dominykas:
Fixes #3831

### DIFF
--- a/lib/headers.js
+++ b/lib/headers.js
@@ -77,22 +77,18 @@ exports.content = async function (request) {
 exports.state = async function (request) {
 
     const response = request.response;
-    const names = {};
     const states = [];
 
     for (const stateName in request._states) {
-        names[stateName] = true;
         states.push(request._states[stateName]);
     }
 
     try {
         for (const name in request._core.states.cookies) {
             const autoValue = request._core.states.cookies[name].autoValue;
-            if (!autoValue || names[name]) {
+            if (!autoValue || name in request._states || name in request.state) {
                 continue;
             }
-
-            names[name] = true;
 
             if (typeof autoValue !== 'function') {
                 states.push({ name, value: autoValue });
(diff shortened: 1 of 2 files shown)


dominykas:
I get the node 11 failure on `master` too, so assuming it's unrelated to this PR.

PR #3894, opened 2018-12-17
Replace config with options to match newer Hapi naming

mariecl:
The documentation headers all use `route.options.something` but some examples were still using `config` instead of `options`. I updated the remaining ones.


Marsup:
Thanks!

PR #3897, opened 2019-01-07
Return credentials and artifacts from server.auth.test()

marcuspoehls:
**This PR includes a breaking change. If merged, this should be released with v18.**

Previously, the credentials from the tested authentication strategy were passed through. This PR changes the return value of `server.auth.test()`.

Changes:
- `server.auth.test()` now returns an object with the properties
  - `credentials (required)
  - `artifacts` (optional)

---

PS: If merged, I’ll send another PR to update the docs

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -94,7 +94,7 @@ exports = module.exports = internals.Auth = class {
             throw response.error;
         }
 
-        return response.data.credentials;
+        return response.data;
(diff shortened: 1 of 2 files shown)


marcuspoehls:
@hueniverse Eran, thank you for the additional commit to update the `API.md`! Thanks for merging :blush:

Issue #3899, opened 2019-01-09
v16: look into compatibility with node 10


AdriVanHoudt:
@hueniverse you stated that no version is compatible with node 10 and I saw that node 10 was removed from Travis for v16.
We want to move to node 10 but are still on v16 so I thought I could look into this.
From what I can tell there are some deprecation warnings for Buffer usage in the tests which are easy to solve.
Other than that there is 1 test failing on node 10.
Test `807) transmission transmit() does not leak stream data when request timeouts before stream drains`
It times out due to it waiting for the stream `end` event to fire.
https://github.com/hapijs/hapi/blob/v16/test/transmit.js#L1944-L1986
In node 8 it does. In node 10 `close` is being fired but not `end`.
I don't know enough of streams to be sure why this is happening but according to the docs `The 'end' event will not be emitted unless the data is completely consumed.` which seems to me that node 10 is doing the correct thing? (https://nodejs.org/docs/latest-v10.x/api/stream.html#stream_event_end)

#### What was the result you received?
test `807) transmission transmit() does not leak stream data when request timeouts before stream drains` timing out
#### What did you expect?
All tests passing 
#### Context

* *node version*: 10
* *hapi version*: 16
* *os*: macOS
* *any other relevant information*:

Are you ok with an effort to support node 10 for v16? Are there issues for node 10 I might be missing?

AdriVanHoudt:
It seems that master has an updated version of the test
https://github.com/hapijs/hapi/blob/master/test/transmit.js#L1310-L1344
But unless I'm missing something that test is missing an `await team.work` to make sure the `end` event is being called?

AdriVanHoudt:
Sounds good 👌 
I think we need to start with #3900, which after resolving I can probably port to v16. Together with the other buffer stuff in the tests we can get the tests passing on node 10 at least.
Are the benchmarks you talk about just for hapi or also for individual modules?

(side note: we might be moving to 17 soon-ish, but if I can help people on LTS with this I will gladly try to help get at least test passing on node 10.)

WesTyler:
Given the impending demise of v16 in #3932 , can this be closed?

hueniverse:
Not yet. It might still happen for the commercial license option.

hueniverse:
Fixed in #3933

PR #3900, opened 2019-01-10
fix(transmit test): add missing await on team.work

AdriVanHoudt:
This break the test and shows the test doesn't actually work on node 10.
This points out something I bought up in https://github.com/hapijs/hapi/issues/3899
This test is also broken on v16 on node 10 for probably the same reason.


AdriVanHoudt:
As an fyi, I think the underlying principle that is being tested does work, but the test itself listen to the wrong event.

AdriVanHoudt:
Looks like this also breaks node 8. The same test on v16 on node 8 works (locally and CI)

Issue #3925, opened 2019-02-08
Update hapijs/wreck to 14.1.4 from 14.1.3


hueniverse:
Not a concern, just good housekeeping.

PR #3940, opened 2019-04-01
chore(deps): Upgrade wreck.

aorinevo:
- Resolves https://github.com/hapijs/hapi/issues/3925


hueniverse:
Thanks but I do these after integration testing and review.

PR #3946, opened 2019-04-26
Add support for Chrome Apps and WebExtensions

lidel:
Stream detection via `stream instanceof Stream` does not work correctly in browser contexts such as Chrome Apps or [WebExtensions in browsers exposing `chrome.sockets`](https://github.com/ipfs-shipyard/ipfs-companion/issues/664),  especially when different `stream` polyfils are used and mixed together.

This change replaces `instanceof Stream` checks with feature-detection of stream contract, enabling Hapi to be used in browserified form in Chrome Apps and other browser environments.

cc https://github.com/ipfs-shipyard/ipfs-companion/issues/716

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -98,7 +98,7 @@ exports = module.exports = internals.Response = class {
             this.variety = 'buffer';
             this._contentType = 'application/octet-stream';
         }
-        else if (source instanceof Stream) {
+        else if (Streams.isReadableStream(source)) {
             this.variety = 'stream';
             this._contentType = 'application/octet-stream';
         }
@@ -529,7 +529,7 @@ exports = module.exports = internals.Response = class {
 
         // Stream source
 
-        if (source instanceof Stream) {
+        if (Streams.isStream(source)) {
             if (typeof source._read !== 'function') {
                 throw Boom.badImplementation('Stream must have a readable interface');
             }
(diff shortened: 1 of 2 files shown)


hueniverse:
First, I am not sure I want to support this use case. Can you give me an example where doing so would make sense? What kind of app would you be building with it?

Second, experience tells me this is not the only thing stopping it from working properly. We've gone through this with joi and other modules that required a lot of changes to work properly.

lidel:
Real live example: js-ipfs in Brave browser (https://github.com/ipfs-shipyard/ipfs-companion/issues/716)

[js-ipfs](https://github.com/ipfs/js-ipfs) is using Hapi to expose HTTP2IPFS Gateway. [Brave](http://brave.com/) whitelisted  [our browser extension](https://github.com/ipfs/ipfs-companion/) to have access to `chrome.sockets` and we are able to start HTTP2IPFS Gateway from within WebExtension context, without installing any additional software.  

We tested it in practice and _it works as expected_ (https://github.com/ipfs-shipyard/ipfs-companion/issues/664#issuecomment-479535264). We use [chrome-net](https://github.com/feross/chrome-net) and [http-node](https://www.npmjs.com/package/http-node) to polyfill calls to `net` and `http` expected by Hapi. 

The only remaining issue with Hapi was the Stream detection, and change from this PR solved that (we use code form this PR).  

Stream-related entries from `test/response.js` pass, but I'm happy to add any additional tests you feel are necessary.

hueniverse:
Thanks for the background. I hope to get to it this week and let you know if/how I can support it.

lidel:
Anything I can do to help with evaluating this?

PR #3949, opened 2019-05-08
Destructed methods of Tooltip

Firanolfind:


hueniverse:
Did you even test this before opening the PR?!

Issue #3956, opened 2019-06-17
ETags behind gateway that compresses responses


kanongil:
I encountered an issue with the ETag validation for `if-none-match` requests that are behind an nginx gateway.

**The issue**

When a Hapi server sits behind an nginx reverse proxy, with compression enabled, responses that contain an `ETag` are returned weak. Eg. `etag: "hi"` is received as `etag: W/"hi"`. This means that later conditional requests will check against the weak version: `if-none-match: W/"hi"`, and thus _never matches_.

**Fixes** 

While this is probably an nginx issue, I put it here, as Hapi might want to workaround this behaviour. This can be done by matching weak etags against non-weak etags, and taking care to return the weak `ETag` in the response (otherwise the client would receive `304` response with what looks like a "new" etag, since nginx does not try to gzip a 304 response).

Alternatively, I guess api could add a "weak" option on routes, to force a weak response, making the check work.

See https://stackoverflow.com/a/55525109/248062 for some more background.

kanongil:
I should probably add that I remove the `Accept-Encoding` header in the proxied requests to force nginx to do the compression.

PR #3957, opened 2019-06-18
 Match etags against weak if-none-match test

kanongil:
This implements the weak test solution from #3956.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -220,7 +220,7 @@ exports = module.exports = internals.Response = class {
         if (request.method !== 'get' &&
             request.method !== 'head') {
 
-            return false;
+            return {};
         }
 
         // Strong verifier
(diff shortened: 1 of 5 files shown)


lib/response.js
```diff
@@ -231,45 +231,49 @@ exports = module.exports = internals.Response = class {
             const ifNoneMatch = request.headers['if-none-match'].split(/\s*,\s*/);
             for (const etag of ifNoneMatch) {
                 if (etag === options.etag) {
-                    return true;
+                    return { code: 304 };
                 }
 
                 if (options.vary) {
+                    if (etag === `W/${options.etag}`) {
```
hueniverse: Any reason why this is inside the vary?

hueniverse:
This was a bit too messy for me, but I took the important bits.

I was considering whether this should be behind an option but decided that almost no one will be able to understand how and when to use it. I had to go read the RFC again just to review this and understand the changes.

PR #3986, opened 2019-10-11
Fix server events request description

mgartner:
Since https://github.com/hapijs/hapi/commit/770cc7bad15122f796d938738b7c05b66d2f4b7f#diff-204116a0e3044ec4220e84420843efbcR53, there _are_ `'request'` events that are generated internally, but the current v16 docs say there are not.

I think this is the right place to update the v16 docs on hapi.dev, but lmk if that's not the case.


hueniverse:
That's actually a bug. The event needs to move to internal.

Issue #3993, opened 2019-11-06
Allow payload on GET requests


AlonMiz:
#### Support plan
* *which support plan is this issue covered by?* (e.g. Community, Core, Plus, or Enterprise): Community
* *is this issue currently blocking your project?* (yes/no): yes
* *is this issue affecting a production system?* (yes/no): no

#### Context
* *node version*: 10
* *module version*: 18.4
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi
* *any other relevant information*: 

#### What problem are you trying to solve?
Allowing GET routes to recieve body/payload. as of now, the payload is not being processed at all.
I agree it's not a common use case, but that case, we want to proxy requests which are not controlled by our end, and some of the requests are GET with a payload.
eg. Elasticsearch `_msearch` route allows GET requests with a payload.

```js
server.route({
    method: 'GET',
    path: '/',
    handler: (request, h) => {
        return request.payload; // payload will always be empty
    },
});
```

Looking at the library code, it seems that allowing such a case requires removal of several defenses mechanisms. a payload on a GET request is literally nulled, no matter if the payload exists or not.

#### Do you have a new or modified API suggestion to solve the problem?
Allowing GET routes to receive payload, requires only to remove the existing blocking of that capability

devinivy:
One possibility is to utilize an `onRequest` extension and `request.setMethod()` to rewrite the request to use a method that supports payloads.

hueniverse:
@AlonMiz the suggestion from @devinivy is the way to work around it. I am not going to support it. If you prefer to use express, that's your choice.

PR #3994, opened 2019-11-06
Allow GET routes requests to receive payload

AlonMiz:
https://github.com/hapijs/hapi/issues/3993

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -123,12 +123,7 @@ exports = module.exports = internals.Route = class {
 
         // Payload parsing
 
-        if (this.method === 'get') {
-            this.settings.payload = null;
-        }
-        else {
-            this.settings.payload.decoders = this._core.compression._decoders;        // Reference the shared object to keep up to date
-        }
+        this.settings.payload.decoders = this._core.compression._decoders;        // Reference the shared object to keep up to date
 
         this._assert(!this.settings.validate.payload || this.settings.payload.parse, 'Route payload must be set to \'parse\' when payload validation enabled');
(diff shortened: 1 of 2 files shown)


hueniverse:
You have made everyone's GET requests slower for this odd use case that no one has asked for in 7 years... ;-)

This is never going to be supported by default. I need to think if I want to support it behind a flag. Either way, it is low on my list and I'm currently very short on resources to add new items so realistically, this would be 3-4 months out even with a full PR because the time it takes to review, test, and align with other efforts.

dougwilson:
FWIW, @hueniverse , the HTTP Semantics draft was just updated very recently to continue to reflect the HTTP WG's dislike for support of a payload in a GET request, if you want to point folks there:

https://tools.ietf.org/id/draft-ietf-httpbis-semantics-06.txt

> 7.3.1.  GET
>
>    The GET method is specifically intended to reflect the quality of
   "sameness" identified by the request URI as if it were referenced as
   an ordinary hypertext link.  A client SHOULD NOT generate a body in a
   GET request.  A payload received in a GET request has no defined
   semantics, cannot alter the meaning or target of the request, and
   might lead some implementations to reject the request and close the
   connection because of its potential as a request smuggling attack
   (Section 11.2 of [Messaging]).

AlonMiz:
Thanks for the quick response.
I agree that this is a pretty bad practice, but the world is not perfect. we need this capability in order to supply a fully working product as other solutions still use that bad practice (such as elasticsearch `_msearch`).
We've started a migration to `Express` in order to support that case BTW

PR #3996, opened 2019-11-12
Decorate requests with symbols with apply=true

alexghr:
Hello,

 This PR fixes a bug where Symbol decorations would not get added to request instances when `{ apply: true }` option was used. The issue was caused by iterating over decorations object using `for...in` which only iterates over enumerable properties (symbols are not enumerable)

Since Symbols are expected property names this PR fixes the issue above by also iterating over known symbol properties using `Object.getOwnPropertySymbols`.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -103,7 +103,7 @@ exports = module.exports = internals.Core = class {
         this.Request = class extends Request { };
 
         this._debug();
-        this._decorations = { handler: {}, request: {}, server: {}, toolkit: {}, requestApply: null };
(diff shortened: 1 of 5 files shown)


kanongil:
I would probably refactor `requestApply`, along with the other `_decorations` entries, to use a `Map()`. This makes it explicit which properties are decorated, and should fix this issue.

alexghr:
Thanks for the suggestion @kanongil. It simplifies the code greatly! I'm not very familiar with the codebase but I've tried my best at refactoring the internal `_decorations` object to use a series of maps. 

The test still pass apart from two of them, although they were failing even before this change.

hueniverse:
It's rare I merge a PR... nicely done.

Issue #4000, opened 2019-11-14
auth scheme artifacts are dropped when mode is not 'try'


nlf:
#### Support plan

* *which support plan is this issue covered by?* (e.g. Community, Core, Plus, or Enterprise): Community
* *is this issue currently blocking your project?* (yes/no): Yes
* *is this issue affecting a production system?* (yes/no): No

#### Context

* *node version*: 12.x
* *module version with issue*: 18.x
* *last module version without issue*: n/a 
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): [hapi-rate-limit](https://github.com/wraithgar/hapi-rate-limit)
* *any other relevant information*:

#### What are you trying to achieve or the steps to reproduce?

We're attempting to use artifacts from a request to pass data to lifecycle hooks in order to facilitate several levels of rate limiting. The issue we're having currently is that the `artifacts` from an auth scheme are only attached to the request object after an authentication failure when the mode is set to `try`, while we need the artifacts regardless of the mode.

Relevant code segment appears to be this: https://github.com/hapijs/hapi/blob/master/lib/auth.js#L464-L472

#### What was the result you got?

Artifacts do not get attached to the request

#### What result did you expect?

Artifacts to be attached to the request

PR #4002, opened 2019-11-15
fix(auth): properly populate request.auth on failed auth

wraithgar:
Currently request.auth is only populated on failed auth if the mode is set to 'try'
This will make that object always populated, regardless of auth mode.


wraithgar:
The issue for this is here: https://github.com/hapijs/hapi/issues/4000

Issue #3919, opened 2019-02-02
Change emptyStatusCode to 204 by default


hueniverse:

marcuspoehls:
I like this change!

Does hapi also detect empty response payloads and sets 204 by default?

hueniverse:
That's what this setting is all about.

marcuspoehls:
Nice!

I didn't know that there is already an empty response detection in place. I digged through the code and now know it. Sweet!

PR #4024, opened 2020-01-14
fix(cors): emptyStatusCode not being taken into account for OPTIONS requests

mfeltscher:
With #3919 the default status code for empty responses changed from `200` to `204` while still giving the user the ability to override this behaviour by using the `emptyStatusCode` response setting. This setting however isn't being taken into account when using CORS and OPTIONS requests. This pull request fixes this glitch.

### DIFF
--- a/lib/cors.js
+++ b/lib/cors.js
@@ -141,6 +141,7 @@ internals.handler = function (request, h) {
     response._header('access-control-allow-methods', method);
     response._header('access-control-allow-headers', settings._headersString);
     response._header('access-control-max-age', settings.maxAge);
+    response.code(route.settings.response.emptyStatusCode);
 
     if (settings.credentials) {
         response._header('access-control-allow-credentials', 'true');
--- a/test/cors.js
+++ b/test/cors.js
@@ -35,6 +35,22 @@ describe('CORS', () => {
         server.route({ method: 'GET', path: '/', handler });
 
         const res = await server.inject({ method: 'OPTIONS', url: '/', headers: { origin: 'http://example.com/', 'access-control-request-method': 'GET' } });
+        expect(res.statusCode).to.equal(204);
+        expect(res.headers['access-control-allow-origin']).to.equal('http://example.com/');
+    });
+
+    it('returns OPTIONS response with correct status code for pre-defined emptyStatusCode', async () => {
+
+        const handler = function () {
+
+            throw Boom.badRequest();
+        };
+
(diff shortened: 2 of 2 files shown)


mfeltscher:
@hueniverse I just stumbled upon this issue when upgrading to v19 - could you please have a look?

kanongil:
Hmm, this might not be a safe thing to do: https://stackoverflow.com/a/58794243/248062.

mfeltscher:
@kanongil I can also change this to always return `200` if that's a better solution in your eyes.

kanongil:
Ah, this is a bit tricky actually.

You issue pertains to the `emptyStatusCode`, as set on a route with a cors config, which you would expect to also apply to the `OPTIONS` response. But it currently doesn't.

However, when setting the `emptyStatusCode` as a global route config when creating the server, it does apply to all CORS `OPTIONS` responses. This is now by default `204`.

Whether it should always return `200` will be up to @hueniverse, but there is definitely an issue here.

mfeltscher:
@kanongil How can I configure this as a global route config?

kanongil:
```js
new Server({
    routes: {
        response: {
            emptyStatusCode: 200
        }
    }
});
```

mfeltscher:
@kanongil Thanks, that works like a charm! I'll still leave this PR open for @hueniverse to have a look at.

devinivy:
It looks like kong also went through this.  In short, it sounds like the spec doesn't technically care but browser implementations do care or have cared in the past.  MDN docs used to recommend serving a 200, but now they recommend a 204.  Kong used to serve a 204, was patched to serve a 200, and today has an open issue to move back to 204.

https://github.com/Kong/kong/issues/4008
https://github.com/Kong/kong/pull/4029
https://github.com/Kong/kong/issues/6051
https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

I propose we stick with the 204 default, but create a separate option `route.cors.preflightStatusCode` to control it in hapi v21 rather than `route.response.emptyStatusCode`.

devinivy:
This has been hanging around for a while, and I don't think this PR is going to land as-is.  But the contribution is much appreciated, and helped move the conversation about this issue forward.  There are still some ideas and open questions above, so I am going to close this and move it to an issue for further discussion.

devinivy:
Here it is: https://github.com/hapijs/hapi/issues/4165

PR #4050, opened 2020-02-14
Allow @hapi/joi 16.x.x for v18

maximelkin:
Note: all dependencies of @hapi/hapi@18 requires 16.x.x


maximelkin:
Ok, I see this breaks too many things

PR #4060, opened 2020-02-22
removed the legacy reference that was not working

yuukiii:
Removed the joi-legacy dependency.


hueniverse:
It's a dev dep that is working fine for me.

yuukiii:
You may need to download a fresh branch to see the issue.  I forked and will get updates until this is fixed.

hueniverse:
Why do you care about dev deps?

Marsup:
I don't see any issue either.

Issue #4053, opened 2020-02-17
Post-mortem on a recent hapi logging adventure


cjihrig:
#### Support plan

* *which support plan is this issue covered by?* (e.g. Community, Core, Plus, or Enterprise): community
* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*: 12 and 13
* *module version*: 19.x
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi application
* *any other relevant information*: related to events and logging

#### What problem are you trying to solve?

I was recently working on a logging module (https://github.com/cjihrig/graylogi) and ended up diving into the events and logging deeper than I otherwise have recently. This issue is documenting some of the things that I found a bit confusing, as well as some logging features that would be useful to have from the framework:

- The `'request'` event is not emitted when a new request is received. Due to its name, and the presence of the similar `'response'` event, it took me a minute to realize that I needed the `onRequest` extension instead.
- The fact that `server.log()` triggers `'log'` events, but `request.log()` triggers `'request'` events.
- The fact that request errors also trigger `'request'` events. It's nice that there are a small number of events to handle, but the flip side is that individual event handlers end up more complex than they'd otherwise need to be.

#### Do you have a new or modified API suggestion to solve the problem?

From a logging perspective, the following things would be great to have:

- It would be nice if hapi either provided some logging friendly objects/methods as first class citizens, or `toJSON()` methods on more objects that are otherwise problematic to deal with in JSON (circular structures, `Map`s, etc.). I know this might be tricky, as people will always want different things in their logs.

- Built in support for ignoring events based on path or tags would also be great. For example, ignoring everything related to `/health`.

- Support for log levels, including the ability to set a default log level and a minimum log level threshold.

- A documented log format. That would enable things like logging to `stdout`/`stderr` from hapi, and piping them to another process for more in depth handling (reformatting, shipping, etc.). This is similar to pino's concept of transports.

Sorry if it seems like I'm requesting a lot 😄

devinivy:
Just chiming in here to offer some brief feedback and squeak out a few more details.

> The fact that server.log() triggers `'log'` events, but `request.log()` triggers `'request'` events.

You see this solely as a naming issue, is that right?

[...]

PR #4073, opened 2020-03-15
Add new ext onPostResponse

hueniverse:
This will eventually replace the `response` event (removed next major). Related to feedback from #4053.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -384,12 +384,20 @@ exports = module.exports = internals.Request = class {
         }
     }
 
-    async _invoke(event) {
+    async _invoke(event, options = {}) {
 
         for (const ext of event.nodes) {
             const realm = ext.realm;
             const bind = ext.bind || realm.settings.bind;
-            const response = await this._core.toolkit.execute(ext.func, this, { bind, realm, timeout: ext.timeout, name: event.type });
+            const response = await this._core.toolkit.execute(ext.func, this, { bind, realm, timeout: ext.timeout, name: event.type, ignoreResponse: options.ignoreResponse });
+
+            if (options.ignoreResponse) {
+                if (Boom.isBoom(response)) {
+                    this._log(['ext', 'error'], response);
+                }
+
+                continue;
+            }
 
             if (response === Toolkit.symbols.continue) {
                 continue;
@@ -494,7 +502,12 @@ exports = module.exports = internals.Request = class {
         }
 
         this.info.completed = Date.now();
+
         this._core.events.emit('response', this);
+
+        if (this._route._extensions.onPostResponse.nodes) {
+            this._invoke(this._route._extensions.onPostResponse, { ignoreResponse: true });
+        }
     }
 
     _setResponse(response) {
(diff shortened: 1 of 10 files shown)


lib/request.js
```diff
@@ -494,7 +502,12 @@ exports = module.exports = internals.Request = class {
         }
 
         this.info.completed = Date.now();
+
         this._core.events.emit('response', this);
+
+        if (this._route._extensions.onPostResponse.nodes) {
+            this._invoke(this._route._extensions.onPostResponse, { ignoreResponse: true });
+        }
```
kanongil: Why the conditional? It should work the same without.

lib/server.js
```diff
@@ -244,14 +246,28 @@ internals.Server = class {
 
     ext(events, method, options) {        // (event, method, options) -OR- (events)
```
kanongil: This doesn't handle `(event, options)`, which the docs imply is possible.

lib/request.js
```diff
@@ -384,12 +384,20 @@ exports = module.exports = internals.Request = class {
...
-            const response = await this._core.toolkit.execute(ext.func, this, { bind, realm, timeout: ext.timeout, name: event.type });
+            const response = await this._core.toolkit.execute(ext.func, this, { bind, realm, timeout: ext.timeout, name: event.type, ignoreResponse: options.ignoreResponse });
+
+            if (options.ignoreResponse) {
+                if (Boom.isBoom(response)) {
+                    this._log(['ext', 'error'], response);
+                }
+
+                continue;
+            }
```
kanongil: This logic needs to be executed in parallel for `onPostResponse` handlers. Otherwise the timing can get messed up.

This will make `onPostResponse` a good place to apply _tail_ logic, allowing errors to be logged and all.

lib/server.js
hueniverse: I'll clarify the docs.

lib/request.js
hueniverse: Should be faster avoiding the call + promise generation.

lib/request.js
hueniverse: I don't follow. Can you give an example where serial processing is going to be a problem here (and how is it different from current response event)?

kanongil: Say you add an `onPostResponse` handler to increment a value in your db, and a handler to log to stdout. The log handler would be delayed an indeterminate amount of time while the previous handler talks to the db. This will in turn mean that the log entry will appear later than intended, and multiple log entries can easily appear out of order.

Also, the current response event is essentially a parallel handler, since it is a fire and forget thing.

hueniverse: The `await` lets you enforce order but it is not required. There is nothing blocking you for implementing extension handlers that return right away and schedule work for a next tick. I'm inclined not to special case this to parallel execution because it makes reasoning about ext order harder. Keeping it the same means the same plugin order logic applies, etc.

It is probably worth documenting it so that people will write smarter handlers if they rely on getting quick notification of the event and then doing something.

kanongil: It makes sense that you want to keep the existing reasoning about how it runs, though I believe it can cause issues.

Maybe an extension isn't the right fit for this? Most of the `h` properties / methods doesn't make sense anyway.

hueniverse: It's not a perfect fit but it benefits for a lot of other similar facilities that I think it still works. We'll see how it goes.

Issue #4083, opened 2020-04-02
Default auth scope not compatible with route auth scope


hovmand:
#### Support plan

* *which support plan is this issue covered by?* (e.g. Community, Core, Plus, or Enterprise): Community
* *is this issue currently blocking your project?* (yes/no): Yes
* *is this issue affecting a production system?* (yes/no): No

#### Context

* *node version*: 10.15.3
* *module version with issue*: 18.4.1
* *last module version without issue*: 16.x
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): An auth scheme
* *any other relevant information*: -

#### What are you trying to achieve or the steps to reproduce?

I have a default auth setting and want to add a route with `options.auth.mode = optional`. This results in an error.

When `scope` is present in the default auth settings a route with auth options but without `options.auth.access.scope` will result in an error.

The `scope` of the default auth settings is stored in the expanded form, which is not compatible with how the routes are configured.

Example to provoke the error:
```js
'use strict';

const Hapi = require('@hapi/hapi');


const validate = async (request, username, password, h) => {

    return { isValid: true };
};

const main = async () => {

    const server = Hapi.server({ port: 4000 });

    await server.register(require('@hapi/basic'));

    server.auth.strategy('simple', 'basic', { validate });
    server.auth.default({
      strategy: 'simple',
      access: {
        scope: [ 'user' ]
      }
    });

    server.route({
        method: 'GET',
        path: '/',
        options: {
          auth: {
            mode: 'optional'
          }
        },
        handler: function (request, h) {

            return 'welcome';
        }
    });

    await server.start();

    return server;
};

main()
.then((server) => console.log(`Server listening on ${server.info.uri}`))
.catch((err) => {

    console.error(err);
    process.exit(1);
});
```

I used an online editor to create this example, and it didn't support v19 of hapi, but looking at the code the result should be the same.

#### What was the result you got?

```
TypeError: access.scope is not iterable
    at Object.internals.setupScope (/home/runner/LimeImpracticalStaff/node_modules/@hapi/hapi/lib/auth.js:424:32)
    at module.exports.internals.Auth._setupRoute (/home/runner/LimeImpracticalStaff/node_modules/@hapi/hapi/lib/auth.js:172:42)
    at new module.exports.internals.Route (/home/runner/LimeImpracticalStaff/node_modules/@hapi/hapi/lib/route.js:139:71)
[...]

PR #4089, opened 2020-04-08
[default-scope] Accept default scope as a valid route scope

jonathansamines:
**How to reproduce?**
When a default scope is specified by the default strategy (`server.auth.default()`) and a route is configured with auth options (but not `access.scope`), the default scope is not taking effect and instead throws an error as described by #4083.

**Root cause**
Default strategy's scopes are [parsed](https://github.com/hapijs/hapi/blob/master/lib/auth.js#L419), changing the original value from an array (access.scope = []) to an object (access.scope = { ... }). When a route does not specify it's own scopes, the default strategy scopes are used, but at that point they are no longer an array.

**Fix**
This PR fixes the issue by ignoring `access.scope` if it has already been setup in default strategy

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -418,8 +418,16 @@ exports = module.exports = internals.Auth = class {
 
 internals.setupScope = function (access) {
 
+    // No scopes
+
     if (!access.scope) {
         return false;
+    }
+
+    // Already setup
+
+    if (!Array.isArray(access.scope)) {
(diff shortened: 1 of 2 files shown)


devinivy (approved):

Nargonath (approved):

nlf (approved):

devinivy:
Resolves #4083

PR #4115, opened 2020-07-16
Replace joi with validate

hueniverse:
This moves hapi to use the `@hapi/validate` modules instead of `@hapi/joi` which is moving outside of the hapi org. Since joi was only used internally for API input validation, this is not a breaking change. However, this fixes a bug that was overlooked in v19 where rules are used with schemas and were still being compiled by the internal joi module instead of the registered validator.


hueniverse:
Please wait to merge until v19.2.0 is released.

PR #4117, opened 2020-07-21
Implement `isInjected` read-only request property to indicate request…

### LINKED ISSUE #4116: Improved injected request detection
#### Support plan

#### Context

* *node version*: any
* *module version*: >= 4.0.2
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi

We have a use case where we have designed a handler that needs to determine if a request has been injected by `@hapi/shot` to bypass trusted context validation.  Consider the following code:

```js
function skipValidationCheck(req) {
    // inspect request to determine if injected by server, in which case it is already trusted.
    const isRequestInjected = require('@hapi/shot').isInjection(request.raw.req);
    return isRequestInjected;
}
```

This approach works when a single version of `@hapi/shot` is installed locally because the underlying symbol on the request object prototype will match the one from the singleton method exported from `@hapi/shot`.  This however breaks down when multiple versions of `@hapi/shot` are installed due to exported symbols mismatching.

Proposal is to do one of the following:

- Modify [Request.prototype](https://github.com/hapijs/shot/blob/master/lib/request.js) to add `isInjected` property boolean to consistently identify injected requests.
- or, modify [Symbols](https://github.com/hapijs/shot/blob/master/lib/symbols.js) to export less unique value

I would probably lean to the first bullet as it maintains uniqueness check of symbol, but I don't fully understand the requirements of that.  Second bullet would be a simpler approach.

It might also help to modify `isInjected` documentation to clarify returns true only when versions of module match, unless that was not the intended behavior.

neenhouse:
Closes #4116 

Implements `isInjected` read-only request property to indicate request is injected by `server.inject()`

### DIFF
--- a/API.md
+++ b/API.md
@@ -4585,7 +4585,7 @@ Note that the `request.info` object is not meant to be modified.
 
 Access: read only.
 
-When `true`, indicates request was created via [`server.inject`](#server.inject()).  Defaults to `false`.
+When `true`, indicates request was created via [`server.inject`](#server.inject()). 
 
 #### <a name="request.logs" /> `request.logs`
 
--- a/API.md
+++ b/API.md
@@ -4585,7 +4585,7 @@ Note that the `request.info` object is not meant to be modified.
 
 Access: read only.
 
-When `true`, indicates request was created via [`server.inject`](#server.inject()). 
+`true` if the request was created via [`server.inject`](#server.inject()), and `false` otherwise.
 
 #### <a name="request.logs" /> `request.logs`
 
--- a/API.md
+++ b/API.md
@@ -4585,7 +4585,7 @@ Note that the `request.info` object is not meant to be modified.
 
 Access: read only.
 
-`true` if the request was created via [`server.inject`](#server.inject()), and `false` otherwise.
+`true` if the request was created via [`server.inject()`](#server.inject()), and `false` otherwise.
 
 #### <a name="request.logs" /> `request.logs`
 
--- a/API.md
+++ b/API.md
@@ -4581,6 +4581,12 @@ Request information:
 
 Note that the `request.info` object is not meant to be modified.
 
+#### <a name="request.isInjected" /> `request.isInjected`
+
+Access: read only.
+
+When `true`, indicates request was created via [`server.inject`](#server.inject()).  Defaults to `false`.
+
 #### <a name="request.logs" /> `request.logs`
 
 Access: read only.
--- a/lib/request.js
+++ b/lib/request.js
@@ -29,6 +29,7 @@ exports = module.exports = internals.Request = class {
         this._eventContext = { request: this };
         this._events = null;                                                                                // Assigned an emitter when request.events is accessed
         this._expectContinue = !!options.expectContinue;
+        this._isInjected = !!options.isInjected;
         this._isPayloadPending = !!(req.headers['content-length'] || req.headers['transfer-encoding']);     // Changes to false when incoming payload fully processed
         this._isReplied = false;                                                                            // true when response processing started
         this._route = this._core.router.specials.notFound.route;                                            // Used prior to routing (only settings are used, not the handler)
@@ -39,7 +40,6 @@ exports = module.exports = internals.Request = class {
 
         this.app = options.app ? Object.assign({}, options.app) : {};                                       // Place for application-specific state without conflicts with hapi, should not be used by plugins (shallow cloned)
         this.headers = req.headers;
-        this.isInjected = !!options.isInjected;
         this.jsonp = null;
         this.logs = [];
         this.method = req.method.toLowerCase();
@@ -100,6 +100,11 @@ exports = module.exports = internals.Request = class {
         }
 
         return this._events;
+    }
+
+    get isInjected() {
+
+        return this._isInjected;
     }
 
     get url() {
--- a/lib/request.js
+++ b/lib/request.js
@@ -15,7 +15,7 @@ const Transmit = require('./transmit');
 
 const internals = {
     events: Podium.validate(['finish', { name: 'peek', spread: true }, 'disconnect']),
-    reserved: ['server', 'url', 'query', 'path', 'method', 'mime', 'setUrl', 'setMethod', 'headers', 'id', 'app', 'plugins', 'route', 'auth', 'pre', 'preResponses', 'info', 'orig', 'params', 'paramsArray', 'payload', 'state', 'jsonp', 'response', 'raw', 'domain', 'log', 'logs', 'generateResponse']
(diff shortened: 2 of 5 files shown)


API.md
```diff
@@ -4581,6 +4581,12 @@ Request information:
 
 Note that the `request.info` object is not meant to be modified.
 
+#### <a name="request.isInjected" /> `request.isInjected`
+
+Access: read only.
+
+When `true`, indicates request was created via [`server.inject`](#server.inject()).  Defaults to `false`.
```
hueniverse, line 4588: Not really a "default"...

neenhouse: Removed `Defaults to false` statement.

test/core.js
```diff
@@ -1134,6 +1134,20 @@ describe('Core', () => {
             expect(res.result).to.be.true();
         });
 
+        it('sets `request.isInjected = true` for requests created via `server.inject`', async () => {
```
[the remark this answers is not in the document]
neenhouse: Implemented unit test for normal HTTP request created via `Wreck`.  I also clarified the preceding test statement where `request.auth.isInjected` is set to true when auth option is specified to clarify delineation between isInjected props.

neenhouse:
> I would also use this `request.isInjected` property here:
> 
> https://github.com/hapijs/hapi/blob/63352fabc1bdb59f1cf231c0180e20c105d5c458/lib/transmit.js#L97
> 
> .

Implemented.

neenhouse:
Can you guys take a look / merge https://github.com/hapijs/hapi/pull/4125 to address the 2 failing unit tests on master?  I'll rebase this one when it lands

lib/request.js
```diff
@@ -15,7 +15,7 @@ const Transmit = require('./transmit');
 
 const internals = {
     events: Podium.validate(['finish', { name: 'peek', spread: true }, 'disconnect']),
-    reserved: ['server', 'url', 'query', 'path', 'method', 'mime', 'setUrl', 'setMethod', 'headers', 'id', 'app', 'plugins', 'route', 'auth', 'pre', 'preResponses', 'info', 'orig', 'params', 'paramsArray', 'payload', 'state', 'jsonp', 'response', 'raw', 'domain', 'log', 'logs', 'generateResponse']
+    reserved: ['server', 'url', 'query', 'path', 'method', 'mime', 'setUrl', 'setMethod', 'headers', 'id', 'app', 'plugins', 'route', 'auth', 'pre', 'preResponses', 'info', 'isInjected', 'orig', 'params', 'paramsArray', 'payload', 'state', 'jsonp', 'response', 'raw', 'domain', 'log', 'logs', 'generateResponse']
```
devinivy, line 18: This could _arguably_ be seen as a breaking change.  Maybe this should go into v20 for good measure CC @cjihrig.

cjihrig: I don't think we should consider this semver major since the request object is owned by hapi. Any user code that is attaching things to the request object can use a symbol, `request.app`, etc.

cjihrig:
I cleared the Travis cache. Hopefully the CI will be green now.

API.md
```diff
@@ -4581,6 +4581,12 @@ Request information:
 
 Note that the `request.info` object is not meant to be modified.
 
+#### <a name="request.isInjected" /> `request.isInjected`
+
+Access: read only.
+
+When `true`, indicates request was created via [`server.inject`](#server.inject()).
```
cjihrig, line 4588:
```suggestion
`true` if the request was created via [`server.inject`](#server.inject()), and `false` otherwise.
```

API.md
```diff
@@ -4581,6 +4581,12 @@ Request information:
 
 Note that the `request.info` object is not meant to be modified.
 
+#### <a name="request.isInjected" /> `request.isInjected`
+
+Access: read only.
```
cjihrig, line 4586: Is this actually enforced at all? If not, we could use a getter like some other properties in `Request`.

neenhouse: Will implement in next change-set + unit test.

test/core.js
```diff
@@ -1134,6 +1134,57 @@ describe('Core', () => {
...
+
+            const server = Hapi.server();
+            server.route({ method: 'GET', path: '/', handler: (request) => {
+
+                try {
+                    request.isInjected = false;
+                }
+                catch (e) {
+                    // silence error to keep unit test output clean.  We expect above line to fail because isInjected() is getter type
+                }
```
cjihrig, line 1161: Can you use `expect(...).to.throw(...)` here.

API.md
```diff
@@ -4581,6 +4581,12 @@ Request information:
...
+
+`true` if the request was created via [`server.inject`](#server.inject()), and `false` otherwise.
```
cjihrig, line 4588: Sorry, I should have noticed this before. Based on the other uses of `server.inject()` in this file, I think this should be:

```suggestion
`true` if the request was created via [`server.inject()`](#server.inject()), and `false` otherwise.
```

lib/request.js
devinivy: I was thinking of `server.decorate('request', ...)` usage, but either way I see your point.

cjihrig: What you're saying makes sense. One thing we could do: automatically consider any additions to the `reserved` lists across the codebase as semver major.

Another longer term, but not mutually exclusive, idea: have `decorate()` update some new object (`request.decorations`?) on requests so that they won't ever conflict with hapi.

test/core.js
neenhouse: Implemented here https://github.com/hapijs/hapi/pull/4117/commits/0a99e50fec8a57c32df06b87a5ac84b84132fb8b, but then I rewrote it to https://github.com/hapijs/hapi/pull/4117/commits/674f3becf0eb14954381876fbc820f2ab47ec53a so that when unit test would fail, the failure message would be direct / obvious (In first commit, if unit test failed it would be that a 500 was received instead of a 200, but not immediately obvious to why that was)

test/core.js
```diff
@@ -1134,6 +1134,57 @@ describe('Core', () => {
...
+            expect(res.result).to.be.true();
+        });
+
+        it('`request.isInjected` access is read-only', async () => {
+
+            const server = Hapi.server();
+            let request = null;
+            server.route({ method: 'GET', path: '/', handler: (_request) => {
+
+                request = _request; // hoist request object to perform illegal assignment assertion
```
cjihrig, line 1157: This is a little convoluted. You can just move the assertion inline here:

```js
expect(() => {
  request.isInjected = false;
}).to.throw(...);
```

neenhouse: Sure, I reverted to the previous commit where I did inline assertion, I'm not sure if you saw my previous comment but my reasoning to do the above was to make failed assertion clearer.  Consider this assertion:

```
expect(illegalAssignment).to.throw('The wrong thing');
```

would produce:

Screen Shot 2020-07-25 at 8 20 19 PM

So it would fail the test but might not as clear as performing assertion outside route handler.  Happy to do it this way however as the code itself seems to be cleaner.

cjihrig:
Thanks for the contribution!

Issue #4122, opened 2020-07-23
Payload timeout must be shorter than socket timeout


kobelb:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*: v12.9.0
* *module version with issue*:  @hapi/hapi 19.2.0
* *last module version without issue*: N/A
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): standalone
* *any other relevant information*:

#### What are you trying to achieve or the steps to reproduce?

Configuring a route with `route.options.timeout.socket` smaller than `route.options.payload.timeout`

```js
 server.route({
        method: 'POST',
        path: '/',
        options: {
            payload: {
                timeout: 1000
            },
            timeout: {
                socket: 200
            }
        },
        handler: (request, h) => {
            return 'Hello World!';
        }
    });
```

#### What was the result you got?

Causes the following error to be throw:

```
AssertionError [ERR_ASSERTION]: Payload timeout must be shorter than socket timeout
```

#### What result did you expect?

I expected to be able to configure the `route.options.timeout.socket` to be smaller than `route.options.payload.timeout`. The payload timeout controls the total payload reception time; where-as `route.options.timeout.socket` is actually the "idle socket timeout", not the "total time socket is open timeout". For example, we should be able to configure the payload to be sent within 1 second, while ensuring that the socket isn't idle for 500ms during the entire duration of the request and response.

Issue #4114, opened 2020-07-14
Commercial License


hueniverse:
### Running open source versions of hapi prior to v20 are a technical and legal liability due to known bugs and security issues

* **If you are on version v20 or newer, you are all set, no need to read further.**
* If you are able to **upgrade to version v20 or newer**, that is your **best option** and it's completely free and open source.
* If you are unable or unwilling to upgrade to version v20, you should seriously consider a commercial license as **every published open source version of hapi prior to v20 has known bugs and security issues** only patched in the commercial distributions.

The new project leadership is fantastic and includes the best people for the job. They have my full trust and support and so should you. While I'm happy to sell you a commercial license, the best thing to do is upgrade to the latest free open source version which is v20 or newer.

## Background

The hapi project changed leadership from being maintained by Eran Hammer (me) to a new technical steering committee composed of veteran hapi contributors. This change was acknowledged with the release of hapi v20.0.0 (which is identical to v19.2.0).

All hapi users are encouraged to upgrade to v20 as soon as possible. It contains the best features, bug fixes, and security patches. It is the only supported open source licensed version.

## License Goals

A paid commercial license for versions v16, v17, v18, and v19 is being offered exclusively by Eran Hammer, the framework creator and previous lead maintainer.

The commercial license is offered to:
- allow companies to evaluation their options moving forward without a hard deadline.
- allow companies to stay on their current version of hapi (v16 - v19) due to lack to upgrade resources, product roadmap, or usage in a finished product or a product that is nearing its natural end-of-life.
- help cover on-going open source development cost of **joi** which was moved out of the hapi ecosystem.

The commercial license is **not** offered to:
- sponsor or support the hapi community - all funds will go exclusively to Eran Hammer, and will not help support future work on the hapi open source project. If you want to support hapi, this is not the right way to do it!
- provide access to any other services related to hapi - the commercial license is just that, a license granting access to use the hapi commercial distributions. It does not include any other professional services, priority support, custom development, etc.

What to expect from the commercial distributions:
- well maintained code that is free of **known** security issues and severe defects.
- tested and verified on the current LTS versions of node.

Companies paying for a commercial license can request non-critical bug fixes be ported to a covered version. While such requests are not guaranteed, they will be accomodated based on the compatibility of the fix with the legacy code base.

## Cost

The commercial license is offered on a monthly or annually basis.

For terms starting before January 1st, 2021:
Monthly: $500/month
Annually: $5000/year

[...]

PR #4123, opened 2020-07-23
Removing the route timeout validation

kobelb:
Resolves https://github.com/hapijs/hapi/issues/4122

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -83,11 +83,6 @@ exports = module.exports = internals.Route = class {
         const settings = internals.config([core.settings.routes, handlerDefaults, realm.settings, rulesConfig, config]);
         this.settings = Config.apply('routeConfig', settings, method, path);
 
-        // Validate timeouts
-
-        const socketTimeout = this.settings.timeout.socket === undefined ? 2 * 60 * 1000 : this.settings.timeout.socket;
-        this._assert(!this.settings.timeout.server || !socketTimeout || this.settings.timeout.server < socketTimeout, 'Server timeout must be shorter than socket timeout');
-        this._assert(!this.settings.payload.timeout || !socketTimeout || this.settings.payload.timeout < socketTimeout, 'Payload timeout must be shorter than socket timeout');
 
         // Route members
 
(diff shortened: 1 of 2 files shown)


lib/route.js
```diff
@@ -363,9 +358,7 @@ exports = module.exports = internals.Route = class {
             return;
         }
 
-        if (this.method[0] !== '_') {
```
kobelb, line 366: This conditional is no longer needed, and having it here reduced the code-coverage.

test/route.js
```diff
@@ -450,20 +450,20 @@ describe('Route', () => {
         expect(res.payload).to.contain('hapi');
     });
 
-    it('throws when server timeout is more then socket timeout', () => {
+    it('allows server timeout more then socket timeout', () => {
```
cjihrig, line 453: We should probably have similar tests for the payload timeout.

lib/route.js
cjihrig: Is this actually not needed anymore, or just not exercised by the tests anymore?

kobelb: I haven't been able to find a code-path that relies on this branch, but I have to admit I'm not fluent in Hapi's codebase.

devinivy: The method used for the internal not found, bad request, and cors preflight routes is `_special`.  I think we'll want to keep this around and find a way to exercise it with one of those routes.

test/route.js
kobelb: Done!

lib/route.js
kobelb: You all were right. I found a way to exercise this code-path with https://github.com/hapijs/hapi/pull/4123/commits/3dc68cc90f713e35e72dcb23a84665c3a2778ca7

kobelb:
Assuming others approve this PR, which versions will include this fix? Is it possible to have it included in v17?

cjihrig:
If this lands, it will only be for v20. If you need support for an older version, you'll have to check on a commercial license - https://github.com/hapijs/hapi/issues/4114.

cjihrig:
Thanks for the contribution!

PR #4127, opened 2020-07-29
Update transmit.js

jonaswindey:
This prevents the message "Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client"


Marsup:
Don't you actually *want* to know that? This can be a creepy bug if you're never aware of it.

kanongil:
As @Marsup points out, this might be the right fix for whatever issue you are having. At the very least, this PR needs a failing test case to maintain 100% coverage. This would also provide more insight into the issue.

PR #4141, opened 2020-08-11
docs: ✏️ improve docs for server.rules()

damusix:
server.rules() was missing a good example and explanation was a bit obscure


API.md
```diff
@@ -2366,8 +2366,65 @@ Defines a route rules processor for converting route rules object into route con
...
+
+    const options = {};
+
+    if (rules.auth) {
+        options.auth = {
+            strategy: rules.auth,
+            validate: {
+                entity: 'user'
+            }
+        }
```
devinivy, line 2393:
```suggestion
        };
```

API.md
```diff
@@ -2366,8 +2366,65 @@ Defines a route rules processor for converting route rules object into route con
...
+};
+
+const realmRules = {
+    validate: { schema: validateSchema }
+};
+
+server.rules(processor, realmRules);
+
+server.route({
+    ...,
```
devinivy, line 2417: It would be great to replace this placeholder with some `method` and `path` to make this example more complete.

API.md
```diff
@@ -2366,8 +2366,65 @@ Defines a route rules processor for converting route rules object into route con
...
-processor if they overlap. The route `config` overrides the rules config if the overlap.
+realm. This means the processor defined by the plugin overrides the config generated by the root
+processor if they overlap. `route.config` overrides `route.rules` if the resulting processed config overlaps.
+
+```js
+const validateSchema = {
+  auth: Joi.string(),
+  myCustomPre: Joi.array().min(2).items(Joi.string()),
+  payload: Joi.object()
+}
```
devinivy, line 2377:
```suggestion
};
```

API.md
```diff
@@ -2366,8 +2366,65 @@ Defines a route rules processor for converting route rules object into route con
...
+    }
+
+    return options;
+};
+
+const realmRules = {
+    validate: { schema: validateSchema }
+};
+
+server.rules(processor, realmRules);
```
devinivy, line 2414: The goal here is just to make things more concise and avoid the "realms" reference for those unfamiliar with that hapi concept.

```suggestion
server.rules(processor, {
    validate: { schema: validateSchema }
});
```

API.md
devinivy:
```suggestion
    method: 'GET',
    path: '/',
```

API.md
```diff
@@ -2366,8 +2366,65 @@ Defines a route rules processor for converting route rules object into route con
...
+
+    if (rules.myCustomPre) {
+
+        options.pre = [
+            myPreHelper(...rules.myCustomPre)
+        ];
+    }
+
+    if (rules.payload) {
+        options.validate.payload = Joi.object(rules.payload);
```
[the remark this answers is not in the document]
devinivy:
```suggestion
        options.validate = { payload: Joi.object(rules.payload) };
```

API.md
damusix: I will update this when I get a chance some time this week. Thank you for the feedback!

damusix:
@devinivy changes have been made!

API.md
```diff
@@ -2366,8 +2366,65 @@ Defines a route rules processor for converting route rules object into route con
...
+  auth: Joi.string(),
+  myCustomPre: Joi.array().min(2).items(Joi.string()),
+  payload: Joi.object()
+}
+
+const processor = (rules, info) => {
+
+    if (!rules) {
+        return null;
+    }
```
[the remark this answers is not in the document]
kanongil: This is redundant, as it should never trigger for the example. Otherwise hapi has botched the validateSchema validation of the rules object.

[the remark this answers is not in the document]
damusix: @kanongil perhaps this entire block could be wrapped in an if block https://github.com/hapijs/hapi/blob/master/lib/route.js#L67

```js
...

const rules = route.rules || config.rules;
let rulesConfig = {};

if (rules) {
    const rulesConfig = internals.rules(rules, { method, path, vhost }, server);
}
...
```

devinivy (approved):
There's chatter in slack about `rules()` and there may be some followup to do here, but for now this is great and is a very useful contribution.  Thanks!

PR #4150, opened 2020-08-25
bump minimum shot dependency version

cjihrig:


devinivy:
I see there's a failure.  Darn— on https://github.com/hapijs/shot/pull/129 I almost specifically requested that someone try linking it to hapi and giving it a test run.  It might be the case that it will work once #4095 lands.

cjihrig:
So it looks like this fixes the test:

```diff
diff --git a/test/payload.js b/test/payload.js
index b776a44f..0379bdfa 100755
--- a/test/payload.js
+++ b/test/payload.js
@@ -71,10 +71,10 @@ describe('Payload', () => {
 
         const responded = server.ext('onPostResponse');
 
-        server.inject({ method: 'POST', url: '/', payload: 'test', simulate: { close: true, end: false } });
+        server.inject({ method: 'POST', url: '/', payload: 'test', simulate: { close: true } });
         const request = await responded;
         expect(request._isReplied).to.equal(true);
-        expect(request.response.output.statusCode).to.equal(500);
+        expect(request.response.output.statusCode).to.equal(400);
     });
 
     it('handles aborted request', { retry: true }, async () => {
```

The issue is that after https://github.com/hapijs/shot/pull/129, we no longer force a `'close'` event to be emitted, and instead rely on Node to do it for us (which is why we need `'end'` to be emitted.

The question is whether we want to update the test and move on with life, or fix/revert the behavior change in `shot`. The only issue with just updating the test here is that the `shot` release was a patch release and this is a subtle behavioral change.

cc: @kanongil

Issue #4072, opened 2020-03-13
Response Event For Aborted Requests Has Status Code Of 200


crumkev:
#### Support plan

* *which support plan is this issue covered by?* (e.g. Community, Core, Plus, or Enterprise): Community
* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): yes

#### Context

* *node version*: 12.16.0
* *module version with issue*: 18.4.1, 19.1.1
* *last module version without issue*: unsure
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi application
* *any other relevant information*: similar to https://github.com/hapijs/hapi/issues/3878

#### What are you trying to achieve or the steps to reproduce?

I have a hapi server with a route that performs a potentially long-running calculation before responding.  Sometimes clients timeout or abort their request before the server has responded.  When this occurs, the server logs the response status code as 200 rather than 499.  I don't believe this is an issue with the logger, because both good/good-console and hapi-pino have the same behavior.  

```js
const Hapi = require('@hapi/hapi');

const server = Hapi.server({
  host: 'localhost',
  port: 8080,
});

server.route({
  method: 'GET',
  path: '/abort-test',
  handler: async function() {
    return new Promise(function (resolve) {
      setTimeout(function () {
        resolve('OK');
      }, 4000);
    });
  },
})

server.events.on('request', (request, event) => {
  console.log(`Event: ${JSON.stringify(event)}`);
});

server.events.on('response', (request) => {
  console.log(`Response Status Code: ${request.response.statusCode}`);
});

server.start();
console.log(`Server running at: ${server.info.uri}`);
```

```shell
curl --max-time 1 http://localhost:8080/abort-test
```

#### What was the result you got?

```shell
Server running at: http://localhost:8080
Event: {"request":"<SHORTENED>","timestamp":1584125752386,"tags":["request","error","abort"],"channel":"internal"}
Response Status Code: 200
```

#### What result did you expect?

Response status code to be 499 (not 200).

dnechay:
Hi!
I've faced the same issue in `19.2.0`, `20.0.1` and `20.1.2`.
Also it looks like `res.end()` isn't called in this case. Could anyone please help with that?

Nargonath:
@dnechay please file a new issue for your problem.

PR #4161, opened 2020-09-23
fix: check if _isPayloadSupported is set

jaulz:
I noticed that in certain cases the server crashes because it tries to access `response._isPayloadSupported` but it is not a function. Unfortunately, I wasn't really able to reproduce the issue and find the root cause but I checked the request object immediately before it crashes and the `response` property looks like this:
```
  response: Error: Request aborted
      at IncomingMessage.internals.event (/Users/Julian/Documents/dev/projects/myapprovals/myapprovals-server/node_modules/@hapi/hapi/lib/request.js:717:30)
      at IncomingMessage.emit (events.js:314:20)
      at IncomingMessage.EventEmitter.emit (domain.js:486:12)
      at abortIncoming (_http_server.js:535:9)
      at socketOnEnd (_http_server.js:551:5)
      at TLSSocket.emit (events.js:326:22)
      at TLSSocket.EventEmitter.emit (domain.js:486:12)
      at endReadableNT (_stream_readable.js:1244:12)
      at processTicksAndRejections (internal/process/task_queues.js:80:21) {
    data: null,
    isBoom: true,
    isServer: false,
    output: { statusCode: 499, payload: [Object], headers: {} }
  },
```

So it might be related to aborted requests.

### DIFF
--- a/lib/headers.js
+++ b/lib/headers.js
@@ -54,7 +54,8 @@ exports.content = async function (request) {
             response._header('content-length', response._payload.size(), { override: false });
         }
 
-        if (!response._isPayloadSupported()) {
+        if (response._isPayloadSupported &&
+            !response._isPayloadSupported()) {
             response._close();                              // Close unused file streams
             response._payload = new internals.Empty();      // Set empty stream
         }
--- a/lib/headers.js
+++ b/lib/headers.js
@@ -35,7 +35,7 @@ exports.cache = function (request) {
 exports.content = async function (request) {
 
     const response = request.response;
-    if (response._isPayloadSupported() ||
+    if ((response._isPayloadSupported && response._isPayloadSupported()) ||
         request.method === 'head') {
 
         await response._marshal();


kanongil:
Right, this can never happen for response objects. However, `request.response` can also be a `Boom` error in some cases, so this might be what is happening.

I don't think this fix is an appropriate way to handle this, as it just tries to fix a symptom of the underlying problem.

What I think is happening, is that the response is changed to an error during the `marshal()` cycle, which can happen if the request is aborted at the "right" time here: https://github.com/hapijs/hapi/blob/147b6fd14f7e03f9da1748ba09f191f154b4e7e2/lib/request.js#L717

This can cause the marshal to throw a 'system' error, when it tries to call `response. _isPayloadSupported()`, which will make hapi crash.

Looking at the code, this bug was introduced in https://github.com/hapijs/hapi/commit/e67e33fd181fbb8cc4ff4923d1e544e4e1ec22a8, as a fix for #4072.

jaulz:
@kanongil great analysis, thanks! Any idea how that could be properly fixed?

kanongil:
Yes, I have created a proper fix in #4162.

devinivy:
@jaulz thanks for bringing this up and helping us understand the issue through a proposed fix!  Will be moving to #4162.

PR #4162, opened 2020-09-23
Use response instead of request when marshalling

kanongil:
This is a proper fix for the issue in #4161. Also includes a failing test.

### DIFF
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -278,7 +278,7 @@ internals.end = function (env, event, err) {
     env.team = null;
 
     if (request.raw.res.finished) {
-        if (event !== 'aborted') {
+        if (!event) {
             request.info.responded = Date.now();
         }
 
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -24,21 +24,21 @@ exports.send = async function (request) {
             return;
         }
 
-        await internals.marshal(request);
+        await internals.marshal(response);
         await internals.transmit(response);
     }
     catch (err) {
         Bounce.rethrow(err, 'system');
-        request._setResponse(err);
+
         return internals.fail(request, err);
     }
 };
 
 
-internals.marshal = async function (request) {
+internals.marshal = async function (response) {
 
-    for (const func of request._route._marshalCycle) {
-        await func(request);
+    for (const func of response.request._route._marshalCycle) {
+        await func(response);
     }
 };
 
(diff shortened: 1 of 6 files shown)


lib/transmit.js
```diff
@@ -24,32 +24,32 @@ exports.send = async function (request) {
             return;
         }
 
-        await internals.marshal(request);
+        await internals.marshal(response);
         await internals.transmit(response);
     }
     catch (err) {
         Bounce.rethrow(err, 'system');
-        request._setResponse(err);
```
devinivy, line 32: The one thing I could use some confirmation on is whether it's alright that up on line 23 we don't have a comparable change to this one.

I suppose now we are calling `_setResponse()` (inside `fail()`) in the case a boom error (on line 23), where in the past we were not.  In theory my guess is that this is okay because `_setResponse()` will only close the response in the case of non-boom and completed responses.  Does that sound right?  I would feel more confident if `_setResponse()` were explicitly idempotent, but I can live with this as long as someone else agrees that this is alright.

devinivy (approved):
Left one open question/comment that I wouldn't mind another set of eyes on, but this looks great.  Thanks for the speedy fix.

cjihrig (approved):

lib/transmit.js
kanongil: The previous logic was hacky. There might be a slight functional change in that if the `Boom` error has a `_close` property, it could try to call it (depending on whether `this.info.completed` can be set for such a response). 

https://github.com/hapijs/hapi/blob/147b6fd14f7e03f9da1748ba09f191f154b4e7e2/lib/request.js#L528-L534

I can revert that part, as it is not essential to the fix.

jaulz:
Wow, you are great! Thanks a lot! 😊

test/transmit.js
```diff
@@ -531,6 +532,42 @@ describe('transmission', () => {
...
+            await server.start();
+
+            const log = server.events.once('response');
+            const client = Net.connect(server.info.port, () => {
+
+                client.write('GET / HTTP/1.1\r\n\r\n');
+            });
+
+            const response = await log;
+            expect(response).to.exist();
```
nlf, line 569: can we add an assertion or two here to make sure that the response event is the one we expect?

nlf (changes requested):
this change makes sense to me, i think the test leaves a little bit to be desired though. it's worth asserting that the response event is the client aborted event we expect it to be.

once that's done this is a :+1: from me

devinivy (approved):

nlf (approved):
looks great!

PR #4173, opened 2020-10-21
Propagate "peek" and "finish" errors as error response

### LINKED ISSUE #4172: Unable to cause an error during payload parsing
#### Support plan

* *is this issue currently blocking your project?* (yes/no): yes

#### Context

* *node version*: v14.14.0
* *module version with issue*: 19.2.0
* *last module version without issue*:  19.2.0
* *any other relevant information*: n.a.

I try to implement a plugin that validates that the `Digest` [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Digest) matches the payload.

I want to output a nice error message if the payload does not match the header

```js
server.ext('onRequest', (request, h) => {

    const { algorithm, digest } = getDigest(request.headers)

    const hash = Crypto.createHash(algorithm);
    request.events.on('peek', (chunk) => {

        hash.update(chunk);
    });

    request.events.once('finish', () => {

        if (digest !== hash.digest('base64')) throw new Boom.badData('nice error message')
    });

    return h.continue;
});
```

The server crashes due to an uncaught exception.

An http error response should be returned.

matthieusieben:
Fixes #4172

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -770,12 +770,14 @@ internals.Response.Peek = class extends Stream.Transform {
 
     async _transform(chunk, encoding, callback) {
 
-        try {
-            await this._podium.emit('peek', [chunk, encoding]);
-        }
-        catch (err) {
-            callback(err);
-            return;
+        if (this._podium.hasListeners('peek')) {
+            try {
+                await this._podium.emit('peek', [chunk, encoding]);
+            }
+            catch (err) {
+                callback(err);
+                return;
+            }
         }
 
         this.push(chunk, encoding);
@@ -784,12 +786,14 @@ internals.Response.Peek = class extends Stream.Transform {
 
     async _final(callback) {
 
-        try {
-            await this._podium.emit('finish');
-        }
-        catch (err) {
-            callback(err);
-            return;
+        if (this._podium.hasListeners('finish')) {
+            try {
+                await this._podium.emit('finish');
+            }
+            catch (err) {
+                callback(err);
+                return;
+            }
         }
 
         callback();
--- a/lib/response.js
+++ b/lib/response.js
@@ -774,7 +774,8 @@ internals.Response.Peek = class extends Stream.Transform {
             await this._podium.emit('peek', [chunk, encoding]);
 
             callback(null, chunk);
-        } catch (err) {
+        }
+        catch (err) {
             callback(err);
         }
     }
@@ -785,7 +786,8 @@ internals.Response.Peek = class extends Stream.Transform {
             await this._podium.emit('finish');
 
             callback(null);
-        } catch (err) {
+        }
+        catch (err) {
             callback(err);
         }
     }
--- a/lib/response.js
+++ b/lib/response.js
@@ -772,23 +772,25 @@ internals.Response.Peek = class extends Stream.Transform {
 
         try {
             await this._podium.emit('peek', [chunk, encoding]);
-
-            callback(null, chunk);
         }
         catch (err) {
             callback(err);
+            return;
         }
+
+        callback(null, chunk);
     }
 
     async _final(callback) {
 
         try {
             await this._podium.emit('finish');
-
-            callback(null);
         }
         catch (err) {
             callback(err);
+            return;
         }
+
+        callback();
     }
 };
--- a/lib/response.js
+++ b/lib/response.js
@@ -766,13 +766,27 @@ internals.Response.Peek = class extends Stream.Transform {
         super();
 
         this._podium = podium;
-        this.on('finish', () => podium.emit('finish'));
     }
 
-    _transform(chunk, encoding, callback) {
+    async _transform(chunk, encoding, callback) {
 
-        this._podium.emit('peek', [chunk, encoding]);
-        this.push(chunk, encoding);
-        callback();
+        try {
+            await this._podium.emit('peek', [chunk, encoding]);
+
+            callback(null, chunk);
+        } catch (err) {
+            callback(err);
+        }
+    }
+
+    async _flush(callback) {
+
+        try {
+            await this._podium.emit('finish');
+
+            callback(null);
+        } catch (err) {
+            callback(err);
+        }
     }
 };
--- a/lib/response.js
+++ b/lib/response.js
@@ -778,7 +778,8 @@ internals.Response.Peek = class extends Stream.Transform {
             return;
         }
 
-        callback(null, chunk);
+        this.push(chunk, encoding);
+        callback();
     }
 
     async _final(callback) {
--- a/lib/response.js
+++ b/lib/response.js
@@ -780,7 +780,7 @@ internals.Response.Peek = class extends Stream.Transform {
         }
     }
 
-    async _flush(callback) {
+    async _final(callback) {
 
         try {
             await this._podium.emit('finish');
--- a/test/payload.js
+++ b/test/payload.js
@@ -5,6 +5,7 @@ const Http = require('http');
 const Path = require('path');
 const Zlib = require('zlib');
 
+const Boom = require('@hapi/boom');
 const Code = require('@hapi/code');
 const Hapi = require('..');
 const Hoek = require('@hapi/hoek');
@@ -288,6 +289,48 @@ describe('Payload', () => {
         const payload = '0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789';
         await server.inject({ method: 'POST', url: '/', payload });
         expect(peeked).to.be.true();
+    });
+
+    it('propagates peek errors', async () => {
+
+        const ext = (request, h) => {
+
+            request.events.once('peek', () => {
+
+                throw Boom.badRequest();
+            });
+
+            return h.continue;
+        };
+
+        const server = Hapi.server();
+        server.ext('onRequest', ext);
+        server.route({ method: 'POST', path: '/', options: { handler: () => null, payload: { parse: false } } });
+
+        const payload = '0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789';
+        const res = await server.inject({ method: 'POST', url: '/', payload });
+        expect(res.statusCode).to.equal(400);
+    });
+
+    it('propagates finish errors', async () => {
+
+        const ext = (request, h) => {
+
+            request.events.once('finish', () => {
+
+                throw Boom.badRequest();
+            });
+
+            return h.continue;
+        };
+
+        const server = Hapi.server();
+        server.ext('onRequest', ext);
+        server.route({ method: 'POST', path: '/', options: { handler: () => null, payload: { parse: false } } });
+
+        const payload = '0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789';
+        const res = await server.inject({ method: 'POST', url: '/', payload });
+        expect(res.statusCode).to.equal(400);
     });
 
     it('handles gzipped payload', async () => {


lib/response.js
```diff
@@ -766,13 +766,29 @@ internals.Response.Peek = class extends Stream.Transform {
...
+            await this._podium.emit('peek', [chunk, encoding]);
+
+            callback(null, chunk);
+        }
+        catch (err) {
+            callback(err);
+        }
+    }
+
+    async _flush(callback) {
```
kanongil, line 783: You should not abuse the `_flush()` to handler to handle a `finish` event. If you need to signal a stream error, call `this.destroy(err)`.

lib/response.js
```diff
@@ -766,13 +766,29 @@ internals.Response.Peek = class extends Stream.Transform {
...
-    _transform(chunk, encoding, callback) {
+    async _transform(chunk, encoding, callback) {
 
-        this._podium.emit('peek', [chunk, encoding]);
-        this.push(chunk, encoding);
-        callback();
+        try {
+            await this._podium.emit('peek', [chunk, encoding]);
+
+            callback(null, chunk);
```
kanongil, line 776: Missing the encoding argument.

kanongil: This callback() call should be outside the try-catch. Otherwise we risk calling it twice (if it somehow throws).

kanongil (changes requested):
Looks good. Just needs a few tweaks.

lib/response.js
matthieusieben: If what you mean is doing the following, then it won't work because the the error will be thrown asynchronously, and even possibly after the stream was already destroyed.

```js
    constructor(podium) {

        super();

        this._podium = podium;
        this.on('finish', async () => {

            try {
                await this._podium.emit('finish');
            } catch (err) {
                this.destroy(err);
            }
        });
    }
```

Reading node's stream doc I realize that `_final` is a better place to perform this.

However I am not sure I follow you regarding `this.destroy(err)`. [the doc](https://nodejs.org/api/stream.html#stream_writable_final_callback) seem to specify that `callback` should always be called, with an error if need be.

matthieusieben: Reading node's stream doc I realize that `_final` is probably better place to perform this.

lib/response.js
```diff
@@ -766,13 +766,31 @@ internals.Response.Peek = class extends Stream.Transform {
...
+
+        try {
+            await this._podium.emit('peek', [chunk, encoding]);
+        }
+        catch (err) {
+            callback(err);
+            return;
+        }
+
+        callback(null, chunk);
```
matthieusieben, line 781: [callback](https://nodejs.org/api/stream.html#stream_transform_transform_chunk_encoding_callback) does not seem to take a third argument (for `encoding`).

Event Node's [PassThrough](https://github.com/nodejs/node/blob/9c62e0e384e2ea5a45a4011f31c97be60f2a0737/lib/internal/streams/passthrough.js#L45-L47) just calls `callback(null, data)`

lib/response.js
```diff
@@ -766,13 +766,31 @@ internals.Response.Peek = class extends Stream.Transform {
...
+    }
+
+    async _final(callback) {
+
+        try {
+            await this._podium.emit('finish');
+        }
+        catch (err) {
+            callback(err);
```
matthieusieben, line 790: no sure why you'd want to call `this.destroy(err)` directly. It [seems](https://github.com/nodejs/node/blob/9c62e0e384e2ea5a45a4011f31c97be60f2a0737/lib/internal/streams/writable.js#L651-L657) that that's what happening under the hood anyway (respecting the `autoDestroy` option).

lib/response.js
matthieusieben: Do you want me to revert to the following ?

```js
        this.push(chunk, encoding);
        callback();
```

lib/response.js
matthieusieben: [callback](https://nodejs.org/api/stream.html#stream_transform_transform_chunk_encoding_callback) does not seem to take a third argument (for `encoding`).

Event Node's [PassThrough](https://github.com/nodejs/node/blob/9c62e0e384e2ea5a45a4011f31c97be60f2a0737/lib/internal/streams/passthrough.js#L45-L47) just calls `callback(null, chunk)`

Do you want me to revert to the following anyway ?

```js
        this.push(chunk, encoding);
        callback();
```

lib/response.js
matthieusieben: no sure why you'd want to call `this.destroy(err)` directly.
It is my understanding that `this.destroy(err)` should only be performed if the `autoDestroy` option is set, which is what `callback(err)` is [actually doing](https://github.com/nodejs/node/blob/9c62e0e384e2ea5a45a4011f31c97be60f2a0737/lib/internal/streams/writable.js#L651-L657).

lib/response.js
matthieusieben: Well, i just did revert to the way it was. Let's not change what's working, am i right ?!

matthieusieben:
I don't think the failing tests are related to the changes of this PR.

lib/response.js
kanongil: Great

kanongil (approved):
LGTM.

lib/response.js
kanongil: Placing it in `_final()` seems better.

kanongil:
CI errors are due to https://github.com/hapijs/lab/pull/997.

matthieusieben:
Any idea on when this could land ?

matthieusieben:
As workaround, one can currently do:

```js
request.raw.req.emit('error', unauthorized('Corrupted payload'))
```

as in:

```js
server.ext('onRequest', (request, h) => {
  const hash = createHash('sha512')

  request.events.on('peek', (chunk, encoding) => {
    hash.update(chunk, encoding)
  })

  request.events.on('finish', () => {
    const payloadDigest = hash.digest('base64')
    if (payloadDigest !== getDigestHeader(request, 'sha512')) {
      /**
       * @hack Until we can throw from the 'finish' listener
       * @see {@link https://github.com/hapijs/hapi/pull/4173}
       */
      request.raw.req.emit('error', Boom.unauthorized('Corrupted payload'))
      // throw unauthorized('Corrupted payload')
    }
  })  
});

```

devinivy:
This is nice, I think I do like the behavior.  Thanks for the contribution.  Two questions here:

1. I see there's only one class shared by both request and response peeking.  Perhaps we should add tests for response peeking to demonstrate that we understand the effect that this change has there.  Seems like it could be more complex than request peeking.
2. This is a pretty hot code path (when there are event handlers), and I don't have a sense for what kind of effect this might have on performance.  Does anyone have interest in doing some benchmarking so that we know what to expect if we land this?  I foresee many microticks with those `await`s and I'm curious just how much they might add-up in the middle of request and response streaming.

matthieusieben:
I started writing some tests but encountered the following issue with `server.inject()`: https://github.com/hapijs/hapi/issues/4182

kanongil (commented):
Given the performance considerations that @devinivy brought up, I am no longer certain that this should be patched. I might defer to my original stance, which was that _it is a fatal programming error, if an event handler throws_.

Looking at podium itself, and the usage across the hapijs org, it is apparent that handling thrown errors is mostly a non-consideration. The two other modules that use it, catbox and beam, both call the emitter without handling the returned promise, resulting in an `unhandledRejection` emit on `process`.

For hapi itself, the `emit()` promise is only handled for the server `start` and `stop` events. Every other usage is presently unhandled. I would in fact consider any await of the returned promise to be a bug, since the application flow should not delay to handle any fallout of an event. `emit()` is by its nature a fire and forget thing.

All-in-all, the conclusion seems to be that hapi should stop awaiting `start` and `stop` events (breaking change), and podium should have an optimization pass to not return a promise (breaking change), and that this PR is unfortunately not appropriate.

matthieusieben:
It is kinda sad that there is no clean way to fix #4172. (Ie. Preventing any processing of the request payload) As for the performance issues, you know that stream heavily rely on micro tasks internally, right?

matthieusieben:
Also, what is the point of having a podium at all (instead of node's event emitter) if not for being able to await listeners execution?

kanongil:
> What is the point of having a podium at all (instead of node's event emitter) if not for being able to await listeners execution?

Podium contains channels and tags, that allow filtering and more efficient processing. The registration step prevents emitting mistyped event names.  It also ensures that all emitters are called on an `emit()`, even if one throws (though it is up to the application to handle such a throw). The promise returned from `emit()` is not even part of the documented API.

> For example, being able to execute all async setup tasks during start is so elegant and efficient (all tasks run in parallel).

It is still possible to stall a start() / stop(), you just need to use the exts `onPreStart` / `onPostStop`, which will have slightly different timing, but are more powerful, since you can control the order they are run in.

FYI, your contribution has helped us gain insight into a part that is not clearly handled. A clarification can go either way, and in this case it seems more consistent to do something other than you proposed.

kanongil:
FYI, if podium is changed to not emit async, you will likely be able to solve your #4172 issue with:

```js
    request.events.once('finish', () => {

        if (digest !== hash.digest('base64')) {
            request.raw.req.destroy(throw new Boom.badData('nice error message'));
        }
    });
```

matthieusieben:
I did some benchmarking to see the impact of microtasks. Here are my conclusions:


<details>
<summary>Computer</summary>

macOS Catalina 10.15.7 (19H15)
MacBook Pro (13-inch, 2019, Four Thunderbolt 3 ports)
2,8 GHz Intel Core i7 quad core
16 Go 2133 MHz LPDDR3

</details>

<details>
<summary><code>server.js</code></summary>

```js
const Hapi = require('.');
const noop = () => {};
const server = Hapi.server({ port: 8080 });

server.route({
  method: 'POST',
  path: '/',
  handler: (request, h) => h.response(null)
});

server.ext('onRequest', (request, h) => {
  if (request.query.peek === "1") request.events.on('peek', noop);
  if (request.query.finish === "1") request.events.on('finish', noop);
  return h.continue;
});

server.start()
```

</details>

<details>
<summary><code>benchmark.js</code></summary>

```js
const Stream = require('stream');
const Http = require('http');
const noop = () => {}

class TestStream extends Stream.Readable {
  constructor(chunks, chunkSize = undefined) {
    super({ highWaterMark: chunkSize, autoDestroy: true, objectMode: false })
    this.chunks = chunks
  }
  _read(size) {
    if (this.chunks--) this.push(Buffer.alloc(size))
    else this.push(null)
  }
}

const mesure = (qs) => new Promise((resolve, reject) => {
  const req = Http.request({
    protocol: 'http:',
    host: "127.0.0.1",
    port: 8080,
    method: 'POST',
    path: `/?${qs}`,
  }, (res) => {
    res.on('error', end);
    res.on('data', noop);
    res.on('end', end);
  });

  const end = (err) => {
    const diff = process.hrtime(start)
    if (err) reject(err)
    else resolve(diff[0] * 1e9 + diff[1])
  }
  const stream = new TestStream(100000, 1)
  const start = process.hrtime()
  stream.pipe(req)
});

const suite = async () => ({
  'no-tap': (await mesure('peek=0&finish=0')) / 1e6,
  'finish': (await mesure('peek=0&finish=1')) / 1e6,
  'peek': (await mesure('peek=1&finish=0')) / 1e6,
});


const benchmark = async (count = 10) => {
  let result = await suite()
  for (let i = 0; i <= count; i++) {
    const newResult = await suite()
    for (const key in newResult) result[key] += newResult[key]
  }
  for (const key in result) result[key] /= count
  return result
};

benchmark(100).then(console.log)
```

</details>

The idea is to compare the following behavours:
- No listeners (this will disable `tapping` the request stream)
- Finish listener only (`tap` enabled. no microtasks generated by `peek` listeners)
- Peek listeners only (`tap` enabled. several microtasks generated by `peek` listeners)

The bench script is generating a stream payload composed of `100 000` chunks, to try and trigger the peek listeners a lot of times. This corresponds to the number of times `peek` would happen during a transfer of `1,6384 GB` of data  (readable stream's default `highWaterMark` is `16384` bytes). The `server.js` script can easily be adapted to check that the number of chunks (= calls to `peek` listeners) correspond to that value.

Here are the results:
```
  no-tap: 378.2443061600001 ms (avg)
  finish: 370.87199968999994 ms (avg)
  peek: 377.50655967 ms (avg)
```

As you can see (and this comes as a surprise), enabling `tapping` proves to actually improve performances. Enabling `peeking` causes an increase of `6.6` mili-seconds for every 100 000 chunks of data. The impact of micro tasks seems very dim to me...

kanongil:
In my mind, this is no longer a performance issue, but an conceptual _`emit()` should never be awaited on_ issue.

The provided benchmark is not really appropriate, as there is not apparent what is really benchmarked here. It just shows that the difference is too small to detect in relation to the request processing itself (which does not surprise me). Given that this will waste much time waiting for IO, there is no way to determine the actual CPU load of the change.

PR #4174, opened 2020-10-22
Update joi documentation links

Nargonath:
Someone reported on the Slack channel that we have broken Joi documentation link on the website. These links were picked up from the hapi API file.

Fixes https://github.com/hapijs/hapi.dev/issues/414


Nargonath:
We need to update `lab` version for the tests to pass on Node v15.

geek (approved):

PR #4176, opened 2020-10-28
Sync v20 branch with master

Nargonath:
This is needed in order to update the API.md for hapi.dev to pick up.


cjihrig (approved):
LGTM. FWIW, I think it's fine to just push to the v20 branch when doing sync up like this.

Issue #4017, opened 2020-01-10
19.0.0 Release Notes


hueniverse:
## Summary

**hapi** v19.0.0 is a small size release with a few changes to improve performance, support for multiple versions of **joi** schemas, and standards compliance.

- **Upgrade time**: low - none to a couple of hours for most users
- **Complexity**: low - requires following the list of changes to verifying their impact
- **Risk**: low - small number of changes with simple ways to retain previous behavior in most cases
- **Dependencies**: medium - most existing plugins will work as-is, however plugins with scoped names might require changes to retain scope in exposed plugin APIs

## Breaking Changes

- Only support node v12 or newer. Will not work at all on node v11 or older (#3910, #4012, #4013).
- `request.payload` is initialized to `undefined` in `onResponse` and if modified, will cause payload processing to bypass (#4015).
- `server.expose()` when used in a plugin with a scoped name (e.g. `@hapi/example`) will no longer include the scope (e.g. `@hapi/`) prefix in `server.plugins`, unless the new `scope` option is set (#4011).
- No **joi** schema compiler is loaded by default and must be explicitly loaded using `server.validator(require('@hapi/joi'))` if uncompiled schemas are using in configuration (#4006).
- Remove support for request queueing (`server.load.concurrent`) (#3977).
- Change `request.payload.multipart` to `false` by default (#3920).
- Change default empty status code to `204` from `200` (#3919).

## New Features

- Support multiple **joi** schema compilers at the same time allowing each plugin to specify the version of **joi** it is using (#4006).
- Support `SameSite=None` and the ability to customize cookie settings per-request in runtime which is required for some Chrome 80 migrations (#3987).
- Always include auth artifacts regardless of auth status (#4000).

## Bug fixes

- Support request decorations using symbols and runtime apply (#3996).

### Node version

Make sure your node version is v12.14.1 or newer. This release uses language features that are only available in node v12 or higher and will not even start on older versions. Node v12 is the current LTS and the recommended version of node.

### Request payload override

The `request.payload` property is now initialized to `undefined` instead of `null` when accessed in an `onResponse` extension handler. If modified, it will cause payload processing to bypass instead of just being overridden later by the framework.

Checklist:

- Look for `request.payload` access in `onRequest` handlers and make sure you are not expecting `null` specifically. If the value is being modified, note that it will now "stick".

### Exposed plugin APIs

[...]

PR #4186, opened 2020-11-10
Fix default `emptyStatusCode`  in API.md

michalstocki:
According to the release notes of 19.0.0 https://github.com/hapijs/hapi/issues/4017, the default `emptyStatusCode` is now `204` which is how it really works. The documentation here was out of date.


devinivy:
Perfect, thanks for the fix!

PR #4187, opened 2020-11-12
Update documented query parser argument name

jonathansamines:
Updated function signature to keep it consistent with the documented params


Nargonath (approved):

Nargonath:
Thanks @jonathansamines. Nice catch 👍 .

Issue #4194, opened 2020-11-19
Add a 'closing' event


kanongil:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): yes (but have a workaround)
* *is this issue affecting a production system?* (yes/no): yes

#### Context

* *node version*: any
* *module version*: 20.0.2
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): standalone
* *any other relevant information*:

#### What problem are you trying to solve?

Detect when the server has called `close()` on the listener in order to abort some slow processing pending responses.

The correct time to do this is right after `server.listener.close()` has been called since clients that retry the request will be denied at the socket level. Doing it earlier in an `onPreStop` extension risks clients reconnecting.

Currently I manage this by making a custom listener where I intercept the call to `close()`.

#### Do you have a new or modified API suggestion to solve the problem?

Yes. Add a new `'closing'` event on `server.events`, which triggers right after [`server.listener.close()` is called](https://github.com/hapijs/hapi/blob/75e901498f2914ebf9c2ec274b8abef072c274cf/lib/core.js#L462).

Note that the `'close'` event on the listener cannot be used, as it is first called once all listeners have been fully closed.

cjihrig:
Sounds like a reasonable feature request.

Nargonath:
I agree. Are you already working on it @kanongil? Otherwise I'd be happy to.

kanongil:
I use my workaround at the moment, and haven't had the opportunity to make any implementation here. You are very welcome to implement this.

Nargonath:
Closed by #4203

Issue #4184, opened 2020-11-09
Inconsistent event emits


kanongil:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*: n/a
* *module version*: 20.0.1
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...):
* *any other relevant information*:

This is a followup from comment in https://github.com/hapijs/hapi/pull/4173#pullrequestreview-525793514.

#### What problem are you trying to solve?

Have a consistent event interface for Hapi, where event emits cannot delay / abort server processing.

#### Do you have a new or modified API suggestion to solve the problem?

Don't wait for `emit()` promise, on server `"start"` and `"stop"`, and subsequently don't return a promise from `events.emit()`. This should also enable much better performance, especially for events with no listeners.

PR #4203, opened 2020-12-04
Add closing event

Nargonath:
Closes (#4194)

### DIFF
--- a/API.md
+++ b/API.md
@@ -597,7 +597,7 @@ server.events.on('start', () => {
 
 ##### <a name="server.events.closing" /> `'closing'` Event
 
-The `'closing'` event type is emitted when the server is stopped using [`server.stop()`](#server.stop()) but before all underlying connections have been closed, and thus before the [`'stop'`](#server.events.stop) event is triggered.
+The `'closing'` event type is emitted when the server is stopped using [`server.stop()`](#server.stop()). It is triggered when incoming requests are no longer accepted but before all underlying active connections have been closed, and thus before the [`'stop'`](#server.events.stop) event is triggered.
 The `'closing'` event handler uses the function signature `function()`.
 
 ```js
--- a/API.md
+++ b/API.md
@@ -595,6 +595,18 @@ server.events.on('start', () => {
 });
 ```
 
+##### <a name="server.events.closing" /> `'closing'` Event
+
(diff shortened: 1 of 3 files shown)


API.md
```diff
@@ -595,6 +595,18 @@ server.events.on('start', () => {
 });
 ```
 
+##### <a name="server.events.closing" /> `'closing'` Event
+
+The `'closing'` event type is emitted when the server is stopped using [`server.stop()`](#server.stop()) but before all underlying connections have been closed thus before the [`'stop'`](#server.events.stop) event is triggered.
```
cjihrig, line 600:
```suggestion
The `'closing'` event type is emitted when the server is stopped using [`server.stop()`](#server.stop()) but before all underlying connections have been closed, and thus before the [`'stop'`](#server.events.stop) event is triggered.
```

cjihrig (approved):

API.md
Nargonath: Thanks for your suggestion.

kanongil (commented):
Looks good.

I would probably explain that it triggers when incoming connections are no longer accepted, but active request are still being processed.

A test to verify this would also be nice, since there could be a timing issue between the call to `listener.close()` and it actually applying to the underlying socket. Especially, if we later change the `emit()` method to immediately trigger, as suggested in #4184.

devinivy:
Looks good!  If you're going to make any changes to the tests, I might also recommend supporting this statement from the docs with a test:
> and thus before the [`'stop'`](#server.events.stop) event is triggered

In my opinion it could be a part of the same test you already have written.

Nargonath:
Thanks @kanongil for your feedbacks. i'll add a test for that.

Yeah good idea Devin. I'll update my test at the same time.

Nargonath:
@kanongil I have added another test. Just to make sure I understand properly what you suggested, you wanted a test to make sure the events emission order was consistent even though there is an active request while the server is stopped. Also making sure that a denied incoming request would not trigger the closing event. Am I right?

Nargonath:
@kanongil have you got any time to take a look at the second test I added? Is it what you were looking for?

Nargonath:
Is there anything blocking this PR? Should we proceed and merge it? If you think it needs some rework, please feel free to voice your concerns. 😃

devinivy (approved):

Nargonath:
Alright I'll merge this. The feature works and if we want to improve the tests suite later on we can still come back to it. Thanks everyone for your feedbacks.

Issue #4104, opened 2020-06-12
Track cache policies


jonathansamines:
#### Support plan

* *which support plan is this issue covered by?* (e.g. Community, Core, Plus, or Enterprise):  Community
* *is this issue currently blocking your project?* (yes/no): No
* *is this issue affecting a production system?* (yes/no): No

#### Context

* *node version*: Node 12
* *module version*: Hapi 19
* *environment* (e.g. node, browser, native): Node
* *used with* (e.g. hapi application, another framework, standalone, ...): Hapi application
* *any other relevant information*: No

#### What problem are you trying to solve?
We are trying to build a general introspection mechanism for all hapi applications cache, similar to [spring boot's actuator](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-endpoints) endpoints, allowing consumers to:

+ [List all available cache names](https://github.com/spring-projects/spring-boot/blob/v2.3.1.RELEASE/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CachesEndpointWebExtension.java#L43)
+ [Get details about a particular cache stats](https://github.com/spring-projects/spring-boot/blob/v2.3.1.RELEASE/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CachesEndpointWebExtension.java#L43)
+ [Drop cached data](https://github.com/spring-projects/spring-boot/blob/v2.3.1.RELEASE/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CachesEndpointWebExtension.java#L55)

With the existing public API's we have not been able to list all available cache policies created by the framework, either directly through [server methods](https://hapi.dev/api/?v=19.1.1#server.method()) or indirectly by using the [server.cache](https://hapi.dev/api/?v=19.1.1#server.cache()) functionality on plugins

So far, by making use of existing public API's we have been able to:

##### Server methods
We could iterate over [server.methods](https://hapi.dev/api/?v=19.1.1#server.methods) and use the currently exposed `.cache` property to get the policy `stats` and `drop` methods.

###### Get existing cache policies tracked by server methods

```javascript

function getMethodName(ns, methodName) {
  const nsf = ns.length ? `${ns}.` : ns;

  return `${nsf}${methodName}`;
}

function getMethodsStats(ns, stats, methods) {
  for (const [methodName, methodFn] of Object.entries(methods)) {
[...]

PR #4210, opened 2020-12-29
Add support for cache policies instrospection

jonathansamines:
Adding support for cache policies introspection as described by #4104. I am copying the original issue problem statement for easy reference.

We are trying to build a general introspection mechanism for all hapi applications cache, similar to [spring boot's actuator](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-endpoints) endpoints, allowing consumers to:

+ [List all available cache names](https://github.com/spring-projects/spring-boot/blob/v2.3.1.RELEASE/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CachesEndpointWebExtension.java#L43)
+ [Get details about a particular cache stats](https://github.com/spring-projects/spring-boot/blob/v2.3.1.RELEASE/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CachesEndpointWebExtension.java#L43)
+ [Drop cached data](https://github.com/spring-projects/spring-boot/blob/v2.3.1.RELEASE/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CachesEndpointWebExtension.java#L55)

So far, by making use of existing public API's we have been able to introspect cache policies created by:

##### Server methods
We can iterate over [server.methods](https://hapi.dev/api/?v=19.1.1#server.methods) and use the exposed `.cache` property to get the policy `stats` and `drop` methods.

###### Get existing cache policies tracked by server methods

While it is possible to introspect cache policies from server methods, it is not ideal as we have to implement several safeguards to get it right.

```javascript

function getMethodName(ns, methodName) {
  const nsf = ns.length ? `${ns}.` : ns;

  return `${nsf}${methodName}`;
}

function getMethodsStats(ns, stats, methods) {
  for (const [methodName, methodFn] of Object.entries(methods)) {
    // server method without namespace
    if (typeof methodFn === 'function') {
      // not all server methods have a cache available
      if (methodFn.cache) {
        stats[getMethodName(ns, methodName)] = methodFn.cache.stats;
      }
    } else {
      // namespaced server methods
      getMethodsStats(getMethodName(ns, methodName), stats, methodFn);
    }
  }

  return stats;
}
```

###### Get a particular server method caching policy
```javascript
const method = Hoek.reach(server.methods, 'segment-name');

if (method && method.cache) {
  return method.cache.drop('key');
}
```

##### Plugins
Unfortunately, a large source of application cache policies often comes from external plugins (e.g [yar](https://hapi.dev/module/yar/)) and while some of them expose their cache policies, is not practical to expect each plugin author to expose their policies and manually keep track of them.

#### What changes are we introducing?
The framework now keeps track of all cache policies created by using `server.cache` and exposes them through a single `server.caches` property.

```javascript
const cache = server.cache({ segment: 'countries', expiresIn: 60 * 60 * 1000 });
const policy = server.caches._default.countries;

assert(cache === policy);
```

**Note:** The main difference from proposal as described by #4104 is that segments are not directly available, as I didn't originally consider that segments can be reused across cache provisions.

**Note 2:** As you can see, when the default cache is provisioned, it is available through the `_default` identifier.

With that API in place, we can now create some API endpoints to introspect the cache policies:

##### List all cache policy stats
```javascript
server.route({
  method: 'GET',
  path: '/actuator/cache',
  handler() {
    const stats = {};

    for (const [cacheName, cache] of Object.entries(server.caches)) {
      stats[cacheName] = {};
      for (const [segmentName, policy] of Object.entries(cache.segments)) {
        stats[cacheName][segmentName] = policy.stats;
      }
    }

    return stats;
  }
});
```

##### Get a single cache policy stats

```javascript
server.route({
  method: 'GET',
  path: '/actuator/cache/{cacheName}/{segment},
  handler(request) {
    const policy = server.caches?.[request.params.cacheName]?.segments[request.params.segment];

    if (policy) {
      return policy.stats;
    }

    throw Boom.notFound('cache policy segment or key not found');
});
```

##### Drop a cache key within a cache policy

```javascript
server.route({
  method: 'GET',
  path: '/actuator/cache/{cacheName}/{segment}/{key}',
  async handler(request, h) {
    const policy = server.caches?.[request.params.cacheName]?.segments[request.params.segment];

    if (policy) {
      await policy.drop(request.params.key);
      return h.response().code(204);
    }

    throw Boom.notFound('cache policy segment or key not found');
});
```

Please let me know your thoughts!


test/server.js
```diff
@@ -195,6 +195,37 @@ describe('Server', () => {
...
 
+    describe('caches', () => {
+
+        it('shows cache segments within default cache provision', () => {
+
+            const server = Hapi.server();
+            const cache = server.cache({ segment: 'test', expiresIn: 1000 });
+
+            expect(server.caches._default.shared).to.be.equal(false);
+            expect(server.caches._default.segments.test).to.be.equal(cache);
```
devinivy, line 206: This assertion is a bit simpler and stricter.
```suggestion
            expect(server.caches._default.segments.test).to.shallow.equal(cache);
```

test/server.js
```diff
@@ -195,6 +195,37 @@ describe('Server', () => {
...
+        });
+
+        it('shows cache segments within named cache provision', async () => {
+
+            const server = Hapi.server();
+            await server.cache.provision({ provider: CatboxMemory, name: 'named' });
+            const cache = server.cache({ cache: 'named', segment: 'test', expiresIn: 1000 });
+
+            expect(server.caches.named.shared).to.be.equal(false);
+            expect(server.caches.named.segments.test).to.be.equal(cache);
```
devinivy, line 216:
```suggestion
            expect(server.caches.named.segments.test).to.shallow.equal(cache);
```

test/server.js
```diff
@@ -195,6 +195,37 @@ describe('Server', () => {
...
+        });
+
+        it('shows cache segments within shared cache provision', async () => {
+
+            const server = Hapi.server();
+            await server.cache.provision({ provider: CatboxMemory, name: 'shared', shared: true });
+            server.cache({ cache: 'shared', segment: 'test', expiresIn: 1000 });
+
+            expect(server.caches.shared.shared).to.be.equal(true);
```
devinivy, line 225: Is there a nice way we could disambiguate the cache name from the `shared` property here?  Naming is hard!

test/server.js
jonathansamines: Nice, I like the explicitness of the `shallow` modifier. Will update.

API.md
```diff
@@ -400,6 +400,23 @@ Access: read only.
...
+
+- `segments` - an object where each key is a segment name and the values are [cache policies](https://hapi.dev/family/catbox/api#policy) created with [server.cache()](#server.cache()).
+- `shared` - indicates if the cache provision allows shared segments.
+
+```js
+const Hapi = require('@hapi/hapi');
+const server = Hapi.server({ port: 80 });
+const cache = server.cache({ segment: 'countries', expiresIn: 60 * 60 * 1000 });
+
+console.log(server.caches['_default'].segments.countries === cache);  // true
```
devinivy, line 417: I believe this is the first time that the name of the default cache has been public.  If we want to give it a different name or expose a symbol for it, now might be a good time.

devinivy: @hapijs/tsc any thoughts on that?

devinivy:
I like how this came out, and I support landing it as a feature.  Thanks for the complete contribution and laying out all the details!

test/server.js
jonathansamines: I looked for [similar words](https://www.thesaurus.com/browse/shared?s=t) and `common` looks like a possible alternative. Thoughts?

devinivy: Sounds good!  Can we name the cache `'common'` for this test?

jonathansamines: Sure. will update

API.md
cjihrig: I agree with @devinivy here.

kanongil: I'm fine with `_default`. It is already a reserved name since you can't provision a cache with such a name.

API.md
```diff
@@ -400,6 +400,23 @@ Access: read only.
...
 [`server.auth.default()`](#server.auth.default()).
 
+#### <a name="server.caches" /> `server.caches`
+
+Access: read only.
+
+An object containing the cache provisions, where each key is a cache name and the values are objects with the following properties:
+
+- `segments` - an object where each key is a segment name and the values are [cache policies](https://hapi.dev/family/catbox/api#policy) created with [server.cache()](#server.cache()).
+- `shared` - indicates if the cache provision allows shared segments.
```
kanongil, line 410: What is the point of exposing `shared`?

API.md
```diff
@@ -400,6 +400,23 @@ Access: read only.
 Contains the default authentication configuration if a default strategy was set via
 [`server.auth.default()`](#server.auth.default()).
 
+#### <a name="server.caches" /> `server.caches`
+
+Access: read only.
+
+An object containing the cache provisions, where each key is a cache name and the values are objects with the following properties:
+
+- `segments` - an object where each key is a segment name and the values are [cache policies](https://hapi.dev/family/catbox/api#policy) created with [server.cache()](#server.cache()).
```
kanongil, line 409: Given that this is likely only ever going to be iterated over, I would prefer that it is exposed as a `Map`.

Actually, I would prefer an array with objects containing `segment` & `policy` properties. This would allow multiple entries for the same segment when sharing is enabled.

kanongil (commented):
It definitely makes sense to expose the created policies. I would do it in a slightly different way, as per my comments.

As for the implementation, I skipped the review for now. But I can say that it is faulty, since it does not expose all policies when multiple are created for the same segment (when `shared` is enabled).

API.md
kanongil: And that array could just be the policy objects if we make the `_segment` property on the policy public (as `segment`).

kanongil:
Instead of exposing the cache provisions, with associated policies, this would be simpler if just exposed as an array of created policies. The same information would be exposed, as each policy includes a reference to the provisioned cache client.

lib/core.js
```diff
@@ -361,7 +366,7 @@ exports = module.exports = internals.Core = class {
 
         try {
             const caches = [];
-            this.caches.forEach((cache) => caches.push(cache.client.start()));
+            this.caches.clients.forEach((client) => caches.push(client.start()));
```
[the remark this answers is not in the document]
devinivy: I believe that it is done this way because `caches.clients` is a `Map`.

API.md
devinivy: In my opinion the underscore doesn't just imply that it's internal, but that it's actually private.

jonathansamines: So, by @kanongil comment, renaming `_default` may actually lead to a breaking change because an application could already have such name in use. Using a symbol might be a way to avoid that. 

@devinivy I am assuming that your suggestion would be to use an internal symbol, which we would have to expose in order to somehow "serialize" it's name right?

API.md
```diff
@@ -400,6 +400,23 @@ Access: read only.
 Contains the default authentication configuration if a default strategy was set via
 [`server.auth.default()`](#server.auth.default()).
 
+#### <a name="server.caches" /> `server.caches`
```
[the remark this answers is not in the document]
jonathansamines: I tried to adhere to the existing documentation groups. I am under the impression that *properties* are explicitly set in a separate group than *methods*. For instance, that is why `server.method()` and `server.methods` are not documented together.

Please let me know if I should still proceed and document them together. As for the API name change, I do agree it better describes what the property contains.

API.md
jonathansamines: I like the idea of exposing the `_segment` property as `segment`. We are referring to [this property](https://github.com/hapijs/catbox/blob/60db7cd7fbed5521f0eb690630868f4e42609eb5/lib/policy.js#L78) right?

jonathansamines:
@kanongil I think that would work as long as we also expose the *cache name* associated to each policy, we'll need to have that, to disambiguate between policy objects created by different shared caches. Perhaps by exposing an additional property? i.e `policy.cache`

jonathansamines:
Considering the feedback received, I am thinking a better API would be:

```js
  const server = Hapi.server();
  await server.cache.provision({ provider: CatboxMemory, name: 'common', shared: true });
  const cacheA = server.cache({ cache: 'common', segment: 'test', expiresIn: 1000 });
  const cacheB = server.cache({ cache: 'common', segment: 'test', expiresIn: 2000 });
  
  const [policyA, policyB] = server.cache.policies.filter((policy) => policy.segment === 'test' && policy.cache === 'common');

  console.log(policyA === cacheA);  // true
  console.log(policyB === cacheB);  // true
```

That would solve the following problems:

- We no longer expose provision information. May be a good thing, as provisions were not previously exposed in any way, while policies where always returned by `server.cache()`
- Would account for the cases when segments can be shared. With the previous iteration, shared caches would result in only the last policy segment to be tracked.

The new API will require the following changes:
- Update catbox to rename the internal `_segment` property to `segment`, therefore exposing it as part of it's public API
- Extend policies returned by catbox, with a new `cache` property. This would allow to disambiguate different provisions with the same segment name

We'll still have to define whether or not we want to rename the default cache. As for now, I am more inclined to keep the current `_default` name, as it would allow for this change to be released without breaking changes. 

Thoughts?

kanongil:
> @kanongil I think that would work as long as we also expose the _cache name_ associated to each policy, we'll need to have that, to disambiguate between policy objects created by different shared caches. Perhaps by exposing an additional property? i.e `policy.cache`

Ah, yes the name would not be exposed (which would actually make the `_default` naming issue a moot point).

I would probably add an optional `name` property to the policy `options` and expose it through a `.name` property. If we map the `"_default"` name to `undefined`, we can also avoid naming it.

Alternatively, we would need a way to map clients to names. But this is impossible, since multiple provisions could share the same client.

jonathansamines:
@kanongil I am thinking it would also make sense to update `Client` with the optional `.name` property instead of `Policy`. Since `Policy` already has a public property to access `Client` (`.client`) that could be enough for our needs. 

It also would accommodate nicely the existing model where we name provisions (clients). But now, catbox would also have access to those names.

jonathansamines:
Created https://github.com/hapijs/catbox/pull/236 to add the missing public APIs on catbox.

kanongil:
I started thinking about how this tracking fits into lifecycle management, and I am afraid there are bad news towards your efforts.

Fundamentally, there are no lifecycle management for policies generated from `server.cache()`, and they can in principle be used ephemerally. Once a consumer is done with a policy, it just needs to dereference it. As such, remembering these inside the server (in `server.cache.policies`) can lead to memory leaks, and should not be done.

So how do we enable your use-case? This would need to be done using a configurable hook, which is called whenever a policy is generated with the appropriate data (`policy`, `cacheName`, `segment`). An event seems suitable for this, but policies can be provisioned at any point after server creation, so registering a listener in a plugin might miss some early ones. Alternatively it could be exposed as a callback server option, but that would make plugin integration very cumbersome.

Regardless of how this is triggered, your system would just listen for this event/callback, and do your own tracking (preferably using [`WeakRef`s and finalizers](https://v8.dev/features/weak-references) which are supported since node v12.4). 

How does this sound? Or am I overthinking it?

API.md
Nargonath: I believe that's what @devinivy meant.

jonathansamines:
> Fundamentally, there are no lifecycle management for policies generated from server.cache(), and they can in principle be used ephemerally. Once a consumer is done with a policy, it just needs to dereference it. As such, remembering these inside the server (in server.cache.policies) can lead to memory leaks, and should not be done.

Right, would that only be a problem if an application creates an indefinite (say, creating policies on demand) amount policies, or normal usage would also represent the same problem? The same could be said of cache provisions?

> Regardless of how this is triggered, your system would just listen for this event/callback, and do your own tracking (preferably using WeakRefs and finalizers which are supported since node v12.4).

I am not quite familiar with WeakRefs yet, but if I can safely implement this feature at a plugin level by using events and WeakRefs, is it possible to implement the same at framework level?

Finally, how about tracking policies as an opt-in feature? I am thinking a server option would do well do enable this kind of functionality.

kanongil:
> > Fundamentally, there are no lifecycle management for policies generated from server.cache(), and they can in principle be used ephemerally. Once a consumer is done with a policy, it just needs to dereference it. As such, remembering these inside the server (in server.cache.policies) can lead to memory leaks, and should not be done.
> 
> Right, would that only be a problem if an application creates an indefinite (say, creating policies on demand) amount policies, or normal usage would also represent the same problem? The same could be said of cache provisions?

For server method caching I guess the lifetime follows the server, so this only applies to `server.cache()` policies. And yes, here it is possible to create on-demand policies. I don't know of any ephemeral use of this, but it is something that we implicitly allow.

> > Regardless of how this is triggered, your system would just listen for this event/callback, and do your own tracking (preferably using WeakRefs and finalizers which are supported since node v12.4).
> 
> I am not quite familiar with WeakRefs yet, but if I can safely implement this feature at a plugin level by using events and WeakRefs, is it possible to implement the same at framework level?

We could do this, but adding somewhat complicated logic to track something that we don't use internally seems not ideal. Besides, even applied correctly, it won't perfectly track the usage and expired policies could linger for some time.

As I propose, you would be responsible for the tracking, and you could choose to track the lifetimes, or just add a condition to using your tracking and require consumers to not use this kind of rarely-used?? ephemeral policies. Or live with the leaks depending on how serious they are.

> Finally, how about tracking policies as an opt-in feature? I am thinking a server option would do well do enable this kind of functionality.

I don't think we want to do this in hapi itself. We presently lack a mechanism for plugins to track the usage, but this will be fixed if we add the proposed event. Eg.

```js
server.events.on('policy', (policy, cacheName, segment) => {
    …
});
```

jonathansamines:
@kanongil That sounds good enough for my use-case. Will give the implementation try.

jonathansamines:
Added policy event in https://github.com/hapijs/hapi/pull/4219

Nargonath:
It seems the changes proposed in this PR won't be implemented and a new PR has been created to provide an event that would allow cache policies tracking in userland. Shouldn't we close this PR in favor #4219 @jonathansamines?

PR #4221, opened 2021-01-19
Fix tests for citgm

devinivy:
There are a few issues to fix for failures in citgm:

1. Tests that rely on lsof on node v14.  @cjihrig previously made a fix for this in https://github.com/hapijs/hapi/pull/3744, but on node v14 the way to detect this error changed. https://github.com/nodejs/citgm/pull/436#issuecomment-754156977
2. Some flaky/failing tests on macos and AIX.  As far as I can tell the issue here is flakiness due to timing-dependent tests and slow IO in CI.  I attempted to make these tests less timing-dependent and to better cleanup after themselves, which allows them to be retryable.  I would love some review of these changes, as they are some of the more complex tests in the hapi test suite. https://github.com/nodejs/citgm/pull/436#issuecomment-751786801

CC @AdriVanHoudt I know in https://github.com/nodejs/citgm/pull/436 you were able to reproduce the macos issues.  If you have time and would be open to trying out these changes it would be very much appreciated.

### DIFF
--- a/test/request.js
+++ b/test/request.js
@@ -863,6 +863,7 @@ describe('Request', () => {
         it('skips onPreResponse when validation terminates request', { retry: true }, async (flags) => {
 
             const server = Hapi.server();
+            const abortedReqTeam = new Teamwork.Team();
 
             let called = false;
             server.ext('onPreResponse', (request, h) => {
@@ -890,6 +891,8 @@ describe('Request', () => {
 
                                 const raw = context.app.request;
                                 await Events.once(raw.req, 'aborted');
+
+                                abortedReqTeam.attend();
                             }
                         }
                     }
@@ -901,6 +904,8 @@ describe('Request', () => {
 
             const req = Http.get(server.info.uri, Hoek.ignore);
             req.on('error', Hoek.ignore);
+
+            await abortedReqTeam.work;
 
             await server.events.once('response');
 
@@ -2324,17 +2329,14 @@ describe('Request', () => {
             const req = Http.request(options);
             onCleanup.unshift(() => req.destroy());
 
-            req.on('error', (err) => {
-
-                expect(err).to.not.exist();
-            });
-
             req.write('\n');
 
             const [res] = await Events.once(req, 'response');
 
             expect([503, 408]).to.contain(res.statusCode);
             expect(timer.elapsed()).to.be.at.least(80);
+
+            await Events.once(req, 'close'); // Ensures that req closes without error
         });
     });
 
--- a/test/request.js
+++ b/test/request.js
@@ -278,22 +278,19 @@ describe('Request', () => {
 
         it('exits handler early when request is no longer active', { retry: true }, async (flags) => {
 
+            let testComplete = false;
+
             const onCleanup = [];
             flags.onCleanup = async () => {
+
+                testComplete = true;
 
                 for (const cleanup of onCleanup) {
                     await cleanup();
                 }
             };
 
-            let testComplete = false;
-            onCleanup.unshift(() => {
-
-                testComplete = true;
-            });
-
             const server = Hapi.server();
-            const enterHandlerTeam = new Teamwork.Team();
             const leaveHandlerTeam = new Teamwork.Team();
 
             server.route({
@@ -302,7 +299,7 @@ describe('Request', () => {
                 options: {
                     handler: async (request, h) => {
 
-                        enterHandlerTeam.attend();
+                        req.abort();
 
                         while (request.active() && !testComplete) {
                             await Hoek.wait(10);
@@ -319,15 +316,11 @@ describe('Request', () => {
             });
 
             await server.start();
-            onCleanup.unshift(async () => await server.stop());
+            onCleanup.unshift(() => server.stop());
 
             const req = Http.get(server.info.uri, Hoek.ignore);
-
             req.on('error', Hoek.ignore);
-            onCleanup.unshift(() => req.off('error', Hoek.ignore));
 
-            await enterHandlerTeam.work;
-            req.abort();
             const note = await leaveHandlerTeam.work;
 
             expect(note).to.equal({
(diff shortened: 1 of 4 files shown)


test/request.js
```diff
@@ -2294,26 +2337,17 @@ describe('Request', () => {
...
-
-                    expect([503, 408]).to.contain(res.statusCode);
-                    expect(timer.elapsed()).to.be.at.least(80);
-                    resolve();
-                });
+            const res = await new Promise((resolve, reject) => {
 
-                req.on('error', (err) => {
-
-                    expect(err).to.not.exist();
```
devinivy, line 2308: In some (flaky) cases this error was being triggered late and causing the next test to fail.

test/request.js
```diff
@@ -2294,26 +2337,17 @@ describe('Request', () => {
...
 
-            await new Promise(async (resolve) => {
-
-                const req = Http.request(options, (res) => {
-
-                    expect([503, 408]).to.contain(res.statusCode);
-                    expect(timer.elapsed()).to.be.at.least(80);
-                    resolve();
-                });
+            const res = await new Promise((resolve, reject) => {
```
kanongil, line 2340: This seems wrong. `req` errors will effectively be ignored since they would trigger after the `response` event, rejecting into an already resolved `Promise`.

test/request.js
```diff
@@ -847,10 +867,10 @@ describe('Request', () => {
 
     describe('_postCycle()', () => {
 
-        it('skips onPreResponse when validation terminates request', async () => {
+        it('skips onPreResponse when validation terminates request', { retry: true }, async (flags) => {
```
kanongil, line 870: This rewrite seems correct. Only some nits.

test/request.js
```diff
@@ -275,45 +276,64 @@ describe('Request', () => {
...
+                testComplete = true;
+            });
 
             const server = Hapi.server();
-            const team = new Teamwork.Team();
-            const team2 = new Teamwork.Team();
+            const enterHandlerTeam = new Teamwork.Team();
+            const leaveHandlerTeam = new Teamwork.Team();
 
-            let rounds = 0;
```
kanongil, line 284: Ah, you essentially replaced the rounds limit with a while loop with a `testComplete` failsafe.

test/request.js
```diff
@@ -275,45 +276,64 @@ describe('Request', () => {
...
+
+                for (const cleanup of onCleanup) {
+                    await cleanup();
+                }
+            };
+
+            let testComplete = false;
+            onCleanup.unshift(() => {
+
+                testComplete = true;
```
kanongil, line 292: This needs to always run before other cleanups. Otherwise you risk that a previous cleanup never resolves, and `testComplete` would never be set, causing an infinite loop in the handler.

I would just put the logic directly into the `onCleanup` handler before the loop.

test/request.js
```diff
@@ -275,45 +276,64 @@ describe('Request', () => {
...
+                            testComplete
+                        });
+
                         return null;
                     }
                 }
             });
 
             await server.start();
+            onCleanup.unshift(async () => await server.stop());
```
kanongil, line 322: No need for `await`.

test/request.js
```diff
@@ -275,45 +276,64 @@ describe('Request', () => {
...
+
+            const req = Http.get(server.info.uri, Hoek.ignore);
 
-            const req = Http.get(server.info.uri, (res) => { });
             req.on('error', Hoek.ignore);
-            await team2.work;
+            onCleanup.unshift(() => req.off('error', Hoek.ignore));
```
kanongil, line 327: Why? This means that a late `req` error will leak into other tests.

test/request.js
```diff
@@ -275,45 +276,64 @@ describe('Request', () => {
...
                 method: 'GET',
                 path: '/',
                 options: {
                     handler: async (request, h) => {
 
-                        team2.attend();
-                        for (let i = 0; i < 100; ++i) {
-                            ++rounds;
-                            await Hoek.wait(10);
+                        enterHandlerTeam.attend();
```
kanongil, line 305: The `enterHandlerTeam` can be replaced by moving the `req.abort()` to here.

kanongil (commented):
Mostly nits, but the "handles race condition between equal client and server timeouts" rewrite seems like it is not doing the same thing.

test/request.js
devinivy: That is true.  But we do know that the test will already be over, so it would be a late error.  This was occurring when stopping the server causes the client to experience a remotely closed connection (presumably racing with `req.end()`?), and sometimes it would land while a future test was running.  I will try to handle it a bit better.

devinivy:
@kanongil I appreciate the review, and incorporated your feedback.

Nargonath (approved):
LGTM

test/request.js
```diff
@@ -2327,25 +2311,28 @@ describe('Request', () => {
...
+            const req = Http.request(options);
+            onCleanup.unshift(() => req.destroy());
 
-                const req = Http.request(options, resolve);
-                req.once('error', reject);
+            req.on('error', (err) => {
 
-                req.write('\n');
-                onCleanup.unshift(() => req.end());
+                expect(err).to.not.exist();
```
kanongil, line 2329: This can trigger after the test is done. I would just listen for the `close` using `Events.once()`:

```js
await Events.once(req, 'close');
```

The registers an 'error' listener, and will fail the test if triggered.

kanongil (commented):
Still a few nits.

test/request.js
devinivy: Good call, this feels a lot better.

kanongil (approved):

devinivy:
I appreciate all the careful review 👍

AdriVanHoudt:
For what it's worth I still see these on master
```
Failed tests:

  393) Request active() exits handler early when request is no longer active:

      Timed out (5000ms) - Request active() exits handler early when request is no longer active


  417) Request _postCycle() skips onPreResponse when validation terminates request:

      Timed out (5000ms) - Request _postCycle() skips onPreResponse when validation terminates request
```
I don't have the bandwidth to dive into this atm but if I can help by running some tests or something let me know. (for quicker response dm me on slack)

PR #4234, opened 2021-03-03
Allow for res to have already closed during transmission

devinivy:
There's a fun story here!

In https://github.com/nodejs/citgm/pull/436 it was brought to our attention that node v15.7.0 caused a test to reliably fail in hapi's test suite.  I found that the test fails due to the relative timing of request `abort` versus response `close` timing: prior to node v15.7.0 the response would close in a later tick than the request would abort, and now these occur in the same tick.  This means that during hapi's transmission step, where it would typically still assume that the request or response would still have some events to emit (namely, a response `close`), that there would be no more event from `req` and `res`.  The first thing I did was adjust this test to remove the timing discrepancy across node versions, so that it always takes the most aggressive approach that we see occurring in node v15.7.0 (always aborts _and closes_ during marshaling).

I searched for a nice node-builtin way to detect this case, but I didn't find anything that stuck.  The most promising one was to check for `res.destroyed`, but it turns out that this was set in node v15, but not in node v14.  So I decided to add some state to hapi's request that tracks whether the response is closed, which to the best of my knowledge indicates that we can't expect to see any more events from `req` and `res`.

I ensured this code path continued to go through hapi's ending routine as though `close` were emitted, even though it ends-up being a no-op.  Honestly, adding some more state to `request` to track this and triggering a `finish` on `res` was not my favorite approach, but it was hard for me to find something more appropriate.  Certainly open to suggestions.  One aspect that I wouldn't mind some review on is whether the new event listener could introduce some kind of leak since it doesn't stop referencing `request` as the other listeners do through `_eventContext`— I think we're all good there, but it was on my mind.  @kanongil of course no pressure, but if you do have time to take a look, I think you are one of the people best-suited to review this area of hapi and any input you have would be appreciated.

### DIFF
--- a/test/transmit.js
+++ b/test/transmit.js
@@ -538,10 +538,15 @@ describe('transmission', () => {
             // Use state autoValue function to intercept marshal stage
 
             server.state('always', {
-                autoValue() {
+                async autoValue(request) {
 
-                    client.destroy();
-                    return team.work;               // Continue once the request has been aborted
+                    const close = new Teamwork.Team();
+                    request.raw.res.once('close', () => close.attend());
+
+                    client.destroy();               // Will trigger abort then close. Prior to node v15.7.0 the res close came
+                    await close.work;               // asynchronously after req abort, but since then it comes in the same tick.
+
+                    return team.work;               // Continue marshalling once the request has been aborted and response closed.
                 }
             });
 
--- a/test/transmit.js
+++ b/test/transmit.js
@@ -543,8 +543,10 @@ describe('transmission', () => {
                     const close = new Teamwork.Team();
                     request.raw.res.once('close', () => close.attend());
 
-                    client.destroy();               // Will trigger abort then close. Prior to node v15.7.0 the res close came
-                    await close.work;               // asynchronously after req abort, but since then it comes in the same tick.
+                    // Will trigger abort then close. Prior to node v15.7.0 the res close came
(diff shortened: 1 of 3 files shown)


test/transmit.js
```diff
@@ -538,10 +538,15 @@ describe('transmission', () => {
...
-                autoValue() {
+                async autoValue(request) {
 
-                    client.destroy();
-                    return team.work;               // Continue once the request has been aborted
+                    const close = new Teamwork.Team();
+                    request.raw.res.once('close', () => close.attend());
+
+                    client.destroy();               // Will trigger abort then close. Prior to node v15.7.0 the res close came
+                    await close.work;               // asynchronously after req abort, but since then it comes in the same tick.
```
nlf, line 547: total nit, but the multiline comment to the right here should be above these two lines. as it is, it makes it look like the second line of the comment is regarding the second line of the code which is not the case

nlf (approved):
i wasn't able to identify a better way to track this either. this is isolated enough and clear enough that i think it's reasonable :+1:

test/transmit.js
devinivy, line 547:
```suggestion
                    // Will trigger abort then close. Prior to node v15.7.0 the res close came
                    // asynchronously after req abort, but since then it comes in the same tick.
                    client.destroy();
                    await close.work;
```

PR #4239, opened 2021-03-20
Fix abort test timing to be consistent from node v12 through v16

devinivy:
This fixes a test failure surfaced by citgm against node v16 nightly https://github.com/nodejs/citgm/issues/852.  In the past we have seen this test be [a bit flaky on OSX](https://github.com/hapijs/hapi/runs/2155975306)— hopefully this addresses that issue as well.

The issue is that the test relies on the relative timing of various node events.  Previously three requests were made to hapi, and during the first one the requests are aborted by the client.  For previous versions of node, the first two of those three requests were able to make it through before the server saw the client disconnect.  In node v16 the disconnect occurs nearly immediately, and only the first of the three requests were able to make it through.  To fix this, I ensure that the second and third requests experience the disconnect during the pre-handler.

### DIFF
--- a/test/request.js
+++ b/test/request.js
@@ -512,7 +512,7 @@ describe('Request', () => {
             expect(info.remoteAddress).to.exist();
         });
 
-        it('handles aborted requests (pre response)', { retry: true }, async () => {
+        it('handles aborted requests (before response)', { retry: true }, async (flags) => {
 
             const server = Hapi.server();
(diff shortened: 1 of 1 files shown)


nlf (approved):
👍

cjihrig (approved):

devinivy:
Thanks for the speedy review!

PR #4242, opened 2021-04-09
Mention response.charset() clear behavior

Nargonath:
Someone opened a PR on DefinitelyTyped repo regarding an undocumented behavior of `response.charset()` in the typings. It's also undocumented in the API documentation so I thought we ought to add it. Source: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52285#issuecomment-816824526

### DIFF
--- a/API.md
+++ b/API.md
@@ -4325,7 +4325,7 @@ Return value: the current response object.
 
 Sets the 'Content-Type' HTTP header 'charset' property where:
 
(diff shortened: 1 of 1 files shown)


lloydbenson (approved):

devinivy (approved):
I confirmed there are tests supporting this behavior 👍

PR #4250, opened 2021-05-08
Update to mimos v6

devinivy:
As followup to https://github.com/hapijs/mimos/pull/37, this PR updates hapi to use mimos v6.


lib/config.js
```diff
@@ -259,7 +259,7 @@ internals.server = Validate.object({
     })
         .unknown()
         .default(),
-    mime: Validate.object().allow(null).default(null),
+    mime: Validate.object().empty(null).default(),
```
devinivy, line 262: This change accounts for the fact that the `Mimos` constructor no longer accepts `null`.

cjihrig (approved):

lloydbenson (approved):

Issue #4256, opened 2021-06-03
inert + hapi may be leaking file descriptors


deej-split:
#### Support plan

Free plan

* *is this issue currently blocking your project?* (yes/no):  no (actually, sorta... )
* *is this issue affecting a production system?* (yes/no): yes

#### Context

* *node version*: 12.18.4
* *module version with issue*: inert@6.0.3
* *last module version without issue*: unknown
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): @hapi/hapi@20.1.3
* *any other relevant information*: There may or may not be a conversation about this on the community slack channel: https://hapihour.slack.com/archives/C6CEEUE06/p1622675671016200

#### What are you trying to achieve or the steps to reproduce?

We are serving static files with Hapi and Inert, and we are using up all the file descriptors on the host system.

We are calling:
```js
      const response = h.file(filename, {
        confine: true,
        etagMethod: false,
      });
```

One day some of our microservices suddenly reported errors that the number of file descriptors on their systems were exhausted.

Some more exploration let me to check the number of file descriptors on the linux systems this was running on `ls /proc/${pid}/fd` and I saw these appeared to be incrementing by one for each file.  I then identified that this also happens on my development machine (a mac) using `lsof -p ${pid}`. From here, I began to debug and `console.log()` may way through inert, and gradually found that the `prepare()` routine in Inert's `file.js` is getting called twice on each request.  But only one of the file descriptors are being closed.  The two calls have these stack traces:

```
    at exports.File.open (/Users/demo/project/node_modules/@hapi/inert/lib/fs.js:35:22)
    at async exports.File.openStat (/Users/demo/project/node_modules/@hapi/inert/lib/fs.js:88:9)
    at async Object.internals.prepare (/Users/demo/project/node_modules/@hapi/inert/lib/file.js:111:22)
    at async exports.Manager.execute (/Users/demo/project/node_modules/@hapi/hapi/lib/toolkit.js:105:28)
    at async Object.internals.handler (/Users/demo/project/node_modules/@hapi/hapi/lib/handler.js:46:20)
    at async exports.execute (/Users/demo/project/node_modules/@hapi/hapi/lib/handler.js:31:20)
    at async Request._lifecycle (/Users/demo/project/node_modules/@hapi/hapi/lib/request.js:372:32)
    at async Request._execute (/Users/demo/project/node_modules/@hapi/hapi/lib/request.js:280:9)
```

```
[...]

PR #4257, opened 2021-06-04
Add response lifecycle tracking and checks

kanongil:
This is in response to #4256.

I added asserts to verify the expected state so this kind of error are less likely to be introduced in future code. This also caught another error, where the `marshal()` method could sometimes be called _after_ `close()`.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -55,10 +55,11 @@ exports = module.exports = internals.Response = class {
 
         this._events = null;
         this._payload = null;                       // Readable stream
-        this._error = null;                         // The boom object when created from an error (used for logging)
+        this._error = options.error || null;        // The boom object when created from an error (used for logging)
         this._contentType = null;                   // Used if no explicit content-type is set and type is known
         this._takeover = false;
         this._statusCode = false;                   // true when code() called
+        this._state = 'init';                       // One of 'init', 'prepared', 'marshalled', 'closed'
 
         this._processors = {
             marshal: options.marshal,
@@ -495,17 +496,24 @@ exports = module.exports = internals.Response = class {
 
     _prepare() {
 
-        this._passThrough();
-
-        if (!this._processors.prepare) {
-            return this;
-        }
+        Hoek.assert(this._state === 'init');
 
         try {
-            return this._processors.prepare(this);
+            this._passThrough();
+
+            if (!this._processors.prepare) {
+                return this;
+            }
+
+            try {
+                return this._processors.prepare(this);
+            }
+            catch (err) {
+                throw Boom.boomify(err);
+            }
         }
-        catch (err) {
-            throw Boom.boomify(err);
+        finally {
+            this._state = 'prepared';
         }
     }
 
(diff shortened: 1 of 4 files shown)


lib/response.js
```diff
@@ -638,19 +653,28 @@ exports = module.exports = internals.Response = class {
...
-        if (this._processors.close) {
-            try {
-                this._processors.close(this);
+        if (this._state === 'closed') {
+            return;
+        }
+
+        try {
+            if (this._processors.close) {
+                try {
```
nlf, line 662: in general i'm not a fan of nesting `try` blocks, and i'm not totally clear why both are needed here.

what was the condition that was causing the outer block to throw?

lib/response.js
```diff
@@ -495,17 +496,24 @@ exports = module.exports = internals.Response = class {
...
 
         try {
-            return this._processors.prepare(this);
+            this._passThrough();
+
+            if (!this._processors.prepare) {
+                return this;
+            }
+
+            try {
```
nlf, line 508: curious about the nested try block here too

lib/response.js
```diff
@@ -558,67 +566,74 @@ exports = module.exports = internals.Response = class {
...
-        if (this._processors.marshal) {
-            try {
-                source = await this._processors.marshal(this);
-            }
-            catch (err) {
-                throw Boom.boomify(err);
+            // Processor marshal
+
+            if (this._processors.marshal) {
+                try {
```
nlf, line 577: this one too

test/response.js
```diff
@@ -1290,6 +1290,37 @@ describe('Response', () => {
             const res = await server.inject('/');
             expect(res.statusCode).to.equal(500);
         });
+
+        it('is only called once for returned responses', async () => {
```
nlf, line 1294: so i only see one added test here, and while it does demonstrate nicely that it fixes the original issue, you commented

> This also caught another error, where the marshal() method could sometimes be called after close().

is this something that we can also add a test for? were you able to determine the conditions that caused them to be called out of order? it would be nice to ensure this doesn't regress

nlf:
as a note, the failing builds here are unrelated to these changes and are fixed in #4258

lib/response.js
kanongil: This is for the `finally` clause which will update the state. It can definitely be done in other ways.

test/response.js
kanongil: Thanks for the concern. The tests already include a case that exercise this.

If you remove the `if` from here, you will see that it fails [this](https://github.com/hapijs/hapi/blob/adf341f25996393e6d3bc557203687b29f10281c/test/transmit.js#L526) test: https://github.com/hapijs/hapi/blob/adf341f25996393e6d3bc557203687b29f10281c/lib/transmit.js#L38-L45

nlf: > The tests already include a case that exercise this.

so the original behavior with that test was running the functions out of order but not crashing because there were no assertions. introducing these assertions that things are run in order caused that test to result in a crash, and the if statement you added there fixes the problem so instead of running them out of order we simply don't run `marshal()` at all.

did i understand that right?

lib/response.js
kanongil: If you configure the github diff view to "hide whitespace changes" using the gear icon at the top, it will be more visible what I did.

nlf: i feel like collapsing these into one `try/catch/finally` is maybe more readable.

though in this case it would have the consequence that if `internals.Response.drain(stream)` were to throw for some reason we would log it with the `response, cleanup, error` tags. the other blocks would have similar concerns that we may need to think about on a case by case basis too.

i don't really think that's a problem though, personally, and it would maintain the existing goal of making sure that the state is updated correctly. what do you think?

test/response.js
kanongil: Yes, that is exactly it.

lib/response.js
kanongil: I think I would rather just set the new state immediately after the initial check. This also means that if the logic somehow recurses to calling itself, it will fail the assert.

nlf: given that the intent is to set the state regardless of the outcome of the logic in the block, i think that's completely reasonable

kanongil: I have pushed a cleanup patch, which undoes the try-finally and changed the verbiage to use present tense. Hope that resolves this.

kanongil:
One outstanding question is whether the asserts should remain in the code. While the only purpose they serve is to catch hapi internal errors, I think that it is warranted to ensure we maintain consistency and limit the risk of regressing.

nlf (approved):
this looks good to me :+1:

nlf:
> One outstanding question is whether the asserts should remain in the code. While the only purpose they serve is to catch hapi internal errors, I think that it is warranted to ensure we maintain consistency and limit the risk of regressing.

i agree, i think keeping the assertions in place for now is fine. i'd rather we blow up with an error a user can open an issue with than do something out of order and send incorrect responses for :sparkles: reasons :sparkles:

kanongil:
Rebased to fix the test failures.

devinivy (approved):
I think it is _slightly_ bold for us to introduce these assertions, but I also appreciate that it's probably best for the longterm.  This looks great, thanks for the contribution as always, Gil.  Release on the way.

PR #4259, opened 2021-06-07
Update invalid path test to match message

arafel:
I noticed on running the tests that 'returns 400 on invalid path' (currently test 392) fails because of a text mismatch. I checked back to at least v20.0.3 and it failed in the same way. The text of the test hasn't changed since 2018.

I wasn't able to find out where the text is generated, unfortunately, but took a guess that the new text is okay. If someone can point me to where the error text comes from I'm happy to look again. :-)

Thanks
Paul

### DIFF
--- a/test/request.js
+++ b/test/request.js
@@ -346,7 +346,7 @@ describe('Request', () => {
 
             const res = await server.inject('invalid');
             expect(res.statusCode).to.equal(400);
-            expect(res.result.message).to.equal('Invalid URL: invalid');
+            expect(res.result.message).to.equal('Invalid URL');
         });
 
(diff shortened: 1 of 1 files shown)


kanongil:
Thanks, but this is already handled in the pending PR #4259.

Nargonath:
I think you meant this PR #4258 which is now merged but thank you @arafel for raising the issue. 😃

kanongil:
Wow, wrong link yeah ☺️

arafel:
Thanks both. I have to confess it didn’t occur to me that was a Node change - no wonder all the commits I tested failed. 😄

Issue #4262, opened 2021-06-18
Streaming uploads cause truncated responses in node 16


achingbrain:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): yes
* *is this issue affecting a production system?* (yes/no): yes

#### Context

* *node version*: 16
* *module version with issue*: 20.1.4
* *last module version without issue*: Unsure
* *environment* (e.g. node, browser, native): node 
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi application
* *any other relevant information*:

#### What are you trying to achieve or the steps to reproduce?

Take a look at the following code sample.  It starts a server, makes a request to the server, sends some data, and finally prints out the response.

It has two methods of reading the data from the client, one treating `request.payload` as an async iterator, one treating it as a readable stream.

```js
const Hapi = require('@hapi/hapi')
const { PassThrough } = require('stream')
const http = require('http')

function receiveDataAsStream (request, stream) {
  request.payload.on('data', () => {
    console.info('receive data')
  })
  request.payload.on('end', () => {
    console.info('end data')

    // send result to client
    stream.end('hello world')
  })
}

async function receiveDataAsAsyncIterator (request, stream) {
  for await (const buf of request.payload) {
    console.info('receive data')
  }

  // send result to client
  stream.end('hello world')
}

async function main () {
  const server = Hapi.server({
    port: 0,
    host: 'localhost'
  })

  server.route({
    method: 'POST',
    path: '/',
    options: {
      payload: {
        output: 'stream'
      }
    },
    handler: (request, h) => {
      const stream = new PassThrough()

      // simulate async work
      process.nextTick(() => {
        // works on node 14 and node 16
        // receiveDataAsStream(request, stream)

        // works on node 14, breaks on node 16
        receiveDataAsAsyncIterator(request, stream).catch(err => console.error(err))
      })

      return h.response(stream)
    }
  })

  await server.start()

  // make a request
  const req = http.request({
    hostname: 'localhost',
    port: server.info.port,
    method: 'POST'
  }, (res) => {
    res.on('data', (chunk) => {
      console.info(chunk.toString())
    })
    res.on('end', () => {
      console.info('response ended')
      server.stop().catch(err => console.error(err))
    })
  })

  console.info('send data')
  req.end('{}')
}

main().catch(err => {
  console.error(err)
  process.exit(1)
})
```

On node 16 the `'hello world'` output is not received by the client, on node 14 it is.

[...]

PR #4263, opened 2021-06-21
Harden stream processing close test

kanongil:
I fixed the test to be compatible with node 16, which allows the `req` to be closed while the connection itself is still active.

I also revised the stream, so we can really validate that the processing is stopped.

This is in response to https://github.com/hapijs/hapi/issues/4262#issuecomment-864614011.

### DIFF
--- a/test/transmit.js
+++ b/test/transmit.js
@@ -1190,7 +1190,9 @@ describe('transmission', () => {
             await log;
         });
 
-        it('stops processing the stream when the request closes', async () => {
+        it('stops processing the stream when the connection closes', async () => {
+
+            let stream;
 
             const ErrStream = class extends Stream.Readable {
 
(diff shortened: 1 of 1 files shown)


devinivy (approved):
That feels much better.

One thing I want to gut-check with you: do you believe that this test simulated bad/unexpected behavior from node's http server, and in turn it's not something critical to cover in node v14 and below?  I see that this test has existed since the very early days of the framework, so perhaps it was more important when node was less stable/mature.

kanongil:
As I see it, the test validates that streams stop being processed when a request stops due to downstream / internal issues. My patch updates it so that it better does this with v16 (emitting `close` on `req` can no longer be used for this).

Seem legit, but might be tested by some other case. Removing it does not change coverage.

PR #4264, opened 2021-06-23
Fix req end during response transmission

devinivy:
Addresses #4262.  On node v16 hapi would cut short any response that was transmitting while the request stream ended.  This is not especially common, partly because the request stream is often used prior to transmitting the response, and partly because this case does not apply when the request stream is not read at all i.e. for GET requests.

I believe this change is sound and backwards compatible because on node versions prior to v16, both req and res `'close'` events have identical meaning: the underlying connection shared by req and res has closed.  For this reason, there is no need to add listeners to both req and res on node <v16.

Starting with node v16 the `'close'` event for req no longer relates to the underlying connection, best documented at https://github.com/nodejs/node/issues/38924 and https://github.com/nodejs/node/pull/33035.  This means that it is possible for req to `'close'` prior to res, notably while transmitting res.  The change in this PR is intended to take that into account.  The test added here was failing on node v16 prior to this change.

See also #4263, which fixes a test that would have failed with the changes made in this PR.

### DIFF
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -257,7 +257,6 @@ internals.pipe = function (request, stream) {
     const end = internals.end.bind(null, env, null);
 
     request.raw.req.on('aborted', aborted);
-    request.raw.req.on('close', close);
 
     request.raw.res.on('close', close);
     request.raw.res.on('error', end);
--- a/test/transmit.js
+++ b/test/transmit.js
@@ -7,11 +7,13 @@ const Net = require('net');
 const Path = require('path');
 const Stream = require('stream');
 const Zlib = require('zlib');
+const Events = require('events');
 
 const Boom = require('@hapi/boom');
 const Code = require('@hapi/code');
 const Hapi = require('..');
 const Hoek = require('@hapi/hoek');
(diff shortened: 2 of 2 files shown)


kanongil (approved):
Great. I would probably have used `server.inject()` for the test but this works.

devinivy:
That's a good point.  Originally it was not intentional to sidestep `server.inject()`, but rather result of working with the original repro provided in #4262.  I did notice early on that the repro provided there using async iterators did not reproduce with `server.inject()`.  I just looked into moving the new test to use it, and in fact shot isn't super useful to simulate this condition, as shot is more tied to the pre-v16 behavior (I am definitely looking at https://github.com/hapijs/shot/pull/139 with new eyes).  I could emit some events from req/res to simulate it manually, but in the end I am glad to have a real http test for this odd case.  Thanks as always for the review!  I hope/intend to land this soon.

nlf (approved):
this seems to maintain the intent of the original handlers while coping with the subtle change in node 16. i think this is fine to land as-is, but we should really closely examine our event handlers at some point in the future

Issue #4214, opened 2021-01-07
Cannot skip required payload authentication with injected credentials


jonathansamines:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): No
* *is this issue affecting a production system?* (yes/no): No

#### Context

* *node version*: v12.19.1
* *module version with issue*:  20.0.3
* *last module version without issue*: I believe this has always behaved this way
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): framework
* *any other relevant information*:

#### What are you trying to achieve or the steps to reproduce?
When trying to inject credentials to an authentication scheme requiring payload authentication, then the payload authentication cannot be skipped.

```js
'use strict';

const assert = require('assert');
const Hapi = require('@hapi/hapi');

async function run() {
  const server = Hapi.server();

  server.auth.scheme('scheme', () => ({
    authenticate() {
      throw new Error('authenticate error');
    },
    payload() {
      throw new Error('payload error');
    },
    options: {
      payload: true,
    },
  }));

  server.auth.strategy('default', 'scheme');

  server.route({
    method: 'POST',
    path: '/test',
    options: {
      auth: 'default',
      handler(request) {
        return request.auth.credentials;
      },
    },
  });

  const credentials = {
    hello: 'world',
  };

  const { result } = await server.inject({
    method: 'POST',
    url: '/test',
    auth: {
      strategy: 'default',
      credentials,
    },
  });

  assert.deepStrictEqual(result, credentials);
}

run()
  .catch((error) => {
    console.log(error);
    process.exit(1);
  });

```

#### What was the result you got?
The test fails because payload authentication could not be skipped, while the scheme authentication was.

#### What result did you expect?
Both the payload authentication and scheme authentication are skipped, when credentials are injected. If we update the authentication scheme to not require payload authentication, the test would pass.

I am not actually sure this is the expected behavior or not, for now I manually skip the payload authentication on my scheme when credentials are injected.

cjihrig:
> I am not actually sure this is the expected behavior or not

It's not entirely clear to me either, but from some quick testing, the test suite fails if payload validation is skipped on injected requests 😄

jonathansamines:
@cjihrig Is there any chance to change that and skip payload authentication when credentials are injected? Does it make sense?

cjihrig:
[...]

PR #4274, opened 2021-08-19
Add optional payload authentication skip on credentials injection

jonathansamines:
Adds optional support to disable payload authentication when credentials are injected. Solves #4214


Nargonath (approved):
LGTM thanks for the contribution 👍

devinivy (approved):

PR #4281, opened 2021-09-05
Fix handling of auth scheme/strategy realms

### LINKED ISSUE #4218: Realms and authentication strategies
#### Support plan

* *is this issue currently blocking your project?* (yes/no): No
* *is this issue affecting a production system?* (yes/no): No

#### Context

* *node version*: v14.15.4
* *module version*: v20.0.3
* *environment* (e.g. node, browser, native): Node

#### How can we help?

While building an authentication strategy, I found myself depending on the current realm "parent" relationship, which I had originally assumed to work just like in plugins. Turns out, it works quite differently for authentication strategies:

```js
'use strict';

const Hapi = require('@hapi/hapi');
const assert = require('assert');

function scheme(server) {
  return {
    authenticate(request, h) {
      const credentials = {
        plugin: server.realm.plugin,
        parentPlugin: server.realm.parent.plugin,
      };

      return h.authenticated({ credentials });
    },
  };
}

async function run() {
  const server = Hapi.server();

  const Plugin = {
    name: 'plugin',
    async register(srv) {
      const Nested = {
        name: 'nested',
        register(srv2) {
          srv2.auth.scheme('scheme-nested', scheme);
        },
      };

      srv.auth.scheme('scheme-plugin', scheme);

      await srv.register(Nested);
    },
  };

  await server.register(Plugin);

  server.auth.scheme('scheme', scheme);
  server.auth.strategy('strategy', 'scheme');
  server.auth.strategy('strategy-plugin', 'scheme-plugin');
  server.auth.strategy('strategy-nested', 'scheme-nested');

  const routes = [['/test', 'strategy'], ['/test-plugin', 'strategy-plugin'], ['/test-nested', 'strategy-nested']];

  for (const [path, strategy] of routes) {
    server.route({
      method: 'GET',
      path,
      options: {
        auth: strategy,
        handler(request) {
          return request.auth.credentials;
        },
      },
    });
  }

  const { result } = await server.inject('/test');
  const { result: resultPlugin } = await server.inject('/test-plugin');
  const { result: resultNested } = await server.inject('/test-nested');

  assert.strictEqual(result.plugin, undefined);
  assert.strictEqual(resultPlugin.plugin, undefined);
  assert.strictEqual(resultNested.plugin, undefined);

  assert.strictEqual(result.parentPlugin, 'nested');
  assert.strictEqual(resultPlugin.parentPlugin, undefined);
  assert.strictEqual(resultNested.parentPlugin, undefined);
}

run()
  .catch((error) => {
    console.error(error);
    process.exit(1);
  });

```

I had a hard time understanding why the test above passed, but after some consideration I found that authentication strategies "parent" realm relationship would refer to the latest "realm" created, regardless of where in the plugin tree it was registered.

Not sure that is the expected behavior or not, but a better documentation of how authentication strategies realm's work would have helped. Would you accept a pull request adding a bit of additional information to the "server.realm" section?

devinivy:
We were seeing some hard-to-explain behavior in #4218 as it relates to the realm of the `server` inside of auth schemes.  In short, each time `server` is cloned, _all_ servers were being mutated such that `server.auth.strategy` was bound to the server of the most recent clone.

Cloning is the process by which hapi creates independent realms for a. each plugin and b. each auth strategy.  When `server` is cloned, the clone's `realm.parent` is assigned to `server.realm`.  Due to the mutation of a shared interface during each clone, the `server.realm` within an auth strategy had a strange/unpredictable chain of ancestor realms.  This work tidies that up by no longer mutating the shared `auth` on each clone, instead creating a new public auth interface for each one.

Resolves #4218.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -18,7 +18,8 @@ exports = module.exports = internals.Auth = class {
     #schemes = {};
     #strategies = {};
 
-    api = {};
+
+    api = {};                   // Do not reassign api or settings, as they are referenced in public()
     settings = {
         default: null           // Strategy used as default if route has no auth settings
     };
@@ -26,6 +27,20 @@ exports = module.exports = internals.Auth = class {
     constructor(core) {
 
         this.#core = core;
+    }
+
+    public(server) {
+
+        return {
+            api: this.api,
+            settings: this.settings,
+            scheme: this.scheme.bind(this),
+            strategy: this._strategy.bind(this, server),
+            default: this.default.bind(this),
+            test: this.test.bind(this),
+            verify: this.verify.bind(this),
+            lookup: this.lookup.bind(this)
+        };
     }
 
     scheme(name, scheme) {
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -18,7 +18,6 @@ exports = module.exports = internals.Auth = class {
     #schemes = {};
     #strategies = {};
 
-
     api = {};                   // Do not reassign api or settings, as they are referenced in public()
     settings = {
         default: null           // Strategy used as default if route has no auth settings
--- a/test/auth.js
+++ b/test/auth.js
@@ -255,7 +255,6 @@ describe('authentication', () => {
                 }
             });
 
-
             const handler = (request) => request.auth.credentials;
             server.route({ method: 'GET', path: '/a', handler, options: { auth: 'a' } });
             server.route({ method: 'GET', path: '/root', handler, options: { auth: 'root' } });
--- a/test/auth.js
+++ b/test/auth.js
@@ -228,6 +228,48 @@ describe('authentication', () => {
 
             expect(server.auth.api.xyz.x).to.equal(5);
         });
+
+        it('has its own realm', async () => {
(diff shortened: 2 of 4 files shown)


lib/auth.js
```diff
@@ -18,7 +18,8 @@ exports = module.exports = internals.Auth = class {
     #schemes = {};
     #strategies = {};
 
-    api = {};
+
```
devinivy, line 21:
```suggestion
```

test/auth.js
```diff
@@ -228,6 +228,48 @@ describe('authentication', () => {
...
+            await server.register({
+                name: 'plugin-a',
+                register(srv) {
+
+                    pluginA = srv;
+
+                    srv.auth.strategy('a', 'custom');
+                }
+            });
+
```
devinivy, line 257:
```suggestion
```

geek (approved):

jonathansamines:
Thanks for taking care of this issue @devinivy!

jonathansamines (approved):

PR #4283, opened 2021-09-12
Fix tests for node@16 error format change

### LINKED ISSUE #4282: Adapt tests to support V8 9.3
#### Support plan

#### Context

* *node version*: v16.9.0
* *module version*: 20.1.5
* *used with* (e.g. hapi application, another framework, standalone, ...): CITGM

Run tests with Node.js 16.9.0 (https://github.com/nodejs/node/pull/40011).

Two of them fail because V8 changed the error message when trying to read properties from `null` or `undefined`.

```
   103) Core _debug() outputs 500 on ext exception:
       actual expected
       "Cannot read propertiesproperty 'here' of null (reading 'here')"
       Expected "Cannot read properties of null (reading 'here')" to equal specified value: "Cannot read property 'here' of null"
       at /home/iojs/tmp/citgm_tmp/08e39eff-f7ef-4ead-92fa-cc5d18384c97/@hapi/hapi/test/core.js:338:44
   231) handler execute() returns 500 on handler exception (next tick await):
       actual expected
       "Cannot read propertiesproperty 'here' of null (reading 'here')"
       Expected "Cannot read properties of null (reading 'here')" to equal specified value: "Cannot read property 'here' of null"
       at /home/iojs/tmp/citgm_tmp/08e39eff-f7ef-4ead-92fa-cc5d18384c97/@hapi/hapi/test/handler.js:76:44
```

https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2754/nodes=fedora-latest-x64/testReport/junit/(root)/citgm/_hapi_hapi_v20_1_5/

Adapt the following assertions to support the new error message:

https://github.com/hapijs/hapi/blob/c2107e9bc9c522c3778e90a3629a0c84f776f9fe/test/handler.js#L65
https://github.com/hapijs/hapi/blob/c2107e9bc9c522c3778e90a3629a0c84f776f9fe/test/core.js#L338

Nargonath:
Closes #4282.

### DIFF
--- a/test/core.js
+++ b/test/core.js
@@ -20,10 +20,13 @@ const Inert = require('@hapi/inert');
 const Lab = require('@hapi/lab');
 const Vision = require('@hapi/vision');
 const Wreck = require('@hapi/wreck');
+const Somever = require('@hapi/somever');
 
 const Common = require('./common');
 
-const internals = {};
+const internals = {
+    isNode16OrHigher: Somever.match(process.version, '>=16')
+};
 
 
 const { describe, it } = exports.lab = Lab.script();
@@ -335,7 +338,8 @@ describe('Core', () => {
             expect(res.statusCode).to.equal(500);
 
             const [, event] = await log;
-            expect(event.error.message).to.equal('Cannot read property \'here\' of null');
+            const expectedErrorMessage = internals.isNode16OrHigher ? 'Cannot read properties of null (reading \'here\')' : 'Cannot read property \'here\' of null';
+            expect(event.error.message).to.equal(expectedErrorMessage);
         });
     });
 
--- a/test/core.js
+++ b/test/core.js
@@ -25,6 +25,7 @@ const Somever = require('@hapi/somever');
 const Common = require('./common');
 
 const internals = {
+    // needed to handle changes in error output
     isNode16OrHigher: Somever.match(process.version, '>=16.9.0')
 };
 
--- a/test/core.js
+++ b/test/core.js
@@ -25,7 +25,7 @@ const Somever = require('@hapi/somever');
 const Common = require('./common');
 
 const internals = {
-    isNode16OrHigher: Somever.match(process.version, '>=16')
+    isNode16OrHigher: Somever.match(process.version, '>=16.9.0')
 };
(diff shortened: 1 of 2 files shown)


Nargonath:
Forgot I had to keep the old message format for node < 16 😅

devinivy:
This update to V8 occurred in [node v16.9.0](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V16.md#2021-09-07-version-1690-current-targos).  Could/should we check for node >= v16.9.0 rather than >= v16, even if just to to help keep this precisely documented?

Another option is to base the check on the version of V8 `process.versions.v8`.  V8 doesn't follow semver, but the first two numbers reflect the chromium milestone divided by 10 (e.g. chromium M93 <-> v8 9.3), so a comparison like this should work:

```js
const toChromiumMilestone = (ver) => parseInt(ver.split('.').slice(0, 2).join(''), 10);

toChromiumMilestone(process.versions.v8) >= 93
```

Nargonath:
Yes you're right @devinivy. I'll make the test more precise.

FWIW I prefer to test against the node version than `v8`. It makes it more easily understandable to others IMO.

kanongil (commented):
This seems a bit over-engineered. I would just change the test to:

```js
expect(event.error.message).to.include('Cannot read prop');
```

Nargonath:
@kanongil The only problem I see with testing that way is that it could shadow a similar error for another field which might not be the problem we want to test here. It's probably unlikely though.

kanongil:
I know my suggestion is not perfect, but I still think it is a better solution due to the simplicity.

cjihrig:
> The only problem I see with testing that way is that it could shadow a similar error for another field

You could add a second assertion for the field name.

Nargonath:
> You could add a second assertion for the field name.

Yes I was just about to say it, you beat me to it.

> I know my suggestion is not perfect, but I still think it is a better solution due to the simplicity.

I like your suggestion @kanongil, I'll do it that way and assert on the field name also.

kanongil (approved):
👍

PR #4284, opened 2021-09-15
Add http2 option to allow creating http2 servers directly

mkg20001:
This makes it more convinient to use http2 in hapi

### DIFF
--- a/API.md
+++ b/API.md
@@ -160,6 +160,15 @@ If the `listener` needs to be manually started, set [`autoListen`](#server.optio
 
 If the `listener` uses TLS, set [`tls`](#server.options.tls) to `true`.
 
+#### <a name="server.options.http2" /> `server.options.http2`
+
+Default value: `false`.
+
+If `true`, hapi will use http2 instead of http/https.
+
+If a certificate is specified, the http2 server will be tls-enabled (h2),
+otherwise it will listen in cleartext mode (h2c).
+
 #### <a name="server.options.load" /> `server.options.load`
 
 Default value: `{ sampleInterval: 0 }`.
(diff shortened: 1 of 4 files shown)


kanongil:
Thanks for the PR. I have previously used the node.js http2 compat server implementation with hapi but found the implementation to be quite broken once you start to poke at the edges. I tried to workaround several issues, but eventually gave up and use a reverse proxy instead.

If hapi adds this option it implies that a HTTP/2 server is supported and stable. Something that I find simply isn't true. As such, I do not believe we should merge this PR at this time.

mkg20001:
Ok. Is there any overview of the HTTP2 issues? Perhaps a branch with failed tests would be a way to start

mkg20001:
> but eventually gave up and now use a reverse proxy instead.

The motivation behind this PR was to use h2c (unencrypted HTTP2) everywhere, including from reverse proxy to app.

kanongil:
It's not hapi that fails, its node. The current implementation is only really suitable for testing purposes. Check the node repo for a list of open issues: https://github.com/nodejs/node/issues?q=is%3Aopen+http2+label%3Ahttp2

For the reverse proxy, you will need to [have it on the same host / container along with node](https://blog.christianposta.com/microservices/00-microservices-patterns-with-envoy-proxy-series/). I have actually considered creating a npm module with an envoy proxy to transparently enable a working HTTP/2 client.

kanongil:
FYI I don't believe node will ever have a stable HTTP/2 implementation. The code needs a complete rework to have a stable foundation, but that will never happen since HTTP/2 is already old news (HTTP/3 is the hotness) and it would inevitably cause breakage with current code that expects it to work like it does in its current broken form.

Issue #4294, opened 2021-10-04
Server crashes if a requested aborted during sending a response. 


mshustov:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): Yes
* *is this issue affecting a production system?* (yes/no): No

#### Context

* *node version*: v14.17.6
* *module version with issue*: v20.2.0
* *last module version without issue*: 
* *environment* (e.g. node, browser, native): browser
* *used with* (e.g. hapi application, another framework, standalone, ...): "@hapi/inert": "6.0.4"
* *any other relevant information*:

#### What are you trying to achieve or the steps to reproduce?

I spent the whole day struggling with the flaky integration tests that were caused by unhandled promise rejection in Hapi sever when an incoming request is **aborted**.
The server crashes with 
```
│ proc [kibana] TypeError: Cannot read property 'on' of null
   │ proc [kibana]     at Object.internals.chain (@hapi/hapi/lib/transmit.js:365:18)
   │ proc [kibana]     at Object.internals.transmit (@hapi/hapi/lib/transmit.js:120:30)
   │ proc [kibana]     at Object.exports.send (@hapi/hapi/lib/transmit.js:28:25)
   │ proc [kibana]     at Request._reply (@hapi/hapi/lib/request.js:459:9)
```

What I learned from the code inspection so far:
- Hapi receives an incoming request and calls a route handler
- the route handler responds with a JSON object
- Hapi started sending a request, https://github.com/hapijs/hapi/blob/master/lib/transmit.js#L28. Note that Hapi already checked the response in not failed! https://github.com/hapijs/hapi/blob/master/lib/transmit.js#L22-L24
- the incoming request is aborted
- Hapi overrides the response with Boom error `Request aborted` https://github.com/hapijs/hapi/blob/master/lib/request.js#L728
- server starts transmitting the original response containing the JSON object
- server crashes. 

I guess it might happen because of a discrepancy between `response` and `response.request.response` content in `transmit` method https://github.com/hapijs/hapi/blob/master/lib/transmit.js#L28
I worked around the problem locally by adding the next snippet at the very top of `transmit`.
```js
    if (response.request.response.isBoom) {
      await internals.fail(request, response.request.response);
      return;
    }
```
I don't think it's a proper fix but I don't know Hapi codebase to fix the problem on my own so I'd appreciate any hints from the maintainers before I can open a PR.

[...]

Issue #4244, opened 2021-04-12
Response event for aborted requests has status code 200 and doesn't call native `response.end`


dnechay:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*: 12.13.1, 14.15.1
* *module version with issue*: 19.2.0, 20.0.1, 20.1.2 (maybe others) 
* *last module version without issue*: unsure
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi-application, another framework
* *any other relevant information*: similar to https://github.com/hapijs/hapi/issues/4072

#### What are you trying to achieve or the steps to reproduce?

```js
const Hapi = require('@hapi/hapi');

const server = Hapi.server({
  host: 'localhost',
  port: 8080,
});

server.route({
  method: 'GET',
  path: '/abort-test',
  handler: async function() {
    return new Promise(function (resolve) {
      setTimeout(function () {
        resolve('OK');
      }, 4000);
    });
  },
})

server.events.on('request', (request, event) => {
  console.log(`Event: ${JSON.stringify(event)}`);
});

server.events.on('response', (request) => {
  console.log(`Response Status Code: ${request.response.statusCode}`);
});

server.start();
console.log(`Server running at: ${server.info.uri}`);
```

```
curl --max-time 1 http://localhost:8080/abort-test
```

#### What was the result you got?
```
Server running at: http://localhost:8080
Event: {"request":"<SHORTENED>","timestamp":1584125752386,"tags":["request","error","abort"],"channel":"internal"}
Response Status Code: 200
```
[Native node.js response.end](https://nodejs.org/docs/latest-v14.x/api/http.html#http_response_end_data_encoding_callback) isn't called.

#### What result did you expect?
`response.end` called, response status code to be 499.

#### Additional info
We're using [opentelemetry.js](https://github.com/open-telemetry/opentelemetry-js) to collect metrics from our application, it has [a module](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-instrumentation-http) that automatically collects HTTP metrics by wrapping native node's `http` and `https` modules.
It works fine with usual HTTP requests to our hapi-app, but when a request is aborted on a client side - it doesn't collect metrics.
[...]

PR #4295, opened 2021-10-05
Revised request / inject abort handling

kanongil:
This is a targeted PR to fix the crash issue in #4294. Unfortunately it grew quite complex due to inconsistencies around how aborts & transmission errors are reported.

The issue in #4294 was mostly covered by this existing test: https://github.com/hapijs/hapi/blob/18495f785d602ee23bb01c6dccebb5297e731d7b/test/transmit.js#L528

The only thing missing is to trigger the `internals.chain()` loop in transmit.js, which I have done by adding a `accept-encoding: gzip` header to the request.

Most of the PR comes from revising `server.inject()` to handle an inconsistency, where it converts transmission errors to appear as actual server responses. This is contrary to regular http clients, which will emit / throw an `error`. I'm not sure whether this can be considered a bug fix, but its plausible given that it is supposed to work similar to a regular http client, and the docs does not cover how transmission errors are signalled.

FYI, I felt I had to do something, since currently 499 request abortion errors have two different response signatures depending on which stage of the requests it arrives at. One is a simple Boom error, while the other is a regular response. Ie. the `statusCode` is either found at the logged `response.statusCode` or `response.output.statusCode`.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1946,6 +1946,9 @@ Return value: a response object with the following properties:
 
 - `request` - the [request object](#request).
 
+Throws a Boom error if the request processing fails. The partial response object is exposed on
+the `data` property.
+
 ```js
 const Hapi = require('@hapi/hapi');
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -363,9 +363,7 @@ exports = module.exports = internals.Request = class {
     async _lifecycle() {
 
         for (const func of this._route._cycle) {
-            if (this._isReplied ||
-                !this._eventContext.request) {
-
+            if (this._isReplied) {
                 return;
             }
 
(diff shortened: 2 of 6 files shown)


kanongil:
Why are there no CI checks active?

devinivy:
@kanongil odd!  Looking into it.

cjihrig:
Yes, there was an issue on GitHub's side: https://www.githubstatus.com/incidents/bdbzpz7qxmbx

kanongil:
> I am pleased to see that this addresses #4244 without running into the awkward coverage issue I ran into, which I think is a good sign.

I'm not sure this PR fully addresses #4244. I have made no effort towards always calling `res.end()`.

PR #4314, opened 2021-11-09
Fix tests for node v17

devinivy:
While we will not officially support node v17 since it is not an LTS version of node, in practice we would like to test hapi against it and ensure everything will be in good shape when node v18 eventually arrives.

The hapi test suite was affected by the change in DNS ordering in https://github.com/nodejs/node/pull/39987, since default hapi servers would need to be connected to using `::1` over ipv6 rather than `127.0.0.1` over ipv4.  In the next major version of hapi we should consider how we might make our `address` default (currently `0.0.0.0`) more ipv6-aware.

We are also currently affected by a bug https://github.com/nodejs/node/issues/40528 slated to be resolved in node v17.1.0.  The one `skip()` in the test suite relates to this: we skip the test for now to avoid a hang caused by the bug.

~~Note that the test suite is going to be all red right now due to https://github.com/nodejs/node/issues/40528.  Once that is resolved I will re-run the tests and turn this into a non-draft PR if everything is passing.~~

### DIFF
--- a/test/transmit.js
+++ b/test/transmit.js
@@ -1196,8 +1196,7 @@ describe('transmission', () => {
             await log;
         });
 
-        // NOTE do not merge with this skip. Can be removed once fix in node v17.1.0 is released.
-        it.skip('stops processing the stream when the connection closes', async () => {
+        it('stops processing the stream when the connection closes', async () => {
 
             let stream;
 
--- a/test/transmit.js
+++ b/test/transmit.js
@@ -25,11 +25,13 @@ const Common = require('./common');
 const internals = {};
 
 
-const { describe, it } = exports.lab = Lab.script();
+const { describe, it, before } = exports.lab = Lab.script();
 const expect = Code.expect;
 
 
 describe('transmission', () => {
+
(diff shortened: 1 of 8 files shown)


test/transmit.js
```diff
@@ -1194,7 +1196,8 @@ describe('transmission', () => {
             await log;
         });
 
-        it('stops processing the stream when the connection closes', async () => {
+        // NOTE do not merge with this skip. Can be removed once fix in node v17.1.0 is released.
+        it.skip('stops processing the stream when the connection closes', async () => {
```
devinivy, line 1200: Node v17.1.0 came out today with the fix for https://github.com/nodejs/node/issues/40528, so this skip shouldn't be needed anymore.  Fingers crossed GH actions pulls the latest version of v17 in the next test run 😆 
```suggestion
        it('stops processing the stream when the connection closes', async () => {
```

cjihrig (approved):

Nargonath (approved):
Thanks for the work @devinivy.

PR #4329, opened 2022-01-17
Update request.info.id docs

jonathansamines:


devinivy (approved):
Wow, a call back to hapi v16— good catch!

Yedfree (approved):

Issue #4343, opened 2022-03-17
Memory leak when passing undefined lifecycle handler


RichPMad:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): No
* *is this issue affecting a production system?* (yes/no): Yes

#### Context

* *node version*: v16.13.2
* *module version with issue*: "@hapi/hapi": "^20.2.1"
* *last module version without issue*: ??
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): standalone
* *any other relevant information*: -

#### What are you trying to achieve or the steps to reproduce?

Hi, I've found a way to trigger a memory leak, by passing an undefined value to a lifecycle hook such as `onPreAuth`. Hapi does not trigger any validation for `server.ext('onPreAuth', UNDEFINED_FN)`.

Passing an undefined value causes hapi to hold on to the request and response objects.

See my example here: https://github.com/RichPMad/hapi-memory-leak-undefined-lifecycle-fn 

#### What was the result you got?

`Request` and `Response` objects are not cleared from memory. No warnings are given, and no validation errors happen when setting up lifecycle hook.

#### What result did you expect?

Lifecycle hooks warn me I'm being dumb and accidentally passing an undefined value.

devinivy:
Hey!  I understand the confusion here, but I think you're running into intentional behavior of the framework.  When you don't pass a method to `server.ext()`, it returns a promise that will resolve with the `request` the next time the extension is invoked.  It's possible that letting these promises hang out could be the cause of a memory leak, though this API is primarily to help make testing more convenient.

Here's the section of the docs: https://hapi.dev/api/#-serverextevent-method-options
> The method may be omitted (if options isn't present) or passed null which will cause the function to return a promise. The promise is resolved with the request object on the first invocation of the extension point. This is primarily used for writing tests without having to write custom handlers just to handle a single event.

Here's a backing test for the behavior: https://github.com/hapijs/hapi/blob/2f80e843d9527c46d5f1ca85e9697e846b65a72b/test/server.js#L1006-L1015

Sorry for any confusion that may have caused!  If you want to ensure you never run into this, you can use `server.ext({ event, method })` instead of `server.ext(event, method)` 👍

RichPMad:
Thanks for the quick response!

Oh wow, I see, sorry I glossed over the docs on this one. 

[...]

PR #4345, opened 2022-04-13
Fix memory leaks from teamwork long standing notes

Nargonath:
This is related to https://github.com/hapijs/teamwork/issues/32. @kanongil spotted a memory leak on https://github.com/hapijs/hapi/issues/4343#issuecomment-1072244342. The leak stems from `hapi`'s `server.ext()` method that leverages long standing Teamwork instances when no method are passed for an extension point. Since `teamwork` didn't clear its internal notes ref once the meeting was over, objects passed to `team.attend()` were never garbage collected and it also kept adding notes to an ended meeting.

The fix was provided in `@hapi/teamwork@5.1.1` thus the PR only updates `hapi` dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@hapi/somever": "^3.0.0",
     "@hapi/statehood": "^7.0.3",
     "@hapi/subtext": "^7.0.3",
-    "@hapi/teamwork": "^5.1.0",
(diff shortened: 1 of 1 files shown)


Nargonath:
Thank you @devinivy . I didn't have bandwidth to test it out as you suggested.

Yedfree (approved):

PR #4346, opened 2022-04-18
Drop node v12 support

### LINKED ISSUE #4279: Drop node v12 support
#### Support plan

#### Context

* *node version*: n/a
* *module version*: v20.x
* *used with* (e.g. hapi application, another framework, standalone, ...): standalone

Allow to develop hapi and modules using more current node.js and javascript APIs and features. v12 is in maintenance mode and support ends 2022-04-30.

New JS features we can use:

1. Optional chaining syntax (`?.`).
2. Nullish coalescing operator (`a ?? b`).
3. Private class methods (`this.#privateMethod()`).
4. `WeakRef` support.

From node 14 (LTS with v14.15.0) we get:

1. ESM modules (stable from v14.17.0, but also in v12.22.0).
2. Top-Level `await` support.
3. A more stable `stream.pipeline()`.
4. `Fs.rm()` method.

If we require v14.17+, we could also sneak in:

1. `AbortSignal` support in node APIs.

N/A.

devinivy:
Dropping node v12 support.  There's a good discussion in #4279 that outlines other steps we may want to take as we end support for node v12.  Here's my approach here:
 - I adopted nullish coalescing and optional chaining.
 - Recognizing that it might break code that relies on the old behavior, I did adopt `??` in the place of `||` when evaluating non-boolean expressions.  I am open to other thoughts, but I am of the mind that it usually gives a better, more precise behavior than `||`.  In some cases I kept `||` when I thought it made sense, e.g to fall through empty header values of `''`.
 - I did not address the usage of private methods since hapi treats private methods as private to hapi and not just the class in question.  One great example is `request._log()` which is called all over the place.  I feel that we should adopt private methods in a given class only if we're willing to remove usage of `_`-prefixed methods from that class that may be used by other classes.
 - I addressed usage of any deprecated APIs I could find: `res.writeableEnd` rather than `res.finished`, `req.socket` rather than `req.connection`.
 - I did not implement anything with `AbortSignal` since I didn't want to make any substantive API changes, though I am in support of it.

This PR will not pass CI until https://github.com/hapijs/shot/pull/144 and https://github.com/hapijs/lab/pull/1036 are published.  I intend to publish beta versions of these (and other) modules in the near future.

Resolves #4279

### DIFF
--- a/.github/workflows/ci-module.yml
+++ b/.github/workflows/ci-module.yml
@@ -3,6 +3,7 @@ name: ci
 on:
   push:
     branches:
+      - v21
       - master
   pull_request:
   workflow_dispatch:
@@ -10,3 +11,5 @@ on:
 jobs:
   test:
     uses: hapijs/.github/.github/workflows/ci-module.yml@master
+    with:
+      min-node-version: 14
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -67,7 +67,7 @@ exports = module.exports = internals.Auth = class {
         Hoek.assert(typeof strategy.authenticate === 'function', 'Invalid scheme:', name, 'invalid authenticate() method');
         Hoek.assert(!strategy.payload || typeof strategy.payload === 'function', 'Invalid scheme:', name, 'invalid payload() method');
         Hoek.assert(!strategy.response || typeof strategy.response === 'function', 'Invalid scheme:', name, 'invalid response() method');
-        strategy.options = strategy.options || {};
+        strategy.options = strategy.options ?? {};
         Hoek.assert(strategy.payload || !strategy.options.payload, 'Cannot require payload validation without a payload method');
 
         this.#strategies[name] = {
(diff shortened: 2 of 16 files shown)


devinivy:
I was able to pull in those beta versions of lab, code, and shot.  We officially have a passing build of hapi on node v18 :)

I ran into a little snag along the way that is now fixed.  Since we've started using `req.socket` in lieu of the deprecated `req.connection`, adding `req.socket` to shot actually put us down some new codepaths in hapi.  Previously these code paths were exercised by shot injection requests, which were always missing `req.socket`.  As a result I needed to make shot's `req.socket` more true to the `Net.Socket` interface, and I needed to add a new test to hapi for when `req.socket` is missing.

geek:
Looks good! I tested on a few of the main plugins as well, no issues.

geek (approved):

PR #4349, opened 2022-04-25
Enforce that response streams are always Stream.Readable

### LINKED ISSUE #4293: Only support native ReadableStream streams
#### Support plan

#### Context

* *node version*: 14+
* *module version*: 20.x
* *environment* (e.g. node, browser, native): 

Handlers currently support returning any `ReadableStream`, even v1 streams from the `readable-stream` module. However using such legacy streams is incompatible with modern streams and requires very careful integration to support.

Simpler, more robust internal stream handling. Eg. we can rely on all streams having the `.destroy()` method and remove tests like: https://github.com/hapijs/hapi/blob/18495f785d602ee23bb01c6dccebb5297e731d7b/test/transmit.js#L1465

Drop node v12 support (#4279) and fail when the returned stream matches `value instanceof Stream && !(value instanceof ReadableStream)`. The [current detection logic](https://github.com/hapijs/hapi/blob/18495f785d602ee23bb01c6dccebb5297e731d7b/lib/streams.js#L14-L23) relies on duck typing.

One drawback is that we increase the reliance on the node runtime, which means that the Chrome "support" as introduced in #3946 will be gone (unless they find a way to properly polyfill `ReadableStream`). I don't know why this was patched in, since we don't do any kind of testing on that platform.

For a more graceful approach, we could pipe() legacy streams through a `Passthrough` stream when returned from a handler. Then internal processing could still rely on the stream being native.

devinivy:
Dropping support for old/non-standard response streams.  If you want to use one of these, you can try repairing it into a `Stream.Readable` by piping it through a `Stream.PassThrough`.  I ended-up keeping the old stream duck-typing logic around for the purpose of helping users catch any issues when moving from v20 to v21, rather than treating legacy streams as an object response.

Resolves #4293

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -676,18 +676,7 @@ exports = module.exports = internals.Response = class {
 
     static drain(stream) {
 
-        if (stream.destroy) {
-            stream.destroy();
-            return;
-        }
-
-        // Fallback for old-style streams
-
-        stream.unpipe();
-
-        if (stream.close) {
-            stream.close();
-        }
+        stream.destroy();
     }
 };
 
--- a/lib/streams.js
+++ b/lib/streams.js
@@ -16,7 +16,7 @@ exports.isStream = function (stream) {
 
     if (!isReadableStream &&
         typeof stream?.pipe === 'function') {
-        throw Boom.badImplementation('Cannot reply with a stream-like object that is not an instance of Stream.Readable.');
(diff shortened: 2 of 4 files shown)


devinivy:
The failure appearing on node v12 will be resolved once #4346 lands and we stop supporting node v12— it's a syntax issue.  I think it's alright to look past this for review purposes.

Nargonath (approved):
Thanks for the work @devinivy

PR #4350, opened 2022-04-26
Make default error available in validation failAction

### LINKED ISSUE #4040: failAction: detailedError to log, defaultError to response
#### Support plan

* *which support plan is this issue covered by?* (e.g. Community, Core, Plus, or Enterprise): Community, Core
* *is this issue currently blocking your project?* (yes/no): yes

#### Context

* *node version*: v12.14.1
* *module version*: 19.0.5
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi app
* *any other relevant information*: -

It wold be awesome to have `failAction` to log [`detailedError`](https://github.com/hapijs/hapi/blob/master/lib/validation.js#L133) to the console, [`defaultError`](https://github.com/hapijs/hapi/blob/master/lib/validation.js#L132) to the response.

Current implementation allows:
```js
failAction: 'error'
// defaultError to log and to response

failAction: (request, h, err) => { request.log('error', err); throw err }
// detailedError to log and to response
```

There is no way to access both `detailedError` and `defaultError`.

https://github.com/hapijs/hapi/blob/master/lib/toolkit.js#L117-L135

```js
if (failAction === 'log-and-error') {
    request._log(options.tags, options.details || err);
    throw err;
}
```

devinivy:
Following from the conversation in #4040, I've added the default validation error to `err.data.defaultError` where `err` is the detailed error passed to `failAction()` for input validation.  Since the common use-case seems to be logging detailed validation errors without modifying the error response, I considered logging the detailed error by default.  But I determined that it contained user input which could be sensitive: I believe users should opt-in to this behavior with a custom `failAction()` rather than it become the default `'log'` behavior.  Here's how that would look with this update:

```js
server.route({
    method: 'get',
    path: '/',
    handler: () => 'ok',
    options: {
        validate: {
            query: {
                a: Joi.string().min(2)
            },
            failAction: function (request, h, err) {
                request.log(['validation', 'error', err.payload.validation.source], err);
                throw err.data.defaultError;
            }
        }
    }
});
```

Resolves #4040

### DIFF
--- a/API.md
+++ b/API.md
@@ -3432,7 +3432,8 @@ Default value: `'error'` (return a Bad Request (400) error response).
 
 A [`failAction` value](#lifecycle-failAction) which determines how to handle failed validations.
 When set to a function, the `err` argument includes the type of validation error under
-`err.output.payload.validation.source`.
+`err.output.payload.validation.source`. The default error that would otherwise have been logged
+ or returned can be accessed under `err.data.defaultError`.
 
 #### <a name="route.options.validate.headers" /> `route.options.validate.headers`
 
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -146,7 +146,7 @@ internals.input = async function (source, request) {
     // Prepare error
 
     const defaultError = validationError.isBoom ? validationError : Boom.badRequest(`Invalid request ${source} input`);
-    const detailedError = Boom.boomify(validationError, { statusCode: 400, override: false });
+    const detailedError = Boom.boomify(validationError, { statusCode: 400, override: false, data: { defaultError } });
     detailedError.output.payload.validation = { source, keys: [] };
     if (validationError.details) {
         for (const details of validationError.details) {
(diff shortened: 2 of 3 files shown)


Nargonath (approved):
LGTM, good addition @devinivy . Thank you.

API.md
```diff
@@ -3432,7 +3432,8 @@ Default value: `'error'` (return a Bad Request (400) error response).
 
 A [`failAction` value](#lifecycle-failAction) which determines how to handle failed validations.
 When set to a function, the `err` argument includes the type of validation error under
-`err.output.payload.validation.source`.
+`err.output.payload.validation.source`. The default error that would otherwise have been logged
+ or returned can be accessed under `err.data.defaultError`.
```
Nargonath, line 3436: Just noticed, is it me or do we have a line return in the middle of the phrase?

Nargonath: Nevermind I checked the file rendering on GitHub and it displays properly. I might have been confused by the diff tool I guess.

PR #4351, opened 2022-04-26
Change default CORS preflight status code w/ configuration

### LINKED ISSUE #4165: Proposal to add route option for CORS preflight status code
#### Support plan
Community

#### Context

* *node version*: v12+
* *module version*: v19+
* *any other relevant information*: PR #4024 was originally opened to address the same underlying issue.

Users would like to be able to control the status code used to respond to CORS preflights.  Currently the only way to do this is to configure the global server setting `routes.response.emptyStatusCode`.  This has two drawbacks:
1. It is not necessarily desirable to couple the response code of preflights with the empty response code used by other endpoints.  One reason for this is that browser quirks [over time have shifted](https://github.com/Kong/kong/issues/6051#issue-644388401) what an ideal preflight response code should be (200 vs 204).  The client for preflights is the browser, and the client for other endpoints is likely the user's application.
2. It is not possible to configure this on a per-route basis.  The main downside here is that it is confusing that `routes.response.emptyStatusCode` configured on the server takes effect, but the analogous route-level setting `options.response.emptyStatusCode` does not.

In order to address both points above my recommendation is to introduce a new route option `cors.preflightStatusCode` to control the preflight status code, which is respected both by global server settings and on a per-route basis.  This would be a breaking change, as users are currently using the global `routes.response.emptyStatusCode` setting to control this, a workaround provided by @kanongil in #4024.  I propose the two be totally decoupled, however there are backwards-compatible ways to introduce this feature prior to decoupling the two options.

devinivy:
This work decouples the status code used for CORS preflight responses from the server's `emptyStatusCode`.  The CORS preflight response status code now defaults to `200` per conversation in #4165, and is configurable to `200` or `204` using `request.options.cors.preflightStatusCode`.

Resolves #4165

### DIFF
--- a/API.md
+++ b/API.md
@@ -2894,6 +2894,9 @@ object with the following options:
 - `credentials` - if `true`, allows user credentials to be sent
   ('Access-Control-Allow-Credentials'). Defaults to `false`.
 
+ - `preflightStatusCode` - the status code used for CORS preflight responses, either `200` or `204`.
+   Defaults to `200`.
+
 ### <a name="route.options.description" /> `route.options.description`
 
 Default value: none.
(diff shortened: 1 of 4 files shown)


Nargonath (approved):
LGTM. I followed the discussion in #4165 and I don't have any strong opinion on it. As you mentioned it seems there isn't any one-size-fits-all here but 200 seemed to be the most likely to go down without any trouble. If a user ever encounters a problem the option will be there for them to change the default status code so I think we're covered.

PR #4352, opened 2022-04-29
Change XSS header default to '0'

devinivy:
This adds support for configuring the `x-xss-protection` security header to `0`, and makes that the default in light of security issues that have been identified with XSS auditors/filters which are enabled by this header.  This is also OWASP's recommendation, and it's widely employed by sites such as Facebook, Google, and this one (GitHub!).

Resolves #4327

PR #4353, opened 2022-04-29
Fix flaky test in test/request.js

Nargonath:
The PR #4352 unveiled a flaky test on OSX for node@16 which relied on `await Hoek.wait(10)`. The timer was supposed to be enough for the socket to send along the data and the server to receive the request but on some OS it's not the case. With this fix we're waiting until the server gets the request or ~~die trying~~ fail on the overall timeout for individual test (current 5000ms). This way the test should be more reliable no matter the OS it's run upon.


devinivy (approved):
Looks good, thanks Jonas!

PR #4354, opened 2022-05-11
Update LICENSE.md

Yahkob:


Nargonath (commented):
Thank you for the suggestion. But I'm not sure we want to do it that way. Since Sideway is not part of the project anymore. We might want to split and add another line above. 

@devinivy did you plan on updating these with your changes for v21? As you did with other repos.

Nargonath:
I believe this is already done in the next major branches throughout the org. I'm going to close this for now. Thank you for the suggestion @Yahkob though. :)

PR #4355, opened 2022-05-30
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ci-module.yml
+++ b/.github/workflows/ci-module.yml
@@ -7,6 +7,11 @@ on:
   pull_request:
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   test:
+    permissions:
+      contents: none
     uses: hapijs/.github/.github/workflows/ci-module.yml@master


Nargonath:
Thank you for your suggestion @naveensrinivasan. I didn't have the time right now to go through all the links. We use a centralized template for the CI of all our modules so the changes won't happen here but you gave us the direction. Thank you. :)

Nargonath:
Looking at the different links you provided plus the documentation, I believe we still need `content: read` permissions for our test since we're cloning the repo to install dependencies and run the tests. We're also less at risk regarding write permissions in public fork as we do not use `pull_request_target` event for our actions but `pull_request`. Stating a GitHub blog:

> The GITHUB_TOKEN is an automatically generated secret that lets you make authenticated calls to the GitHub API in your workflow runs. Actions generates a new token for each job and expires the token when a job completes. The token has write permissions to a number of [API endpoints](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token) except in the case of pull requests from forks which are always read. These new settings allow you to follow a principle of least privilege in your workflows.

Source: https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/. 

I think setting:

```yaml
permissions:
  contents: read
```
at the top of the workflow template should be enough and we shouldn't put `permissions: contents: none` for the tests. I may be wrong so if that's the case please feel free to speak up.

I'll close this issue and handle it in another PR since we don't need the other modification you suggested.

Thank you for the report.

PR #4356, opened 2022-06-11
Set GH token permissions default to contents: read and none for the rest

Nargonath:
This relates to #4355. This PR is to test the settings to make sure they work and then we can apply them to `hapijs/.github` repo and roll them out to the rest of the org.


Nargonath:
Per [GH documentation](https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#permission-on-contents) I don't think we even need the GH token `contents: read` access. I revoked all permissions to the GH token and kept only the `meta: read` (default) access since for now our Actions don't seem to need more.

The CI are unrelated to this changes as I have them on master as well just by `npm i && npm t`.

PR #4357, opened 2022-06-13
Change default host to be IPv6-friendly

devinivy:
This changes hapi's default host of `0.0.0.0` to use node's IPv6-friendly default host of `::` or `0.0.0.0`, depending on whether the machine supports IPv6.

Under node v18 the drawbacks of the current default are emphasized, since DNS resolution in node has changed (https://github.com/nodejs/node/pull/39987) allowing `localhost` to resolve to the IPv6 address `::1`— since `0.0.0.0` was pinned to IPv4, a request to `localhost` from node would not be able to connect to a default hapi server.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -336,7 +336,7 @@ exports = module.exports = internals.Core = class {
                 this.listener.listen(this.settings.port, finalize);
             }
             else {
-                const address = this.settings.address || this.settings.host || '0.0.0.0';
+                const address = this.settings.address || this.settings.host || null;
                 this.listener.listen(this.settings.port, address, finalize);
             }
         });
--- a/lib/core.js
+++ b/lib/core.js
@@ -336,6 +336,7 @@ exports = module.exports = internals.Core = class {
                 this.listener.listen(this.settings.port, finalize);
             }
             else {
+                // Default is the unspecified address, :: if IPv6 is available or otherwise the IPv4 address 0.0.0.0
                 const address = this.settings.address || this.settings.host || null;
                 this.listener.listen(this.settings.port, address, finalize);
             }
--- a/test/request.js
+++ b/test/request.js
@@ -123,7 +123,7 @@ describe('Request.Generator', () => {
 
 describe('Request', () => {
 
-    it('sets client address', async () => {
+    it('sets client address (default)', async (flags) => {
 
         const server = Hapi.server();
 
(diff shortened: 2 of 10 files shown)


Nargonath (approved):
Thanks for the work Devin.

test/request.js
```diff
@@ -131,17 +129,19 @@ describe('Request', () => {
...
 
-            let expectedClientAddress = '127.0.0.1';
-            if (Net.isIPv6(server.listener.address().address)) {
-                expectedClientAddress = '::ffff:127.0.0.1';
+            // Call twice to reuse cached values
+
+            if (Common.hasIPv6) {
+                // ::ffff:127.0.0.1 on node v14 and v16, ::1 on node v18.
+                expect(request.info.remoteAddress).to.match(/^::ffff:127\.0\.0\.1|::1$/);
+                expect(request.info.remoteAddress).to.match(/^::ffff:127\.0\.0\.1|::1$/);
```
kanongil, line 137: Would it make sense to convert these to regular IPv4 adresses in hapi before assigning to `remoteAddress`? The property is not intended to signal the IP layer protocol, so I think it would be fine.

The mechanics of the address are from: https://datatracker.ietf.org/doc/html/rfc3493#section-3.7

lib/core.js
```diff
@@ -336,7 +336,7 @@ exports = module.exports = internals.Core = class {
                 this.listener.listen(this.settings.port, finalize);
             }
             else {
-                const address = this.settings.address || this.settings.host || '0.0.0.0';
+                const address = this.settings.address || this.settings.host || null;
```
devinivy, line 339: Reminder to self to add a comment here and update the docs.

test/request.js
devinivy: That seems reasonable.  It may be a little hard to test if there's a code branch relying on IPv6 networking to hit.  That aside for a moment, do you find this to be rigorous enough?
```js
const socketAddress = this._request.raw.req.socket.remoteAddress;
this._remoteAddress = socketAddress.startsWith('::ffff:') ?
    socketAddress.slice('::ffff:'.length) :
    socketAddress;
```

If after some back and forth this proves complex, I might lean towards continuing to pass through `socket.remoteAddress`.

devinivy: Thanks for the feedback, this is ready for another look.

kanongil (commented):
Looking good 👍

PR #4361, opened 2022-06-20
Upgrade production deps for hapi v21 and node v18

### LINKED ISSUE #4184: Inconsistent event emits

devinivy:
This updates all of hapi's production dependencies to the latest major versions, having dropped support for node v12 and added support for node v18.

Of note is one API change to hapi, where `server.events.emit()` may no longer be awaited due to changes in podium (though this is now supported through `await server.events.gauge()`).  The server `'stop'` and `'start'` events also no longer are awaited during server start and stop: instead use server extension points such as `onPreStart` and `onPostStop`.  Resolves #4184

If this looks good, we will publish this as hapi v21 beta 🎉 !  From there we will update test all hapi plugins and test them against the beta, then update the remaining hapi dev dependencies after the latest plugin versions are published 🎊


Nargonath (approved):

package.json
```diff
@@ -23,33 +23,33 @@
...
+    "@hapi/subtext": "^8.0.0",
+    "@hapi/teamwork": "^6.0.0",
+    "@hapi/topo": "^6.0.0",
+    "@hapi/validate": "^2.0.0"
   },
   "devDependencies": {
-    "@hapi/code": "9.0.0-beta.0",
-    "@hapi/eslint-plugin": "^5.0.0",
+    "@hapi/code": "^9.0.0",
+    "@hapi/eslint-plugin": "^6.0.0",
```
Nargonath, line 47: One thing though, I remember we set this dep to `'*'` with @Marsup in some plugins because we were having issues without it. Perhaps the underlying issue that was causing the behaviour has been fixed and I'm not up to date but I figured I might just voice it.

devinivy: In theory that makes perfect sense to me, but in practice this has been causing build failures.  Here's a failed run for hapi v20 that can be fixed by specifying the eslint-plugin version: https://github.com/hapijs/hapi/runs/6866679357?check_suite_focus=true.  These started failing once the next major of eslint-plugin was published.

Nargonath:
@devinivy I took the liberty to remove the source branch, feel free to restore if that's was a mistake.

PR #4372, opened 2022-08-24
Update documentation for server.events.on()/.once()

jonathansamines:


Nargonath (approved):
Good catch, thank you for the update. 👍

Nargonath:
Well CI is acting up but considering the changes you made there should be no problem since it's only documentation changes.

PR #4387, opened 2022-11-09
Add documentation for server.options.load.maxEventLoopUtilization

jonathansamines:
The `load.maxEventLoopUtilization` option was added in https://github.com/hapijs/heavy/pull/48, which was released as part of `@hapi/hapi` 21.x


devinivy (approved):
Hey, good call!

Issue #4388, opened 2022-11-11
Move types from DT locally


hueniverse:
Working with hapi in a TypeScript project for the first time, I'm finding the public types to be very lacking. Opening PRs against DT repos is painful and usually doesn't get reviewed. I would like to move the DT types into the repo as-is at first (so no breaking changes for those using the types), and then start patching problems as I find them.

I'll do the work if you are open to this approach.

cjihrig:
+1 from me!

Nargonath:
This is a good idea and would greatly benefit the project IMO, thanks for offering your help Eran. I believe somebody already started a similar work: https://github.com/hapijs/hapi/pull/4373. Perhaps there is something that could be of use for you. Would you need to handle the core dependencies that have not yet migrated away from DT first?

kanongil:
I agree with your points. A big challenge in a move, is how to handle typescript API stability and communicate it to users.

I expect a move will be followed by a desire to change it in ways that can break existing usage. Something that I expect we want to do without a major version bump.

It would perhaps be simpler if the typings could be published separately. I'm not sure, but it might be possible to create eg. a `@hapi/hapi-typings` package, add it as a dependency to `@hapi/hapi` and somehow make typescript recognise these typings. If that works, then we would be able to publish the typings with their own version number, making it possible for users to keep the existing compatible version, even if hapi is updated.

If the above can work, this can even be an opt-in thing, where hapi doesn't add it as a dependency but you need to add it yourself. At least until the API has stabilised a bit.

kanongil:
Actually, I think the simplest solution to separate the packages, is to to just re-export the external repo typings from a `lib/index.d.ts` file in hapi:

```ts
export * from '@hapi/hapi-typings';
```

I have done a quick test (you can even just re-export the `@types/hapi__hapi` types), and it seems to work quite nicely.

devinivy:
I believe we went down that path in hapi pal and I believe @damusix ran into some snags with it.  @damusix can you remind me what you found?

kanongil:
> I believe we went down that path in hapi pal and @damusix ran into some snags with it. @damusix can you remind me what you found?

I had a look, and see that you are talking in general terms, and not my specific approach.

What you seem to have attempted with the [`@hapipal/types`](https://github.com/hapipal/types) repo matches my initial approach, which I gave up on since it requires modifying your `tsconfig.json`.

[...]

PR #4373, opened 2022-08-26
WIP: feat: 🎸 typings

damusix:
- Added typings suggested from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61794
- migrated all type tests from DT onto here. 
- This commit is a WIP, given that `@hapi/catbox` has not yet been ported over to use its own typings.
- Feel free to leave suggestions and comments
- If you want to collaborate on this PR, send a message on the slack in the `#typescript` channel and I'll add you to my fork

PR #4392, opened 2022-11-14
External typings PoC

kanongil:
Initial implementation of my suggestion in https://github.com/hapijs/hapi/issues/4388#issuecomment-1312476374.

I have made no changes to the DT typings, which still target v20. I only changed the `@types/hapi__mimos` dependency to `@hapi/mimos`. The version label is copied from the current DT npm release.

While I have imported the DT test suite, I haven't done anything to run it. I have only implemented a very barebones proof-of-concept `test/index.ts` test.

FYI, I haven't tested how this actually works once published, and npm-installed into a project.

Note that I have used the npm *workspaces* feature to improve the ergonomics. This means that the `@hapi/hapi-typings` module lives inside the main hapi repo without any fuss, making it possible to add typings in a single PR along with new hapi features. The main challenge will probably be to remember to actually publish the updated typings with hapi releases, but this should be fixable with a publish script.


kanongil:
The node 14 test failures will be because it uses an older version of npm that doesn't support workspaces. This is a dev-only issue and should be fixable by installing an update version.

kanongil:
If we want to continue with this, it probably makes sense to merge it with the work done in #4373.

devinivy:
Thanks for the rigorous demo!  This was useful in hashing out #4388.  For those who didn't follow along on that issue, we've decide to move forward for now with typings living within each module.

PR #4396, opened 2022-11-16
Fix handling of no remoteAddress available

### LINKED ISSUE #4395: Error from remoteAddr if the socket goes away
#### Support plan

* *is this issue affecting a production system?* (yes/no): yes

#### Context

* *node version*: 14.21.1
* *module version with issue*: 21.0.0
* *last module version without issue*: 20.2.2 
* *environment* (e.g. node, browser, native): Node
* *used with* (e.g. hapi application, another framework, standalone, ...): Hapi application

Accessing `request.info.remoteAddress` may fail if the remote socket has gone away.

This appears to be the result of #4357.

```
TypeError: Cannot read property 'startsWith' of undefined
    at internals.Info.get remoteAddress [as remoteAddress] (/home/deploy/node_modules/@hapi/hapi/lib/request.js:661:31)
```

A successful `undefined` value

joshkel:
Fixes #4395

I'm unfamiliar with Hapi's test designs; please let me know if the test needs any changes.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -658,7 +658,7 @@ internals.Info = class {
         if (!this._remoteAddress) {
             const ipv6Prefix = '::ffff:';
             const socketAddress = this._request.raw.req.socket.remoteAddress;
-            if (socketAddress.startsWith(ipv6Prefix) && socketAddress.includes('.', ipv6Prefix.length)) {
+            if (socketAddress && socketAddress.startsWith(ipv6Prefix) && socketAddress.includes('.', ipv6Prefix.length)) {
                 // Normalize IPv4-mapped IPv6 address, e.g. ::ffff:127.0.0.1 -> 127.0.0.1
                 this._remoteAddress = socketAddress.slice(ipv6Prefix.length);
             }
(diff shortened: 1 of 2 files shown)


DJMcK (commented):
Thanks for working on this, @joshkel! Experiencing the same exception accessing `request.info.remoteAddress` which brought me here.

Not a core maintainer so take my suggestions with a pinch of salt :)

---

FWIW, I'm not entirely convinced that there is a need to go through the whole rigmarole of creating a connection and destroying it for this case.

We could just simplify things down to something along the lines of:

```js
it('handles when client address is undefined', async () => {

    const server = Hapi.server();
    server.route({
        method: 'get',
        path: '/',
        handler(request) {

            delete request.raw.req.socket.remoteAddress;
            return request.info.remoteAddress === undefined;
        }
    });
    const { result } = await server.inject('/');
    expect(result).to.equal(true);
});
```

I believe `delete` should be safe to use in this scenario.

devinivy:
This looks pretty dang close— I should be able to land this within the next 24 hours.

ssnigur:
@devinivy any updates on when will this be merged? Thank you

PR #4398, opened 2022-11-18
Move TypeScript definition from DT (as-is)

hueniverse:
This adds the DT type definition AS-IS without any changes as well as creates a basic type tests setup. After spending a day trying to migrate the DT tests it was clear that they tests were too broken to reuse. The plan is to add more tests as we improve the types.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2509,7 +2509,8 @@ async function example() {
 
     const server = Hapi.server({ port: 80 });
     await server.start();
-    }
+    console.log('Server started at: ' + server.info.uri);
+}
 ```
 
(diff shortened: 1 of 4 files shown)


Marsup (commented):
LGTM overall. Just weird to have to rename the type assertion checker, wondering if we should merge those.

Too bad for their tests, what's so broken about it? (just to keep track)

kanongil:
💯 on board moving Hapi to TS. Though I would probably recommend staying with the commonjs codegen, and leave the dependencies as-is.

PR #4399, opened 2022-11-19
Split Types file

hueniverse:
Follows #4398 

* Adds missing `info.remote` server option.
* Splits the definition file into smaller parts (some still too big but it's a start). No other changes - just a lot of copy/paste.
* Add `server.mime' type.


damusix:
lgtm

Marsup:
Do you want it released? I don't think that's necessary for this one, it can be buffered til next release.

PR #4407, opened 2022-12-17
Add type for preflight status code in cors options

guoard:
Add type for https://github.com/hapijs/hapi/pull/4351 PR.


devinivy:
Thanks!

PR #4419, opened 2023-01-23
Fix mimos import in lib/types/server/server.d.ts

### LINKED ISSUE #4418: [definition files] TypeScript 5.0 errors for @hapi/* imports after 21.2.0 update
On DefinitelyTyped ecosystem the error has been seen for package using @hapi/hapi types (via proper versioning `^21.1.0`, that matched recent `21.2.0` update):
> Error: node_modules/@hapi/hapi/lib/types/server/server.d.ts(5,10): error TS2617: 'Mimos' can only be imported by using 'import Mimos = require("@hapi/mimos")' or by turning on the 'esModuleInterop' flag and using a default import.

`@hapi/mimos` was not updated for months. The last change related looks to be 21.2.0 release with changes to symbol exports to `@hapi/hapi` module:
https://github.com/hapijs/hapi/pull/4399/files#diff-cfc7807f6b174e1c3705c5c2747f56e49a889a52999dc96e4080a3af0131a534

Can you folks review this? I can help, if required. thanks!

sandersn:
@hapi/mimos still uses TS-only `export = Mimos` syntax, so server.d.ts needs to use TS-only `import Mimos = require('@hapi/mimos')` syntax.

A full fix would change the exports of @hapi/mimos, but this PR is a lot smaller.

Fixes #4418

### DIFF
--- a/lib/types/server/server.d.ts
+++ b/lib/types/server/server.d.ts
@@ -2,7 +2,7 @@ import * as http from 'http';
 import * as zlib from 'zlib';
 
 import { Root } from 'joi';
-import { Mimos } from '@hapi/mimos';
+import Mimos = require('@hapi/mimos');
 
 import {
     Dependencies,


Marsup:
It doesn't https://github.com/hapijs/mimos/blob/master/lib/index.d.ts, but DT types do. I'd say the issue is you still have those installed.

PR #4429, opened 2023-03-02
feat: add diagnostic channel for server, route, request creation and response sent

YoannMa:
Reimplementation of what @AdriVanHoudt made on the previous PR #4278 

What changed since ?
 - Node 12 is no longer supported, there is no longer the issue of requiring a polyfill.
 - The API `diagnostics_channel` is stable since Node 16, The only part of the API we use, is identical in Node 14 but marked as experimental. Node 14 being EOL in less than 2 months, I don't really see a big issue there.

Following what Fastify does on their side https://github.com/fastify/fastify-diagnostics-channel, I added 3 more channels for the `route`, `request` and `response`.

I also added documentation of the exposed channels.

Questions : 
 - Should we add more channels (`onTimeout`,`onError`) ?
 - Naming convention of the channel ? 
		I copied Fastify naming scheme, but I don't really have any strong opinions regarding the naming scheme we pick : `hapi:server`, `hapi.onServer` or `hapi.server` 

Cc'ing @Qard and @vdeturckheim as they are probably interested in the outcome of this PR.


test/channels.js
```diff
@@ -0,0 +1,128 @@
...
+
+        const channel = DC.channel('hapi.onRequest');
+
+        it('request should be exposed on creation through the channel hapi.onRequest', async () => {
+
+            const server = Hapi.server();
+
+            server.route({ method: 'GET', path: '/', handler: () => 'ok' });
+
+            let requestExposed;
```
cjihrig, line 110:
```suggestion
            const server = Hapi.server();
            let requestExposed;

            server.route({ method: 'GET', path: '/', handler: () => 'ok' });
```

test/channels.js
```diff
@@ -0,0 +1,128 @@
...
+
+            const eventPromise = new Promise((resolve) => {
+
+                channel.subscribe((req) => {
+
+                    requestExposed = req;
+
+                    resolve();
```
cjihrig, line 118:
```suggestion
                    requestExposed = req;
                    resolve();
```

test/channels.js
```diff
@@ -0,0 +1,128 @@
...
+                });
+            });
+
+            const response = await server.inject('/');
+            await eventPromise;
+
+            expect(response.request).to.equal(requestExposed);
```
cjihrig, line 125:
```suggestion
            await eventPromise;
            expect(response.request).to.equal(requestExposed);
```

API.md
```diff
@@ -5030,3 +5030,26 @@ const plugin = {
     }
 };
 ```
+
+## Diagnostic Channels
+
+see : https://nodejs.org/docs/latest-v16.x/api/diagnostics_channel.html
```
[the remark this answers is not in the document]
YoannMa: Fixed the URL, added a bit of explaining and examples, do you feel like it's enough ?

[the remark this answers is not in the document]
cjihrig: I'm good with it. Let's give other folks a chance to weigh in.

Qard:
It's been awhile since I've used Hapi so I'll have to take another look at it when I have time to know for sure if this satisfies what APMs need. One thing that _does_ seem to be missing is we'd like an event to track errors.

@rochdev Any other suggestions for events that could help us?

rochdev:
@Qard As you mentioned, definitely an event for errors. I would say also an event for when the route is _entered_ as opposed to just _created_, unless the route is already available on the request object by the time it's emitted. Events for when plugins/extensions are entered/exited would also be useful, although this could be out of scope of this PR.

YoannMa:
Route is available on the request object (see https://github.com/hapijs/hapi/blob/master/API.md#request.route) so I believe the hook on route _entered_ is not useful.

Qard:
I should also say, channel naming doesn't really matter as long as the chosen name is documented, and it includes the module name somehow to avoid possible collisions with other modules as diagnostics_channel is a global namespace.

It's also worth mentioning that I have an in-progress API to make the tracing use case easier with diagnostics_channel via https://github.com/nodejs/node/pull/44943. While the API itself will not be available in all the supported versions of hapi for awhile yet, it _does_ document the behaviour as a pattern which can be implemented manually with channels _now_, so long as you follow the guidelines of how it should behave. It could be worth a look to understand what things we capture and how that could be translated to channels introduced for hapi.

test/channels.js
```diff
@@ -0,0 +1,121 @@
...
+                channel.subscribe((srv) => {
+
+                    exposedServer = srv;
+                    resolve();
+                });
+
+                server = Hapi.server();
+            });
+
+            expect(exposedServer).to.equal(server);
```
kanongil, line 35: This needs a `.shallow` to not do a default deep comparison.

test/channels.js
```diff
@@ -0,0 +1,121 @@
...
+
+            let exposedServer;
+            let server;
+
+            await new Promise((resolve) => {
+
+                channel.subscribe((srv) => {
+
+                    exposedServer = srv;
+                    resolve();
```
kanongil, line 29: It is simpler and more clear to resolve with the `server` and assign it the the awaited result of the promise.

```js
             const exposedServer = await new Promise((resolve) => {

                 channel.subscribe(resolve);
             });

             const server = Hapi.server();

             expect(exposedServer).to.shallow.equal(server);
```

test/channels.js
```diff
@@ -0,0 +1,121 @@
...
+const Lab = require('@hapi/lab');
+
+const Hapi = require('..');
+
+const { describe, it } = exports.lab = Lab.script();
+const expect = Code.expect;
+
+describe('DiagnosticChannel', () => {
+
+    describe('onServerChannel', () => {
```
kanongil, line 15: It seems clearer to name it after the exposed channel name, eg. `hapi.onServer`.

test/channels.js
```diff
@@ -0,0 +1,121 @@
...
+
+                    responseExposed = res;
+                    resolve();
+                });
+            });
+
+            const response = await server.inject('/');
+            await eventPromise;
+
+            expect(response.request.response).to.equal(responseExposed);
```
kanongil, line 92: Missing `.shallow`

test/channels.js
```diff
@@ -0,0 +1,121 @@
...
+                channel.subscribe((req) => {
+
+                    requestExposed = req;
+                    resolve();
+                });
+            });
+
+            const response = await server.inject('/');
+            await eventPromise;
+            expect(response.request).to.equal(requestExposed);
```
kanongil, line 118: Maybe add a check to verify that it has not been routed when it is triggered? Eg.

```js
             const event = new Promise((resolve) => {

                 channel.subscribe(resolve);
             });

             const req = server.inject('/');
             const request = await event;
             expect(request).to.exist();
             expect(request.route).to.not.exist();

             const response = await req;
             expect(response.request).to.shallow.equal(request);
```

API.md
```diff
@@ -5030,3 +5030,68 @@ const plugin = {
     }
 };
 ```
+
+## Diagnostic Channels
+
+Diagnostic Channels allows Hapi to report events to other modules. This is useful for APM vendors to collect data from Hapi for diagnostics purposes.
```
kanongil, line 5036: Maybe more explicitly state that this is not intended for use inside your own server / plugin, where there are are more suitable hooks?

API.md
```diff
@@ -5030,3 +5030,68 @@ const plugin = {
...
+### `hapi.onServer`
+
+This event is sent when a new server is created. The [`server`](#server) object is passed as the message.
+
+```js
+const DC = require('diagnostics_channel');
+const channel = DC.channel('hapi.onServer');
+
+channel.subscribe((server) => {
+    // Do something with the server
```
kanongil, line 5049: Needs an empty line before and after to match hapi style.

API.md
```diff
@@ -5030,3 +5030,68 @@ const plugin = {
...
+    console.log(server.version);
+});
```
kanongil, line 5051: Hapi itself doesn't actually use any deprecated API, only these examples and the tests. I'm not a fan of using deprecated API to illustrate how to use, but since this is targeted at APM vendors they should known what they are doing anyway.

Maybe just have a single example at the top?

kanongil (commented):
Looks reasonable.

My main concern is that a `throw` in a registered channel handler can propagate to hapi. It seems that [the current version](https://github.com/nodejs/node/blob/1ccc8cc858b94b97b6c3c76140efe3e36e57aa06/lib/diagnostics_channel.js#L51-L61) guards against this, but a test to verify would be nice.

kanongil:
> Route is available on the request object (see https://github.com/hapijs/hapi/blob/master/API.md#request.route) so I believe the hook on route _entered_ is not necessary.

`route` won't be available yet when `onRequest` is triggered. The earliest it can be determined using public API, is with a `onPreAuth` hook, and it would have to contend with other hooks and not trigger when cookie processing fails.

kanongil:
Regarding channels, I would advocate on limiting it until there is a concrete need. Eg. with just the `onServer` channel, APM's should be able to hook something that matches `hapi.onRoute`, `hapi.onRequest`, and `hapi.onResponse`.

Exposing just a `hapi.onServer` channel would also mean that there is zero runtime overhead when channels are not used.

Qard:
> My main concern is that a `throw` in a registered channel handler can propagate to hapi. It seems that [the current version](https://github.com/nodejs/node/blob/1ccc8cc858b94b97b6c3c76140efe3e36e57aa06/lib/diagnostics_channel.js#L51-L61) guards against this, but a test to verify would be nice.

The diagnostics_channel has guarded against this since the beginning. It's not possible for a throw in a subscriber to ever impact the publisher.

> Exposing just a `hapi.onServer` channel would also mean that there is zero runtime overhead when channels are not used.

The diagnostics_channel API is specifically designed to have zero overhead when a channel is not used. The publish function is a no-op until the channel has subscribers, and if it ever returns to zero subscribers it becomes a no-op again.

test/channels.js
YoannMa: From what I found, `request` are created with the default `notFound` route :

https://github.com/hapijs/hapi/blob/c6d881ff10e3e102c2af99f117901740ee048d7d/lib/request.js#L57

Should I test that instead ?

kanongil: This sounds like a minor hapi bug – the docs state that [it is unassigned](https://github.com/hapijs/hapi/blob/c6d881ff10e3e102c2af99f117901740ee048d7d/API.md?plain=1#L3622) when `onRequest` is processed. I have prepared [a PR to fix this](#4433).

In the meantime you can test for `request.params` to equal `null` instead.

kanongil (commented):
Looking at it again I think it would work better if the channels are triggered before the equivalent user-facing hook. This way the event handlers cannot modify the object before the APM sees it, making it a more "pure" signal.

lib/core.js
```diff
@@ -510,6 +511,8 @@ exports = module.exports = internals.Core = class {
 
             const request = Request.generate(this.root, req, res, options);
 
+            Channels.onRequestChannel.publish(request);
```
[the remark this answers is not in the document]
YoannMa: If this assignment is in the `request._execute()`, perhaps it is to avoid executing it in case of heavy load ?

I would vote to keep the `publish()` here given we need another event when a request is routed. I could add an event early in the `request._lifecycle()` method ? Something like `hapi.onRequestLifecycle` ?

YoannMa:
Added both `hapi.onRequestLifecycle` and `hapi.onError` channels.

I'm not really sure about the name `hapi.onRequestLifecycle`, if you have some ideas for another name, I'm listening.

API.md
YoannMa: I don't mind doing another PR when Hapi drops support for Node 14 to replace those deprecated usage

kanongil:
I would much prefer that this PR does not add any more channel hooks. Ideally each hook gets a PR to allow simpler reviews. Eg. i have comments regarding `hapi.onError`, but I don't want to complicate the review further.

Personally, I would just include the bare minimum (eg. `hapi.onServer`) for an initial PR. As stated in https://github.com/hapijs/hapi/pull/4429#issuecomment-1463799340, it should be enough to get started. For instance, looking at the `elastic-apm-node` APM it is [the only hook it will need](https://github.com/elastic/apm-agent-nodejs/blob/37517521366e8250f6d8401b6e470b7104432b21/lib/instrumentation/modules/hapi.js#L29-L34). The remaining hooks are handled [using the public API](https://github.com/elastic/apm-agent-nodejs/blob/37517521366e8250f6d8401b6e470b7104432b21/lib/instrumentation/modules/hapi.js#L45-L94).

damusix:
@kanongil @YoannMa What's the verdict on this PR? Do we want to want to add it for v22? Is it complete?

PR #4433, opened 2023-03-12
Make pre-routed request.route property null

kanongil:
This fixes the value of `request.route` during `onRequest` processing to closer match the docs. As it is, it just returns the `notFound` route. I updated the docs to clarify that the value will be `null`.

The PR ended up being quite extensive, since the internal logic uses the `route` property when it has not been routed. I fixed this by using the internal `_route` instead.

This PR also contains a fix for a minor issue with `response.escape(true)`, which would be ignored.


kanongil:
Note that this is a bug fix to clarify the public API and make the code match this. It might however cause problems for code that uses `onRequest` extensions that inadvertently relies on the `request.route` value being set.

devinivy:
For that reason I wonder if we should save this for the next major.  I figure, if folks do currently have this use-case then it will cause a breakage for them, if if they don't have this use-case then they are not experiencing an issue.  I suppose the upside here would be that this helps folks who think they have working code but actually do not, due to the ambiguous documentation.

Hmm!  What do others think?

PR #4435, opened 2023-03-16
Explicitly add dependency @hapi/iron

hraban:
Fixes typescript compilation for downstream projects where npm can't flatten this nested dependency into top. The only reason this appears to work here is because it is transitively included by statehood.

Required here:

https://github.com/hapijs/hapi/blob/c6d881ff10e3e102c2af99f117901740ee048d7d/lib/types/server/state.d.ts#L2

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "@hapi/catbox-memory": "^6.0.1",
     "@hapi/heavy": "^8.0.1",
     "@hapi/hoek": "^11.0.2",
+    "@hapi/iron": "^7.0.1",
     "@hapi/mimos": "^7.0.1",
     "@hapi/podium": "^5.0.1",
     "@hapi/shot": "^6.0.1",


hraban:
> Thanks for the PR.
> 
> Just for reference, `@hapi/iron` is already an implicit dependency of hapi through the `@hapi/statehood` dependency.
> 
> While it doesn't change the installed code, I'm not a fan of adding dependencies just to satisfy typescript. Eg. we risk it going out of sync. Maybe there is a better way to handle this?

I guess you could call this a devDependency? Is that transitive? I.e. when installing dev dependencies, does it also install all those dependencies' devDependencies?

Otherwise, the entire point of this .d.ts is moot and it would (absurdly) better be off removed entirely or replaced with some strategic `any`s, because at least then it wouldn't break builds. Of course that's not ideal, but hopefully it highlights the problems with the current situation.

Marsup:
Oops, my PR wasn't supposed to close this one 🤔 Could you test with 21.3.2 and report if it's still an issue?

kanongil:
Statehood [re-exports the iron types](https://github.com/hapijs/statehood/blob/364f7dfefd568288f05135af82f4446ed8db9c21/lib/index.d.ts#L4), so a better fix is to just import those instead.

hraban:
I'm not involved with the code that uses this anymore so I can't test anything nor give feedback I'm afraid. Godspeed!

PR #4457, opened 2023-08-28
Update LICENSE.md

faridSopian:
Make it more proffesional to read


Nargonath (commented):
Thank you for the suggestion but I assume we're going to keep it the way it is right now.

PR #4458, opened 2023-09-13
remove cjihrig from TSC

cjihrig:
I have not had the time to contribute as much as I would like and no longer use Node in my day job, so it's probably best to remove myself.


Nargonath (approved):
It's sad to you go but I understand.

PR #4466, opened 2023-10-15
Change `ServerRoute`'s `handler` type to & `HandlerDecorations`

hhanesand:
Seems like this type definition should use & and not | if the intent is to merge properties from HandlerDecorations into the lifecycle handler.


Marsup:
I don't think it's the same thing, the handler can either be a method or an object matching a decoration (such as [here](https://hapi.dev/tutorials/servingfiles/?lang=en_US#filehandler)), it can't be both.

Issue #4363, opened 2022-06-28
Server crash: Error: Unknown error


sveisvei:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*: v16.15.0
* *module version with issue*:  20.2.2
* *last module version without issue*: ? 
* *environment* (e.g. node, browser, native): node
* *used with* hapi application
* *any other relevant information*:

#### What are you trying to achieve or the steps to reproduce?

So playing with some invalid headers, I found a bug.

Test case hapi server
```js
"use strict";

const Hapi = require("@hapi/hapi");

const init = async () => {
  const server = Hapi.server({
    port: 8080,
    host: "localhost",
  });

  server.route({
    method: "*",
    path: "/{p*}",
    options: {
      payload: {
        parse: true, // turning this off, will remove the issue
      },
    },
    handler: (req, h) => {
      console.log("handler");
      return h.response("fail").code(404);
    },
  });

  await server.start();
  console.log("Server running on %s", server.info.uri);
};

process.on("unhandledRejection", (err) => {
  console.log(err);
  process.exit(1);
});

init();
```

command to reproduce the crash:
```bash
nc 127.0.0.1 8080 < body.txt
```

body.txt
```txt
POST / HTTP/1.1
Host: localhost
Content-Length: 5
Transfer Encoding:chunked

1
Z\n\Q\n\n
```

#### What was the result you got?
```
Error: Unknown error
    at new module.exports (./node_modules/.pnpm/@hapi+hoek@9.2.1/node_modules/@hapi/hoek/lib/error.js:23:19)
    at Object.module.exports [as assert] (./node_modules/.pnpm/@hapi+hoek@9.2.1/node_modules/@hapi/hoek/lib/assert.js:20:11)
    at Response._marshal (./node_modules/.pnpm/@hapi+hapi@20.2.2/node_modules/@hapi/hapi/lib/response.js:566:14)
    at exports.content (./node_modules/.pnpm/@hapi+hapi@20.2.2/node_modules/@hapi/hapi/lib/headers.js:41:24)
    at Object.internals.marshal (./node_modules/.pnpm/@hapi+hapi@20.2.2/node_modules/@hapi/hapi/lib/transmit.js:41:15)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
[...]

PR #4471, opened 2023-11-07
Don't return error response when clientError is a pipelined HPE_INVALID_METHOD

kanongil:
This PR fixes the specific issue in #4363.

This is done by specialcasing `clientError` errors with code `HPE_INVALID_METHOD` that have an active request. Since the active request definitely has a valid method (otherwise it would not have passed through the router), this error originates from request data that is supposed to relate to the next request. As such, it should not respond with an error, but with the normal response. A hook is installed to make the connection close once this has been delivered.

Since the existing test suite only exercises this with `err.code === 'HPE_INVALID_METHOD'`, I have added an extra test to exercise it with `err.code === 'HPE_INVALID_CHUNK_SIZE'` to keep the code coverage at 100%.

I also made a minor cleanup to `request._finalize()` and `request. _setResponse()`. `_finalize()` seems to somewhat assume that any non-Boom `this.response` object is a `Response`, when it can also be a `Symbol`.

Note that ideally, the current request should add a `connection: close` header, to indicate that it will close the connection.


kanongil:
Side issue: The test suite doesn't actually test on node 18, *even though it is the only fully supported node release!*

I don't know why the macOS tests fail. They fine work locally.

Marsup:
Is it ok to chain end and destroy on a stream like that? I would expect destroy to be hooked to an event or something.

kanongil:
Right you are @Marsup. I don't know why I wanted the extra destroys… They are gone now.

Marsup (approved):

PR #4473, opened 2023-11-15
Node 20 fixes

kanongil:
This fixes the test suite for node v18, v20 and v21 on my mac.

### DIFF
--- a/lib/transmit.js
+++ b/lib/transmit.js
@@ -284,9 +284,7 @@ internals.end = function (env, event, err) {
     env.team = null;
 
     if (request.raw.res.writableEnded) {
-        if (!event) {
-            request.info.responded = Date.now();
-        }
+        request.info.responded = Date.now();
 
         team.attend();
         return;
--- a/test/transmit.js
+++ b/test/transmit.js
@@ -428,7 +428,7 @@ describe('transmission', () => {
             const log = server.events.once('response');
             const client = Net.connect(server.info.port, () => {
 
-                client.write('GET / HTTP/1.1\r\naccept-encoding: gzip\r\n\r\n');
+                client.write('GET / HTTP/1.1\r\nHost: host\r\naccept-encoding: gzip\r\n\r\n');
             });
 
             const [request] = await log;
--- a/.labrc.js
+++ b/.labrc.js
@@ -1,0 +1,9 @@
+'use strict';
+
+const Somever = require('@hapi/somever');
+
+module.exports = {
+    'coverage-predicates': {
+        allowsStoppedReq: Somever.match(process.version, '<=18'),
+    }
+};
--- a/lib/core.js
+++ b/lib/core.js
@@ -506,6 +506,12 @@ exports = module.exports = internals.Core = class {
 
         return (req, res) => {
 
+            // $lab:coverage:off$ $not:allowsStoppedReq$
+            if (this.phase === 'stopping') {
+                return req.destroy();
+            }
+            // $lab:coverage:on$ $not:allowsStoppedReq$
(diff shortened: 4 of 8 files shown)


kanongil:
Hmm, strange that the `macOS` runners fail, when they work locally (on M1 macOS 14.1.1).

Looking at the failing tests, it seems that `request.active()` never returns `false`.
The "*Request  active() exits handler early when request is no longer active*" test case seems central.

The "*Request _postCycle() skips onPreResponse when validation terminates request*" test does not directly use `request.active()` but it might still rely on the internal version of this.

I don't have a macOS 12 machine to test on, so I'm not really able to debug further.

Even stranger, the tests [passes on node 18](https://github.com/hapijs/hapi/actions/runs/6800372695/job/18488761001?pr=4472), but [fails again on node 20](https://github.com/hapijs/hapi/actions/runs/6800372695/job/18488760694?pr=4472#step:5:39).

Nargonath:
Looks like we have some flakiness at play here. I restarted the 3 failed macos jobs and now macos node@14 passes whereas it didn't before. This might explain why it works on your machine @kanongil but not in the CI.

lib/transmit.js
```diff
@@ -284,9 +284,7 @@ internals.end = function (env, event, err) {
     env.team = null;
 
     if (request.raw.res.writableEnded) {
-        if (!event) {
-            request.info.responded = Date.now();
-        }
+        request.info.responded = Date.now();
```
Nargonath, line 287: Why did you remove the `!event` check?

test/transmit.js
```diff
@@ -428,7 +428,7 @@ describe('transmission', () => {
             const log = server.events.once('response');
             const client = Net.connect(server.info.port, () => {
 
-                client.write('GET / HTTP/1.1\r\naccept-encoding: gzip\r\n\r\n');
+                client.write('GET / HTTP/1.1\r\nHost: host\r\naccept-encoding: gzip\r\n\r\n');
```
Nargonath, line 431: Why did you add the `Host` header?

Nargonath (approved):
code LGTM, I had some questions for my own curiosity.

lib/transmit.js
kanongil: Because the early `req.destroy()` call removes the only test case that has a non-`null` event, making it fail coverage. I don't think there are any other ways to end up with `res.writableEnded` along with an event, and if there is such a corner case, the damage will be negligible.

test/transmit.js
kanongil: Because it is required to be a valid HTTP/1.1 request, and the node v20+ parser is more pedantic, no longer generating requests from this.

kanongil:
> Looks like we have some flakiness at play here. I restarted the 3 failed macos jobs and now macos node@14 passes whereas it didn't before. This might explain why it works on your machine @kanongil but not in the CI.

Ah, it seems that it only fails on runners using macOS 12.6.9, while runners on macOS 12.7 works. I guess not all `macos-latest` runners are identical…

I have been doing some experiments, and the tests seem to pass when using the `macos-11` or `macos-13` (beta) runners, so that might be a way to fix this for now.

lib/transmit.js
Nargonath: Alright, thanks for the explanation.

test/transmit.js
Nargonath: Ok, thanks.

Nargonath:
Thanks for the investigation on the CI. We probably shouldn't block this PR for this and can handle it in another PR.

PR #4474, opened 2023-11-17
Remove support for pre v16 joi

kanongil:
The last joi@15.1.1 was release more than 4 years ago. It should be safe to drop (in a breaking release).

This also removes several npm deprecation notices when installing dev dependencies.

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -126,7 +126,7 @@ internals.input = async function (source, request) {
         const schema = request.route.settings.validate[source];
         const bind = request.route.settings.bind;
 
-        var value = await (typeof schema !== 'function' ? internals.validate(request[source], schema, localOptions) : schema.call(bind, request[source], localOptions));
+        var value = await (typeof schema !== 'function' ? schema.validateAsync(request[source], localOptions) : schema.call(bind, request[source], localOptions));
         return;
     }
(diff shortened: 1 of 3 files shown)


Marsup (approved):

jonathansamines (approved):

damusix:
@kanongil Is this still a valid PR? Any chance you can update it if need be? We're converging for a major release and may include this if it's ready.

kanongil:
There might be an issue with this PR, let me check…

kanongil (commented):
We need to keep support for non-async validators, since an external validator might only use this.

Ie. I have removed the logic changes, and this patch just replaces the legacy joi test, with a custom non-async validator test.

The test failures seem unrelated to this PR. Note that node@14 also a fails while reporting a false positive.

kanongil:
Note that since this no longer changes the logic, this patch could also be applied against the main branch.

damusix:
@kanongil Can you rebase and fix the lint errors on this? 

Screenshot 2025-01-19 at 1 53 29 AM

kanongil:
No – It is already based of the tip of `next`. I have made a [new PR against master](https://github.com/hapijs/hapi/pull/4539), though.

PR #4489, opened 2024-03-15
fix: 🐛 loosen pre typings to allow for better programability

damusix:


damusix:
@kanongil @Marsup we don't like this approach right? Pre typing has def been painful.

Marsup:
IIRC this approach makes no compatibility check between pres and the route, and gives you `any` for inline pres, so I don't think it's the right move.

PR #4494, opened 2024-03-21
Assert on unsupported route method

kanongil:
Use the node.js built-in parseable http method list to reject routes that can never be reached.

Given that this will make hapi fail to start on existing code that inadvertently uses an unsupported method (maybe a typo), this might be more suited for including in a breaking release.

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const Assert = require('assert');
+const Http = require('http');
 
 const Bounce = require('@hapi/bounce');
 const Catbox = require('@hapi/catbox');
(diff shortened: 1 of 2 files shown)


Marsup:
So this would only fail for people using verbs that are outside of any spec? That doesn't look like a breaking change in my book. Could that be a problem for people using unix sockets maybe?

kanongil:
> So this would only fail for people using verbs that are outside of any http spec? That doesn't look like a breaking change in my book. Could that be a problem for people using unix sockets maybe?

Not exactly. The HTTP spec allows any combination of letters with case sensitivity. However, the *node http parser* only supports a limited subset (as listed in [Http.METHODS](https://nodejs.org/api/http.html#httpmethods)), and any outside this will error. As such, any routes with these methods will be unreachable.

kanongil:
Whether unix sockets are used is irrelevant, as this relates to the content of the stream, not how it is made.

Issue #4414, opened 2023-01-14
accept ZOD validator


Radik24:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*:  16.16.0
* *module version*: 21.1.0
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi application
* *any other relevant information*:

#### How can we help?

Basically I would like to replace Joi with ZOD (better with typescript, type inference). I can set my own validation on each route, validate payload/params/query and return error/pass. But Hapi always adds own `validation` object in returned error, I would say it expects Joi errror and is written to parse it. It would be nice to accept ZOD on route validation.

Just set?:
`server.validator(z);`

and use

```
{
  method: 'POST',
  path: `/validate`,
  handler: (req, h) => h.response().code(200),
  options: {
    validate: {
      payload: z.object({
        a: z.string(),
        b: z.string()
      })
    }
  }
}
```

instead of:

```
{
  method: 'POST',
  path: `/validate`,
  handler: (req, h) => h.response().code(200),
  options: {
    validate: {
      payload: Joi.object({
        a: Joi.string().required(),
        b: Joi.string().required()
      })
    }
  }
}
```

I can create validators and infere types in separate files, and use it as needed.

Thanks

devinivy:
It should just be a handful of lines of code to get basic zod support working.  Here's a workingz example: https://runkit.com/devinivy/63c4627a352d980008baa41b.  The core idea is this:
```js
const zodValidator = {
    compile: (schema) => ({
        validate: (val) => schema.parse(val)
    })
};

server.validator(zodValidator);
```

Radik24:
Is it something that the hapi team considers to be the right way?

cjihrig:
@Radik24, @devinivy is part of the hapi team 😄

Radik24:
I know, I just want to know if it's kind of hack or it could be implemented natively in future :)

cjihrig:
I'm not a zod user, but the solution looks legit to me. I'll let Devin comment further.

devinivy:
Yep, you should be good to go!  Any change to this validator interface would be considered a breaking change to hapi, i.e. it's considered part of our API.  If anyone is interested in calling this out more explicitly in the docs, would happily take a PR!

Radik24:
[...]

PR #4497, opened 2024-04-09
Remove Joi from types

### LINKED ISSUE #4491: Build issues
### Runtime

nodejs

### Runtime version

v20.10.0

### Module version

v20.10.0

_No response_

### Used with

tsoa

> tsc

node_modules/@hapi/hapi/lib/types/route.d.ts:2:68 - error TS2307: Cannot find module 'joi' or its corresponding type declarations.

2 import { ObjectSchema, ValidationOptions, SchemaMap, Schema } from 'joi';
                                                                     ~~~~~

node_modules/@hapi/hapi/lib/types/server/server.d.ts:4:22 - error TS2307: Cannot find module 'joi' or its corresponding type declarations.

4 import { Root } from 'joi';
                       ~~~~~

create any packet.json and have skipLibCheck = true

node_modules/@hapi/hapi/lib/types/route.d.ts:2:68 - error TS2307: Cannot find module 'joi' or its corresponding type declarations.

2 import { ObjectSchema, ValidationOptions, SchemaMap, Schema } from 'joi';
                                                                     ~~~~~

node_modules/@hapi/hapi/lib/types/server/server.d.ts:4:22 - error TS2307: Cannot find module 'joi' or its corresponding type declarations.

4 import { Root } from 'joi';
                       ~~~~~

No error

kanongil:
This PR removes the explicit import of `Joi` in the typings. As it is, any typescript users that use Hapi, will need to add a dependency of Joi, just to resolve the typings.

The explicit dependency is replaced by a generic validator that can be set on the `Server` object, or inferred from the config. This has the additional advantage, that other validators will be able to be typed instead of assuming that Joi is used.

The new types will require a bit of work to adapt to for projects that use `server.validate()`.

This fixes #4491.

### Definition

```js
const MySimpleValidator = {
    compile(schema: MySchema) {

        return {
            validate(value: unknown, options: MyOptions & { context?: Validation.Context }) {

                return { value: 'everything is awesome' };
            }
        };
    }
};
```

The base Joi object already implement this interface.

### Registration

Explicit:
```js
const server = new Server<ServerApplicationState, typeof Joi>({});
server.validator(Joi);
```

Inferred from config:
```js
const server = new Server({
    routes: { validate: { validator: Joi } }
});
```

Using the new `validator()` returned value:
```js
const server = new Server({})
    .validator(Joi);
```

#### For plugins

Explicit:
```js
register(server: Server<unknown, typeof Joi>, options: any) {

    server.validator(Joi);
    …
}
```

Using the new `validator()` returned value:
```js
register(_server: Server, options: any) {

    const server = _server.validator(Joi);
    …
}
```

### Usage

Once registered, the validator typings are used to help type the validation options:
```js
server.route({
    …,
    options: {
        validate: {
            options: {
                // Validated against the options that can be passed to the registered `validate()` method
            },
            query: {
                // Validated against the type of the `schema´ from the registered `compile()` method
            }
        }
    }
});
```

It can be overriden at the route level:
```js
server.route({
    …,
    options: {
        validate: {
            validator: Joi,   // Route-only validator
            options: {
                // Validated against the options that can be passed to the validator `validate()` method
            },
            query: {
                // Validated against the type of the `schema´ from the validator `compile()` method
            }
        }
    }
});
```

It also allows custom inline validators:
```js
server.route({
    …,
    options: {
        validate: {
            query: (value, optionsAndContext) {

                // optionsAndContext is typed with the `context` object
            }
        }
    }
});
```

It is also used to validate rules:
```js
server.rules(processor, {
    validate: {
        schema: validateSchema,    // Validated against the type of the `schema´ from the registered `compile()` method
        options: {
            // Validated against the options that can be passed to the registered `validate()` method
        }
    }
});
```

As it currently is, pre-compiled Joi schemas continue to work, even with no registered validator (though any options won't be validated):

```js
server.route({
    …,
    options: {
        validate: {
            query: Joi.object({
                …
            })
        }
    }
});
```

FYI, it's possible that this could be revised to extracts the allowed options from the passed Joi object.


Marsup:
The PR makes sense to me, but applying it on my project causes a whole bunch of errors along the lines of `Object literal may only specify known properties, and 'q' does not exist in type 'Validator<never> | DirectValidator<"query">'.`. It appears I need to wrap all my object schemas (eg. `validate: { query: { q: Joi.whatever() }}`) into joi objects, I can't have simple objects anymore. Am I missing something?

kanongil:
Thank for the feedback. Unfortunately, this will likely take a bit of work to adapt to.

Specifying a validator is required to type such simple validation objects, since it can no longer default to Joi. You would need to use one of the above methods, or explicit casting, to make your `Server` object accept the `Joi` validator. You can also override it when specifying the route:

```js
server.route<ReqRefDefaults, typeof Joi>({
    …
})
```

Of course, you can still just wrap it as suggested, though that would still lose the typings on any `validate.options`.

Marsup:
Considering the number of times a user would call `server.route`, can't we find a better way to define that with some declaration merging?

Marsup:
I'm torn, you're certainly right, but declaring independent plugins as I do, all the plugins will have to be changed to something like:
```ts
const plugin: Plugin<{}> = {
  register(server: Server<{}, typeof Joi>) {
  }
};
```
It doesn't look like a good DX but I'm not sure there's a better sane choice.

lenovouser:
I am not sure, but this might also fix https://github.com/hapijs/hapi/issues/4414. Currently the solution mentioned in that issue works with raw JS, but with TS (which is the reason you would use zod in the first place) it doesn't work and shows type errors everywhere.


Screenshot 2024-06-16 at 11 21 03
Screenshot 2024-06-16 at 11 21 18
Screenshot 2024-06-16 at 11 21 32

benkcrites:
We are also having issues trying to use zod, hapi, and typescript. While the code works fine, we getting type errors like @lenovouser is. 

Example:
```
    {
        method: 'GET',
        path: '/reservations',
        handler: scheduleController.getReservations,
        options: {
            validate: {
                validator: validateZod,
                query: z.object({
                    startBeforeTs: z.coerce.date().optional(),
                    endAfterTs: z.coerce.date().optional(),
                }),
            },
        },
    }
```
throws 
```
Type 'ZodObject<{ startBeforeTs: ZodOptional<ZodDate>; endAfterTs: ZodOptional<ZodDate>; }, "strip", ZodTypeAny, { startBeforeTs?: Date; endAfterTs?: Date; }, { ...; }>' is not assignable to type 'RouteOptionsResponseSchema'.
  Type 'ZodObject<{ startBeforeTs: ZodOptional<ZodDate>; endAfterTs: ZodOptional<ZodDate>; }, "strip", ZodTypeAny, { startBeforeTs?: Date; endAfterTs?: Date; }, { ...; }>' is not assignable to type 'PartialSchemaMap<any>'.
    Index signature for type 'string' is missing in type 'ZodObject<{ startBeforeTs: ZodOptional<ZodDate>; endAfterTs: ZodOptional<ZodDate>; }, "strip", ZodTypeAny, { startBeforeTs?: Date; endAfterTs?: Date; }, { ...; }>'.ts(2322)
```

JimmyBjorklund:
Any progress on this  ?

PR #4515, opened 2024-07-02
fix: 🐛 allow typed server methods access cache methods

damusix:
~also fix pre typing issue~ would be fixed by https://github.com/hapijs/hapi/pull/4489 . Change removed.

### DIFF
--- a/lib/types/server/methods.d.ts
+++ b/lib/types/server/methods.d.ts
@@ -7,7 +7,7 @@ export type CachedServerMethod<T extends AnyMethod> = T & {
         drop(...args: Parameters<T>): Promise<void>;
         stats: CacheStatisticsObject
     }
-}
+};
 
 /**
  * The method function with a signature async function(...args, [flags]) where:
@@ -16,7 +16,7 @@ export type CachedServerMethod<T extends AnyMethod> = T & {
  * * * ttl - 0 if result is valid but cannot be cached. Defaults to cache policy.
  * For reference [See docs](https://github.com/hapijs/hapi/blob/master/API.md#-servermethodname-method-options)
  */
-export type ServerMethod = AnyMethod
+export type ServerMethod = AnyMethod;
 
 /**
  * The same cache configuration used in server.cache().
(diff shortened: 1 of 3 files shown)


kanongil:
Fix a pre-typing issue? Can you elaborate? To me it reads like you _created_ an issue with this PR.

damusix:
@kanongil I like the strong typing, but for practical purposes, you can't use utilities that don't carry the same `Refs`.

If you run it locally, observe the type test: 

```

        pre: [
            {
                assign: 'some',
                method: ((req) => {

                    return req.app.word;
                }) as Lifecycle.Method<RequestDecorations>
            },
            {
                assign: 'another',
                method: getNum
            }
        ]
```

In the current state of typings, where one is used as a utility (`getNum`) and the other is defined in the object. When we define `pre[].method()` to take a `Refs` generic, it wants all pres to take the same `Refs` generic, or else it errors.

If you have other ideas, that'd be great, cause I agree that this is sub-optimal.

damusix:
@Marsup @kanongil updated commit without any references to pres. Ready for release.

lib/types/server/methods.d.ts
```diff
@@ -1,4 +1,13 @@
...
+import { CacheStatisticsObject, PolicyOptions } from "@hapi/catbox";
+
+type AnyMethod = (...args: any[]) => any;
+
+export type CachedServerMethod<T extends AnyMethod> = T & {
+    cache?: {
+        drop(...args: Parameters<T>): Promise<void>;
+        stats: CacheStatisticsObject
+    }
+}
```
Marsup, line 10:
```suggestion
};
```

lib/types/server/server.d.ts
```diff
@@ -203,7 +204,7 @@ export class Server<A = ServerApplicationState> {
      * server method name is an object property.
      * [See docs](https://github.com/hapijs/hapi/blob/master/API.md#-servermethods
      */
-    readonly methods: ServerMethods;
+    readonly methods: ServerMethods
```
Marsup, line 207:
```suggestion
    readonly methods: ServerMethods;
```

lib/types/server/server.d.ts
```diff
@@ -52,6 +52,7 @@ import {
 } from './methods';
 import { ServerOptions } from './options';
 import { ServerState, ServerStateCookieOptions } from './state';
+import { CacheStatisticsObject } from '@hapi/catbox';
```
Marsup, line 55:
```suggestion
```

Marsup (commented):
Just fixing formatting issues

PR #4518, opened 2024-07-27
Added isoTimestamp option

nkiesel:
When set to true, it renders the timestamp in log entries using `Date.toIsoString()`, which is easier to read by humans than the default of milliseconds.

### DIFF
--- a/API.md
+++ b/API.md
@@ -53,6 +53,13 @@ the framework.
 
 Cannot be set to `false` along with a [`port`](#server.options.port) value.
 
+#### <a name="server.options.isotimestamp" /> `server.options.isoTimestamp`
+
+Default value: `false`.
+
+If `true`, shows the log timestamp in ISO-8601 UTC format. If `false`, shows the log timestamp
+in milliseconds since 1970-01-01.
+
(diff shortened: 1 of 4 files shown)


kanongil (commented):
Thanks for the PR.

I have definitely been bothered about having to convert numbered timestamps to a human readable format, so understand the use case.

I don't however think that this is the best way about it, as other tools might require the timestamp to be a number. This is really more of a formatting issue, which would be better solved in the logging tool that displays the value.

nkiesel:
Given that this can be configured, why is that an issue for other tools? I looked at the podium repo - which I think is used by Hapijs - for adding the formatting, but could not see how/where to add that. I also thought to pass the formatted timestamp, but that is currently a number and anyway looks to me the passed optional timestamp is anyway ignored.

I would be happy to add another PR, but would need some advice how/where to do this.

kanongil:
Hapi will not display the timestamp, so it will be some other plugin or user code that renders it.

Podium just forwards the event to any registered listeners, and should definitely not change the format.

damusix:
@nkiesel Any chance you can see why CI failed on this one? We may merge it for upcoming version.

PR #4519, opened 2024-07-28
Test that logged simple handler errors has a stack

kanongil:
This adds a test for the `error.stack` property to illustrate a node v22 regression.

This should be able to be fixed with https://github.com/hapijs/hoek/pull/390.

Again, feel free to backport my old PR to fix this, but personally I would prefer to declare that Hapi v21 and any modules that use `Hoek.clone()` on potential errors (at least Hoek, Boom, and Podium) does not support node 21+.


Marsup:
Looks like it passes now that hoek 11.0.5 is released. I'll make another PR to force that version.

PR #4525, opened 2024-08-26
Update decorator types to be correct

### LINKED ISSUE #4524: Incorrect Types for decorators
### Runtime

Node.js

### Runtime version

v20.11.1

### Module version

@hapi/hapi v21.3.10

Trying to add an `object` via a decorator to server. Whilst using JSDOC/TypeScript and we are getting an error we have to disable.

> The below example is concerned with server decorate, but this is also appears to be an issue for all Interfaces bar the Handler

### What we can do
When using decorators you can provide any value as the `method` argument and Hapis internals will add it to the  interface being decorated.

fEx:

```
server.decorate('server', 'isSomethingOn', true)
server.decorate('server', 'anObject', { name: 'ben' })
```

The example in Hapi code for a server decorator can be seen in [lib/server.js#L182](https://github.com/hapijs/hapi/blob/22377ee7329b553f9ef72ff869940dce05523920/lib/server.js#L182)

```
// Server

if (typeof property === "string") {
    Hoek.assert(!Object.getOwnPropertyNames(internals.Server.prototype).includes(property), "Cannot override the built-in server interface method:", propertyName);
} else {
    Hoek.assert(!Object.getOwnPropertySymbols(internals.Server.prototype).includes(property), "Cannot override the built-in server interface method:", propertyName);
}

this._core.instances.forEach((server) => {

    server[property] = method;
});
```

### What the Types say you can do
However the Types only allow for functions:

> [lib/types/server/server.d.ts#L312-L319](https://github.com/hapijs/hapi/blob/master/lib/types/server/server.d.ts#L312-L319)

```
decorate(type: 'server', property: DecorateName, method: (existing: ((...args: any[]) => any)) => DecorationMethod<Server>, options: {apply?: boolean | undefined, extend: true}): void;
decorate(type: 'server', property: DecorateName, method: DecorationMethod<Server>, options?: {apply?: boolean | undefined, extend?: boolean | undefined}): void;
```

And the docs hint that it is intended that the `method` argument is used with other values:
> method - the extension function or other value.
> [lib/types/server/server.d.ts#L304](https://github.com/hapijs/hapi/blob/master/lib/types/server/server.d.ts#L304)

### Are we using this correctly?

- Are we using decorators correctly? 
- Is adding an `object` via a decorators `method` argument to `Server` correct?
- Is this better put on `Server.app`? Or `Request.app`?

### Can we raise a PR?

Is an update to the Types around decorators to allow non functions to be passed to `method` a PR you would accept?

We have to bypass decorators that have a non function value provided to the `method`  argument with `// @ts-expect-error`

According to the code, to be able to add any value via a decorators `method` argument

feedmypixel:
- The code currently allows any value to be added as a decorator, but the types do not. This work allows any value to be added to a decorator by updating the types to reflect what the code does
- Add relevant tests
- Update inline docs
- Create and export `DecorationValue` type
- Fixes #4524

### DIFF
--- a/lib/types/plugin.d.ts
+++ b/lib/types/plugin.d.ts
@@ -248,12 +248,7 @@ export interface HandlerDecorationMethod {
 }
 
 /**
- * The general case for decorator values added via server.decorate.
- */
-export type DecorationValue<T> = DecorationMethod<T> | any;
-
-/**
- * Decorator function.
+ * The general case for decorators added via server.decorate.
  */
 export type DecorationMethod<T> = (this: T, ...args: any[]) => any;
 
--- a/lib/types/plugin.d.ts
+++ b/lib/types/plugin.d.ts
@@ -248,7 +248,12 @@ export interface HandlerDecorationMethod {
 }
 
 /**
- * The general case for decorators added via server.decorate.
+ * The general case for decorator values added via server.decorate.
+ */
+export type DecorationValue<T> = DecorationMethod<T> | any;
+
+/**
+ * Decorator function.
  */
 export type DecorationMethod<T> = (this: T, ...args: any[]) => any;
 
--- a/lib/types/server/server.d.ts
+++ b/lib/types/server/server.d.ts
@@ -13,7 +13,7 @@ import {
     ServerRegisterPluginObject,
     ServerRegisterPluginObjectArray,
     DecorateName,
-    DecorationMethod,
+    DecorationValue,
     HandlerDecorationMethod,
     PluginProperties
 } from '../plugin';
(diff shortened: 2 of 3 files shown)


feedmypixel:
@kanongil when you have a min let me know your thoughts on this. Thanks 🙏

kanongil (commented):
This implementation will remove the typing from an inline method declaration, as `X | any` will be reduced to `any`. I believe it should work better if additional declarations for `decorate` are added with `method: any` (or maybe even `value: any`).

feedmypixel:
> This implementation will remove the typing from an inline method declaration, as X | any will be reduced to any. I believe it should work better if additional declarations for decorate are added with method: any (or maybe even value: any).

Understood. I'll have a look at adding other value specific overloads. I was also wondering on changing the name of the arg to signify that any `value` can be passed. I'll refactor and push a commit later today

damusix:
@feedmypixel Agreed with Gil. Perhaps you can type this to be a value from `Request | Toolkit | Server` interface? I imagine it'd be some utility type that removes internal, reserved keys so that it can pickup whatever you extend via `declare module`:

https://github.com/hapijs/hapi/blob/master/lib/request.js#L18
https://github.com/hapijs/hapi/blob/master/lib/toolkit.js#L11
https://github.com/hapijs/hapi/blob/master/lib/response.js#L25
https://github.com/hapijs/hapi/blob/master/lib/server.js#L174

Can you also add type tests? This should guarantee that nothing breaks, and you can see it in your editor in the tests if your intended types are producing `any`, errors, or not. Essentially, it's your place to verify that you're not potentially breaking anything.

https://github.com/hapijs/hapi/blob/master/test/types/index.ts

This is a good fix, though. I'm always typing around it with `X as never`.

feedmypixel:
@damusix Cheers - will look into this ASAP

damusix:
@feedmypixel Any updates on this?

feedmypixel:
> @feedmypixel Any updates on this?

Sorry I haven't had time to get to it. I'm hopeful of having some time over the Christmas period

damusix:
Perhaps this helps:
@feedmypixel https://www.typescriptlang.org/play/?#code/C4TwDgpgBAIhDGB7ATgQ2BAcqgttAvFAM7DICWAdgOZQA+xIOARogDYDcAUJ6JFAEoQiEZADcIAE0EBHAK5DgAaQggiUQgApOUHVADkwsSL11tuvbOSsTtMzr1yRIG3f1h0ACxe79eYB8QJb3McMjxg+2FgAFUrCP0ogFkIf0D4vQ8IVAkRInSyINMfPVQwMHSwVlkqSjyi82REWQx01GavevswZAgKnsEiMEQKYTrbYsoAM0R8ogBJCgArBAxC8fMUMioK1DQcMdc9dz2iAEFkNGdOt1QQVkRs9JJ0Xuu9HsHh4XS0AHd0iSIHCoSjpe7bN7gg7FKgQCgiF4DIYjXqcACUXB44GgAFEAB7wCBgYDYPAAHgAKgA+dRQClQCB4jAUCRqAYicRSCCOEjKVRQAD8UHh4mQUAAXHTMZxJrIKPBgGRhlAckg0BgoJSGUy4azYAgUC9SRAqRpeBBJe9ufISHoADRQbqISDIUCS-GE4nGylUh1+AISSUaAB0od2VElqAoIAA2gBdNHqGlRkCJgDeAF9uAB6bN0jxkNSF4WIYBQVBQD4cyRQADWKl+KAknFVhowGitPOA9v0FEQVAAtFCexpE-gaZmMTm8xSC0W1C2DeqIB2el2e3phw7R0moJP2EA

feedmypixel:
@damusix wonderful thanks 👍🏻

damusix:
Continued in https://github.com/hapijs/hapi/pull/4538

PR #4528, opened 2024-09-26
Add custom graceful shutdown handler

afharo:
#4473 introduced a piece of logic to immediately destroy the requests during the `stopping` phase: https://github.com/hapijs/hapi/blob/22377ee7329b553f9ef72ff869940dce05523920/lib/core.js#L509-L513

This caused a breaking change in my API where I have a custom `onRequest` handler that intercepts the request and returns a _graceful_ `503` message explaining that the server is going down.

I tried to add a custom handler replacement in this PR, but I wanted to cover that piece of code first with the tests. I can't get my head around as to why my new tests are not hitting that code path. Can someone help me to figure it out?

Thanks!

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -285,6 +285,7 @@ internals.server = Validate.object({
         .default(),
     routes: internals.routeBase.default(),
     state: Validate.object(),                                    // Cookie defaults
+    stoppingHandler: Validate.function(),
     tls: Validate.alternatives([
         Validate.object().allow(null),
         Validate.boolean()
--- a/lib/core.js
+++ b/lib/core.js
@@ -81,6 +81,7 @@ exports = module.exports = internals.Core = class {
     toolkit = new Toolkit.Manager();
     type = null;
     validator = null;
+    stoppingHandler = (req, res) => req.destroy();
 
     extensionsSeq = 0;                                                         // Used to keep absolute order of extensions based on the order added across locations
     extensions = {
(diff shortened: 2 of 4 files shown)


damusix (approved):
LGTM, @kanongil what do you think?

kanongil (changes requested):
I don't think we should support this usage, especially since it only works on node <18.

If you want to continue doing this, you will probably need to find some hacky workaround.

afharo:
Thanks for the review! 🧡 

> especially since it only works on node <18

Just FMI, why would it only work on Node <18?

My API currently runs on Node 20.15.1 with `@hapi/hapi@20.2.2` and the _graceful responses_ work as expected. They break after upgrading hapi to the latest because the [_early_ `req.destroy()`](https://github.com/hapijs/hapi/blob/22377ee7329b553f9ef72ff869940dce05523920/lib/core.js#L509-L513) intercepts any incoming connections before my `onRequest` extension kicks in.

kanongil:
Hmm, interesting. How do you know it actually works?

The test cases we have that exercise the `req.destroy()` does not trigger when running node 20+, and you failed to create any as well.

I would be very much interested, if you can produce a test that runs the `req.destroy()` codeline on node 20+.

afharo:
I found the culprit!

[This line](https://github.com/hapijs/hapi/blob/52bb6d2593b4a44ac9a476cb1ee921a05769a426/lib/core.js#L469) closes the listener, so it won't accept more connections after calling `stop()`.

According to [Node.js docs](https://nodejs.org/docs/latest-v20.x/api/http.html#serverclosecallback), it changed its behavior on v19, which is probably why @kanongil recalls different behavior between v18 and 20.

I still don't get why my API behaves differently with the latest version of this library, though. But, in any case, it seems like, if we want to provide _graceful_ rejections we might want to subscribe to `this.listenerr.once('close'` to apply the current callback to the `.close()` method, and call `this.listener.close()` only when all ongoing requests are completed or the shutdown timeout expires.

kanongil:
Thank you very much for your efforts!

I have used this new knowledge to create a PR #4528, which reverts the `req.destroy()` behaviour.

afharo:
Thank you! I'll close my PR then 🧡

PR #4538, opened 2025-01-19
fix: 🐛 typings around server.decorate

damusix:
- Types match the validation a bit more. 
- Type errors when attempting to decorate a builtin.
- No `any` when decorating a non-method
- Can never extend a handler method
- Can never extend or apply a non-method

Continued from https://github.com/hapijs/hapi/pull/4525

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -120,7 +120,7 @@ internals.Server = class {
 
         const existing = this._core.decorations[type].get(property);
         if (options.extend) {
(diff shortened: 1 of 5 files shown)


damusix:
@kanongil @Marsup Need your eyes por favor. I'll fix quickly.

@feedmypixel Thanks for the initiation of this PR. I think I've finished it. If you have suggestions, feel free to comment.

feedmypixel (approved):
LGTM 👍🏻 

@damusix thanks so much for picking this up and apologies I didn't get to it.

PR #4539, opened 2025-01-19
Rework tests to not use @hapi/joi@15

kanongil:
Same as #4474, except applied against master branch.


damusix (approved):

Marsup:
Sorry, I'm late to this one, but why target master additionally to the next branch? Do you actually want it in a non-breaking release?

kanongil:
> Sorry, I'm late to this one, but why target master additionally to the next branch? Do you actually want it in a non-breaking release?

Sure. This now just changes the testing infrastructure to not rely on the old Joi release.

PR #4544, opened 2025-02-14
fix: complies access scope validation to RFC8693

nlepage:
## Problem

[RFC8693](https://datatracker.ietf.org/doc/html/rfc8693) on OAuth2.0 Token Exchange specifies that [credentials scope](https://datatracker.ietf.org/doc/html/rfc8693#name-scope-scopes-claim) may contain a space-separated list of scopes, however hapi’s access scope only supports either a string containing a single scope or an array containing a list of scopes.

## Proposal

Allow using a string containing a space-separated list of scopes.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -555,9 +555,10 @@ internals.validateScope = function (credentials, scope, type) {
         return true;
     }
 
-    const count = typeof credentials.scope === 'string' ?
-        scope[type].indexOf(credentials.scope) !== -1 ? 1 : 0 :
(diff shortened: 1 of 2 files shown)


kanongil:
Thanks for the PR!

I have had a look at how Hapi handles scopes, and I don't agree that Hapi has an issue as such. The format of `credentials.scope` is defined by Hapi (though not explicitly), not any specific spec. It is the job of the auth strategy/scheme to convert to this format when needed. Expanding this interface should be done with great caution.

How are you applying it / what plugin are you using? If it provides a hook, you might be able to massage it in a callback.

PR #4549, opened 2025-07-26
feat: add `br` & `zstd` to supported compression 🗜️ algos

bricss:
This PR is an attempt to bring support for `br` & `zstd` compression 🗜️ algos into the core.

### DIFF
--- a/lib/compression.js
+++ b/lib/compression.js
@@ -16,30 +16,64 @@ const internals = {
         'deflate, gzip',
         'gzip',
         'deflate'
-    ]
+    ],
+    provision: new Map([
+        ['zstd', [
+            (options) => Zlib.createZstdCompress(options),
+            (options) => Zlib.createZstdDecompress(options),
+            {
+                params: {
+                    [Zlib.constants.ZSTD_c_compressionLevel]: 6
+                }
+            }
+        ]],
+        ['br', [
+            (options) => Zlib.createBrotliCompress(options),
+            (options) => Zlib.createBrotliDecompress(options),
+            {
+                params: {
+                    [Zlib.constants.BROTLI_PARAM_QUALITY]: 4
+                }
+            }
+        ]],
+        ['deflate', [
+            (options) => Zlib.createDeflate(options),
+            (options) => Zlib.createInflate(options)
+        ]],
+        ['gzip', [
+            (options) => Zlib.createGzip(options),
+            (options) => Zlib.createGunzip(options)
+        ]]
+    ])
 };
 
 
 exports = module.exports = internals.Compression = class {
 
-    decoders = {
-        gzip: (options) => Zlib.createGunzip(options),
-        deflate: (options) => Zlib.createInflate(options)
-    };
+    decoders = {};
 
-    encodings = ['identity', 'gzip', 'deflate'];
+    encodings = ['identity'];
 
     encoders = {
-        identity: null,
-        gzip: (options) => Zlib.createGzip(options),
-        deflate: (options) => Zlib.createDeflate(options)
+        identity: null
     };
 
     #common = null;
 
-    constructor() {
+    constructor({ compression }) {
 
-        this._updateCommons();
+        if (!compression) {
+            this._updateCommons();
+        }
+
+        for (const [alg, [encoder, decoder, defaults = {}]] of internals.provision.entries()) {
+            let conditions = compression?.encodings?.[alg];
+            if (conditions) {
+                conditions = Hoek.applyToDefaults(defaults, conditions);
+                this.addEncoder(alg, (options = {}) => encoder(Hoek.applyToDefaults(conditions, options)));
+                this.addDecoder(alg, (options = {}) => decoder(Hoek.applyToDefaults(conditions, options)));
+            }
+        }
     }
 
     _updateCommons() {
@@ -123,29 +157,6 @@ exports = module.exports = internals.Compression = class {
         const encoder = this.encoders[encoding];
         Hoek.assert(encoder !== undefined, `Unknown encoding ${encoding}`);
         return encoder(request.route.settings.compression[encoding]);
-    }
-
-    enableBrotliCompression(compressionOptions) {
-
-        const defaults = {
-            params: {
-                [Zlib.constants.BROTLI_PARAM_QUALITY]: 4
-            }
-        };
-        compressionOptions = Hoek.applyToDefaults(defaults, compressionOptions);
-        this.decoders.br = (options) => Zlib.createBrotliDecompress({ ...options, ...compressionOptions });
-        this.encoders.br = (options) => Zlib.createBrotliCompress({ ...options, ...compressionOptions });
-        this.setPriority(['br']);
-    }
-
-    enableZstdCompression(compressionOptions) {
-
-        /* $lab:coverage:off$ */
-        Hoek.assert(!!Zlib.constants.ZSTD_CLEVEL_DEFAULT, 'Zstd is not supported by the engine');
-        this.decoders.zstd = (options) => Zlib.createZstdDecompress({ ...options, ...compressionOptions });
-        this.encoders.zstd = (options) => Zlib.createZstdCompress({ ...options, ...compressionOptions });
-        this.setPriority(['zstd']);
-        /* $lab:coverage:on$ */
     }
 
     setPriority(priority) {
--- a/lib/compression.js
+++ b/lib/compression.js
@@ -23,7 +23,8 @@ const internals = {
             (options) => Zlib.createZstdDecompress(options),
             {
                 params: {
-                    [Zlib.constants.ZSTD_c_compressionLevel]: 6
+                    [Zlib.constants.ZSTD_c_compressionLevel]: 6,
+                    [Zlib.constants.ZSTD_d_windowLogMax]: 0
                 }
             }
         ]],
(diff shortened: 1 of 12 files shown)


lib/compression.js
```diff
@@ -6,25 +6,37 @@ const Accept = require('@hapi/accept');
 const Bounce = require('@hapi/bounce');
 const Hoek = require('@hapi/hoek');
 
-
 const internals = {
-    common: ['gzip, deflate', 'deflate, gzip', 'gzip', 'deflate', 'gzip, deflate, br']
+    common: [
+        'gzip, deflate, br, zstd',
```
kanongil, line 11: It makes sense to update the common encoding with `'gzip, deflate, br, zstd'`, as this is used by both Chrome and Firefox. This can go in a separate PR.

kanongil (changes requested):
Zstd compression cannot be supported by hapi itself, as the [API is still experimental](https://nodejs.org/api/zlib.html#zlibcreatezstdcompressoptions). This will have to be implemented as a custom compressor / decompressor by the user (or a plugin).

I still don't see much point in adding built-in brotli support, which is already supported using a plugin. The built-in gzip compression makes sense, since it is widely supported and can provide major bandwidth savings. Newer compressors only add a marginal benefit and can have other trade-offs. As such they should only be enabled when configured by the user.

I like that it is possible to customise the encoder priority, but your implementation doesn't support user-defined encoders. A proper implementation would be more complex, and should probably go in a separate PR.

damusix:
This looks sane. I think it makes sense to have this be framework-supported out of the box since, especially given that it simply leverages NodeJS builtins. Even if they are experimental, it helps if the framework ships with it. to @kanongil 's point, however, I would add a notice and a link to the MDN docs explicitly stating that Zstd compression is experimental. Thank you @bricss

bricss:
I added a notice 📝 about the experiment status. And also the ability to pass compression options 🎛️ to both algorithms.

damusix (approved):

damusix:
@bricss Hey check out these skip patterns:

[`it('cleans unused file stream when response is overridden', { skip: !Common.hasLsof }, async () => {`](https://github.com/hapijs/hapi/blob/76e6cfeb4d6a4f7d5b945f9c50ac8e12b84e32a5/test/core.js#L1817)
[`it('creates a server listening on a unix domain socket', { skip: process.platform === 'win32' }, async () => {`](https://github.com/hapijs/hapi/blob/76e6cfeb4d6a4f7d5b945f9c50ac8e12b84e32a5/test/core.js#L164)

We're going to need to add some skip patterns for `zstd` tests so they only run over node 22. That said, it should also be documented that it's Node 22+.

Add the util in `./test/common.js` and check against `process.version`
`@hapi/somever` is [installed already](https://github.com/hapijs/hapi/blob/76e6cfeb4d6a4f7d5b945f9c50ac8e12b84e32a5/package.json#L39) so you can leverage that ([how to example here](https://github.com/hapijs/somever/blob/master/test/index.js#L389))

bricss:
All done ✅ added node version note and test skippers 🚧

kanongil (changes requested):
Regarding Zstd support, I don't think it makes sense to make an exception for the policy around node API feature stability. Especially because there is a standard way to integrate custom compressors.

For brotli, I still don't think we need to include it in core, but I'm not opposed to it, if done properly. This implementation still has several issues, most pressingly a default compression level of 11, which is *way too high* for just-in-time compression. The default should probably not be more computationally expensive than the default gzip option. A quick test on a 300KB json file shows gzip to take `20 ms`, while brotli takes `335 ms`! For this file, brotli needs to use level 5 to have comparable speed. I would probably choose 4 as default, improving both compression speed and a size.

I wish you would make a separate PR for the `priority` option, which provides a meaningful feature by itself.

bricss:
The issue with `br` compression level is solved ✅

As for the experimental status of `zstd`, if user gonna enable it manually, it kinda implies that user smart 🎓 enough to understand what he/she/ze is doing 🙂

kanongil:
Hapi core doesn't have "experimental" features, and built-in zstd compression certainly doesn't warrant introducing such a concept

bricss:
I couldn't find any policy that mentioned anything about experimental features is not baing allowed.

The only reason why `zstd` has experimental status is bcos it was recently added to the node core. It uses the official rock-solid 🪨 [zstd](https://github.com/facebook/zstd) library leveraged via `zlib`. Its experimental status is due to the amount of time ⌚ it needs to "marinate" 🥒 in the codebase before it can be unflagged.

bricss:
Meanwhile, [@fastify/compress](https://www.npmjs.com/package/@fastify/compress) & [hono-compress](https://www.npmjs.com/package/hono-compress) already have `zstd` support for months 🧐

damusix:
@kanongil It kills the ecosystem to not support things like this. I'll all for stability and not breaking user space, but this shouldn't break user space— only enhance it. It just enables Hapi to use modern JS features. If people are using Hapi, want these features, and contribute, I can't think of a good reason to not allow it.

@bricss Sorry mate, one last thing: you'll need to tell Lab to not cover those lines since it won't be covered in anything under Node 22 (https://hapi.dev/module/lab/api/?v=26.0.0#inline-enablingdisabling). Screenshot 2025-08-04 at 12 42 04 PM

bricss:
Pushed cure 🩹 for coverage. Hope it all will pass this time 🤞

kanongil:
> @kanongil It kills the ecosystem to not support things like this. I'll all for stability and not breaking user space, but this shouldn't break user space— only enhance it. It just enables Hapi to use modern JS features. If people are using Hapi, want these features, and contribute, I can't think of a good reason to not allow it.

That is quite the hot take. If it "kills the ecosystem" to have to use an alternative, slightly more complex, but more powerful API, then it is a very fragile ecosystem indeed. Many/most? deployments won't care about this, as the compression can already be applied in reverse proxies and CDNs.

Implementing features just because a user wants it, is never a good path. Features need to be maintained indefinitely, causing more maintenance work. As such, it should provide a tangible benefit before it should be considered. As I said, the "priority" option is an interesting candidate. Extending the built-in compressors less so, as it doesn't enable anything new. Using experimental APIs, exposing experimental features, and dropping the 100% code coverage policy, even less so!

I understand that you want to support new contributors, as do I, especially since @bricss seems to care a lot about doing it right. That is why I try to help guide their efforts. Unfortunately for new contributors, this project has (or at least used to have) a high coding standard, making it difficult to get anything but small fixes merged.

kanongil:
> I'd also like to add that I'm trying to add 🧩 these features to the framework 🧰 not just bcos I personally need them, but bcos I believe it'll be beneficial for the framework as a whole, and help attract 🧲 more users. This should be a win-win 🏆 for everyone — both the community and its users.

I very much appreciate your enthusiasm and efforts, and I apologise that our responses are not more aligned, making you get caught in the middle. Though our [contribution guidelines](https://hapi.dev/policies/contributing/) did advise you to "ask first", to avoid working on something that might not get accepted.

damusix:
> IMO, we can withhold `zstd` from this PR, and I’ll make all necessary changes to address remaining issues.

Yeah, makes sense

> Alternatively, we could move compression 🗜️ entirely out of the core, into a standalone package 📦 with the name `@hapi/compress`. But then it will be the same thing as now, only from the other side 🫠

That's what I was proposing to do, and perhaps use Gil's package as the base since he's already laid the ground work for it (up to you Gil). Thoughts?

bricss:
For "full out-of-the-box functionality" it will make more sense just to add missing algos into the core, bump min supported node version to `22.15.0`, and release new semver major. No one need support of all those fossil ⚱️ node versions in any way. And if they do, shame on them.

kanongil:
@bricss Please don't shame people for using older versions of node. There are a multitude of reasons not to upgrade. We try our best to have empathy, even if we don't understand their reasoning.

bricss:
If following semver approach is considered as elitism, then something went wrong. Coz, I was thinking that's how semver versioning works 🤷

And I'm not pushing for anything, I'm just expressing my opinion.

kanongil (changes requested):
We still don't need compression options at the server level. The route level options work fine.

Another major issue, is that this enables `br` and `zstd` for uploads *by default*, with *no option to disable*. Enabling new upload decompressor engines has both security and performance implications, and should not be done without a thorough review.

*This review is mainly aimed at the hapi maintainers. I don't plan to spend more time on this PR.*

bricss:
Summary 📋 list:
- All built-in encodings 🔣 can be enabled or disabled
- Encoder/decoder options 🎛️ can be set globally in server configuration or overridden per route
- Decoding can be disabled 🎚️, which will result in `415` responses ⛔
- Priority 🔢 for content encoding compression algorithms can be changed
- Zero 0️⃣ breaking changes with old behavior 🧩

lib/compression.js
```diff
@@ -5,33 +5,87 @@ const Zlib = require('zlib');
...
 exports = module.exports = internals.Compression = class {
 
-    decoders = {
-        gzip: (options) => Zlib.createGunzip(options),
-        deflate: (options) => Zlib.createInflate(options)
-    };
+    decoders = {};
 
-    encodings = ['identity', 'gzip', 'deflate'];
+    encodings = ['identity'];
```
damusix, line 58: Is this change accounted for somewhere else? Is this intentional? This could change existing behavior.

bricss: All `encodings` are now set in class `constructor` from `provision` map. This PR contains zero breaking changes, and code coverage is 100%.

PR #4550, opened 2025-07-30
Correct type for state object

jpcanepa:
The type for the `state` field of the `ServerOptions` object was qualified with a `TODO` that questioned if the type should be the full `ServerStateCookieOptions` or just those fields defined in the default value. Given [the documentation](https://hapi.dev/api/?v=21.4.0#-serverstatename-options) (and validated by actual usage), the correct type is indeed the one the TODO presumed.  

This PR replaces the ad-hoc type with the proper already-existing type, and gets rid of the TODO. 

This change is intrinsically _safe_ as the type is an strict superset of the one being in use today, with all extra fields being optional.


damusix (approved):
This seems correct @Marsup @kanongil 

https://github.com/hapijs/statehood/blob/master/lib/index.d.ts#L8

kanongil (approved):

PR #4551, opened 2025-08-06
fix: add Refs generic to PluginSpecificConfiguration interface

### LINKED ISSUE #4526: Improve PluginSpecificConfiguration intellisense adding Refs generic
### Runtime

NodeJS

### Runtime version

22.7.0

### Module version

21.3.10

### Used with

hapi application

_No response_

### What problem are you trying to solve?

I'm trying to write a plugin that accepts a function that receives the request as a param and it could be nice to have the generic from the route shared to plugins, something like the following:

```ts
// plugin.d.ts
interface PluginSpecificConfiguration<Refs extends ReqRef = ReqRefDefaults> {}

// routes.d.ts
interface CommonRouteProperties<Refs extends ReqRef = ReqRefDefaults> {
    plugins?: PluginSpecificConfiguration<Refs> | undefined;
}
```

With that, we could configure plugins like this

```ts
declare module '@hapi/hapi {
  interface PluginSpecificConfiguration<Refs extends ReqRef = ReqRefDefaults> {
    pluginName: PluginInterface<Refs>;
  }
}
```

I'm interested in writting a PR if y'all accept the idea :)

### Do you have a new or modified API suggestion to solve the problem?

_No response_

lgraziani2712:
Add Refs generic to plugins key for CommonRouteProperties, extending the generic for logic inside plugins.

This fixes situations where plugins API have lambdas where access the request object and requires to have the same Refs type from the route.

Fix #4526

### DIFF
--- a/lib/types/plugin.d.ts
+++ b/lib/types/plugin.d.ts
@@ -1,4 +1,4 @@
-import { RequestRoute } from './request';
+import { RequestRoute, ReqRef, ReqRefDefaults } from './request';
 import { RouteOptions } from './route';
 import { Server } from './server';
 import { Lifecycle } from './utils';
@@ -47,7 +47,7 @@ export interface PluginRegistered {
 export interface PluginsStates {
 }
 
-export interface PluginSpecificConfiguration {
+export interface PluginSpecificConfiguration<Refs extends ReqRef = ReqRefDefaults> {
 }
 
 export interface PluginNameVersion {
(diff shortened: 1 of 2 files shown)


damusix:
@lgraziani2712 existing tests passed... But I still fear breaking people's builds....

Can you add some typescript testing?

In particular, spell out your use-case in tests (Happy path).
And then make another use-case for where a user doesn't depend on ReqRefs:
- https://github.com/panascais-forks/yaral/blob/bcb1ba5458acdd33e682de70cd9d0c967e3a4611/src/yaral.ts#L35
- https://github.com/kanongil/nipo/blob/4335b6eef850884bc38492e3264483092e864755/lib/index.d.ts#L99
- https://github.com/hapi-swagger/hapi-swagger/blob/master/index.d.ts#L5

In fact, if you can pnpm link to your local build against those packages and see if it breaks their typings, that could be an issue.

If there are no issues, screenshot and post here for us to see and further discuss!

PR #4559, opened 2025-11-06
fix: 🐛 request.url getter fails when using IPv6 and HTTP2 [#4560]

gsoldevila:
### Summary

Fixes https://github.com/hapijs/hapi/issues/4560

#### Problem

In HTTP2 the `Host:` header is no longer present. According to the spec:

> In HTTP/2, the functionality of the HTTP/1.1 Host header is replaced by the :authority pseudo-header, which is a mandatory, colon-prefixed header that specifies the authority (host and port) of the target server. This change is part of HTTP/2's binary framing and header compression (HPACK), and any HTTP/2 request converted to HTTP/1.1 must create a Host header from the :authority pseudo-header.

* When building the request object and extracting its information ([here](https://github.com/hapijs/hapi/blob/master/lib/request.js#L627)), we rely solely on `Host:` header.
* Then, when starting Hapi in HTTP2 mode and performing a request, the `Host:` header is missing, so the [request.url getter](https://github.com/hapijs/hapi/commit/6b67d2435af841c5656d21c230aefa3cc74dda91) uses the `host:port` from the [_core.info](https://github.com/hapijs/hapi/blob/master/lib/request.js#L187).
* Now, if the server is configured with an IPv6 host (such as `::1`), the getter ends up building an invalid URL, as the host name is not surrounded by square brackets. An IPv6 address like `2001:db8::1:8080` would be ambiguous, as `8080` could be interpreted as the last segment of the IP address rather than the port.

#### Solution

The PR updates the request logic in 2 places:

* First, we default to using `:authority` if `Host` header is NOT present.
* Then, if for any reason both headers are missing and we default to using the server info, we make the logic support the IPv6 scenario by wrapping the _hostname_ (e.g. `::1`) in square brackets `[::1]`.

#### Tests

Two new tests cover the changes:

* `sets host info from :authority header when host header is absent` — simulates an HTTP/2 request by injecting with an empty `host` header and a `:authority` pseudo-header, asserting that `request.info.host` and `request.info.hostname` are correctly populated.
* `generates valid URL when server host is IPv6 and host header is absent` — creates a server with an IPv6 host (`::1`), clears `request.info.host` to trigger the fallback path, and asserts that `request.url.host` is correctly formatted as `[::1]:PORT`.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -207,7 +207,7 @@ exports = module.exports = internals.Request = class {
 
         // An IPv6 address contains at least two colons.
 
-        return (host.match(/:/g) || []).length >= 2;
+        return /:[^:]*:/.test(host);
     }
 
     _formatIpv6Host(host, port) {


lib/request.js
```diff
@@ -624,7 +647,7 @@ internals.Info = class {
         this._request = request;
 
         const req = request.raw.req;
-        const host = req.headers.host ? req.headers.host.trim() : '';
+        const host = (req.headers.host || req.headers[':authority'] || '').trim();
```
gsoldevila: Perhaps we could test if the req is an IPv6 request somehow, and choose `Host` or `:authority` accordingly. WDYT?

gsoldevila:
@damusix thanks for the feedback! I've added two tests covering the changes:

* `sets host info from :authority header when host header is absent` — simulates an HTTP/2 request (no `host` header, `:authority` pseudo-header present) and asserts that `request.info.host` and `request.info.hostname` are correctly populated from it.
* `generates valid URL when server host is IPv6 and host header is absent` — starts a server with an IPv6 host (`::1`), triggers the URL fallback path by clearing `request.info.host`, and asserts the resulting URL host is properly bracketed (`[::1]:PORT`).

Both use `server.inject` following the existing test patterns in `test/request.js`, so no real network listener is required. The tests do not need to be conditioned on IPv6 availability since they don't bind to any network interface.

lib/request.js
```diff
@@ -201,6 +203,18 @@ exports = module.exports = internals.Request = class {
         return this._url;
     }
 
+    _isBareIpv6(host) {
+
+        // An IPv6 address contains at least two colons.
+
+        return (host.match(/:/g) || []).length >= 2;
```
Marsup: Since this function is in a hot path, I'd favor a more performant alternative. Having tried a few in [this benchmark](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJ2X1JxNjBkc3BLTTB6bC1NNVFSS0ciLCJjb2RlIjoiREFUQS5pcHY2LnNwbGl0KCc6JykubGVuZ3RoID49IDIiLCJuYW1lIjoiU3BsaXQiLCJkZXBlbmRlbmNpZXMiOltdfSx7ImlkIjoiTGI0WGpaV18xUlc0SFg1V3Fuc3N0IiwiY29kZSI6IkRBVEEubm90aXB2Ni5zcGxpdCgnOicpLmxlbmd0aCA-PSAyIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6IlNwbGl0IChub3QgaXB2NikifSx7ImlkIjoiaUJRM0R2S2x4M0RIUUcyQ1VNc1RfIiwiY29kZSI6IkRBVEEuaXB2Ni5tYXRjaChEQVRBLnJlZ2V4TWF0Y2gpLmxlbmd0aCA-PSAyIiwibmFtZSI6IlJlZ2V4IG1hdGNoIiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6ImNiUUhWTmxScFZ5T09jREpUbUFhayIsImNvZGUiOiIoREFUQS5ub3RpcHY2Lm1hdGNoKERBVEEucmVnZXhNYXRjaCkgPz8gW10pLmxlbmd0aCA-PSAyIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6IlJlZ2V4IG1hdGNoIChub3QgaXB2NikifSx7ImlkIjoiODBpT2drRlRxdGNFQWZvMlh3WjhtIiwiY29kZSI6IkRBVEEucmVnZXhUZXN0LnRlc3QoREFUQS5pcHY2KSIsIm5hbWUiOiJSZWdleCB0ZXN0IiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6ImtrYllUZ28zWUNTOWxITGJpMlg1LSIsImNvZGUiOiJEQVRBLnJlZ2V4VGVzdC50ZXN0KERBVEEubm90aXB2NikiLCJkZXBlbmRlbmNpZXMiOltdLCJuYW1lIjoiUmVnZXggdGVzdCAobm90IGlwdjYpIn1dLCJjb25maWciOnsibmFtZSI6IkJhc2ljIGV4YW1wbGUiLCJwYXJhbGxlbCI6dHJ1ZSwiZ2xvYmFsVGVzdENvbmZpZyI6eyJkZXBlbmRlbmNpZXMiOltdfSwiZGF0YUNvZGUiOiJjb25zdCBpcHY2ID0gJzo6MSc7XG5jb25zdCBub3RpcHY2ID0gJzEyNy4wLjAuMSc7XG5jb25zdCByZWdleE1hdGNoID0gLzovZztcbmNvbnN0IHJlZ2V4VGVzdCA9IC86W146XSo6L2c7XG5yZXR1cm4ge1xuICBpcHY2LFxuICBub3RpcHY2LFxuICByZWdleE1hdGNoLFxuICByZWdleFRlc3QsXG59OyJ9fQ), the regex for the test version seems consistently better both for the positive and negative scenarios. Let me know if my benchmark is biased though.

gsoldevila: Good catch, and thanks for the benchmark! I've validated that `/:[^:]*:/` is semantically equivalent to the previous colon-count approach — a string has two or more colons if and only if it contains two colons with zero or more non-colon characters between them, which is exactly what this regex matches.

I've applied it in 170a005. One small note: I omitted the `g` flag from the benchmark's `/:[^:]*:/g` since `.test()` with `g` advances `lastIndex` on a shared regex, which can cause subtle bugs. Since we're using it as an inline literal the impact is zero here, but it's cleaner to leave it out.

Marsup:
Thanks for the PR, and sorry it fell through the cracks, it's now released.

Issue #4561, opened 2025-12-05
Possible typing regression between v21.3.12 and v21.4.4


lgraziani2712:
### Runtime

node.js

### Runtime version

24.11.1

### Module version

21.4.4

### Last module version without issue

21.3.12

### Used with

hapi application

### Any other relevant information

Is related to changes on TS typings

### What are you trying to achieve or the steps to reproduce?

```ts
declare module '@hapi/hapi' {
    interface ReqRefDefaults {
        AuthCredentialsExtra: { some_id?: string }
    }
}

export function processAuth<Refs extends ReqRef>(req: Request<Refs>): void {
  if (
    !req.auth.user &&
    req.auth.isAuthenticated &&
    // Here, .some_id doesn't exists and will fail with ts2339 https://typescript.tv/errors/#ts2339
    req.auth.credentials.some_id
  ) {
    // Do something
  }
}
```

I need to use a generic here because this function is called in multiple routes where each of them has different specification for Params or Headers properties. For what I've seen, the typing changes introduced a "bug" (IDK if is actually a bug or a strict validation).

If instead of a generic, I define a real interface, the bug won't happen, but then I will get issues on each of the calls. E.g.

```ts
// file a.ts
// No more generic
export function processAuth(req: Request): void {
  if (
    !req.auth.user &&
    req.auth.isAuthenticated &&
    req.auth.credentials.some_id
  ) {
    // Do something
  }
}

// file b.ts
import { processAuth } from './a.js';

// define route and then in handler
server.route<{ Params: SomeParamInterface; }>({
  async handler(req, h) {
    // This will fail because local req and the defined param req have two different ReqRes interfaces.
    // This wasn't an issue with 21.3
    processAuth(req);
  }
})
```

### What was the result you got?

A TS error on the line 

```ts
    req.auth.credentials.some_id
```

### What result did you expect?

To resolve the generics accordingly when merging the attributes between the Generic and the Default interfaces as v21.3.12 does.

Thank you in advance.

lgraziani2712:
Sure thing! Here is:

[...]

PR #4562, opened 2026-02-13
fix: auth typings and reduced type ambiguity

damusix:
- RFC, claude code helped me with this one. It looks correct.
- Fixed regression in `MergeType` utility that broke auth https://github.com/hapijs/hapi/issues/4561
- Add type missed merges
- Remove ambiguous types
- Added more TS tests. LMK if you want me to remove code comments

### DIFF
--- a/lib/types/request.d.ts
+++ b/lib/types/request.d.ts
@@ -252,7 +252,7 @@ export interface RequestLog {
 }
 
 export interface RequestQuery {
-    [key: string]: any;
+    [key: string]: string | string[] | undefined;
 }
 
 /**
@@ -271,9 +271,9 @@ export interface InternalRequestDefaults {
 
     Payload: stream.Readable | Buffer | string | object;
     Query: RequestQuery;
-    Params: Record<string, any>;
+    Params: Record<string, string>;
     Pres: Record<string, any>;
-    Headers: Record<string, any>;
+    Headers: Record<string, string | string[] | undefined>;
     RequestApp: RequestApplicationState;
 
     AuthUser: UserCredentials;
@@ -439,7 +439,7 @@ export interface Request<Refs extends ReqRef = ReqRefDefaults> extends Podium {
     /**
      * Same as pre but represented as the response object created by the pre method.
      */
-    readonly preResponses: Record<string, any>;
+    readonly preResponses: Record<string, unknown>;
 
     /**
      * By default the object outputted from node's URL parse() method.
@@ -472,7 +472,7 @@ export interface Request<Refs extends ReqRef = ReqRefDefaults> extends Podium {
     /**
      * An object containing parsed HTTP state information (cookies) where each key is the cookie name and value is the matching cookie content after processing using any registered cookie definition.
      */
-    readonly state: Record<string, any>;
+    readonly state: Record<string, unknown>;
 
     /**
      * The parsed request URI.
--- a/lib/types/request.d.ts
+++ b/lib/types/request.d.ts
@@ -89,6 +89,8 @@ export interface RequestAuth<
      * set to false.
      */
     isAuthorized: boolean;
+    /** true if the request has been authenticated via the `server.inject()` `auth` option, otherwise `undefined`. */
+    isInjected?: boolean | undefined;
     /** the route authentication mode. */
     mode: AuthMode;
     /** the name of the strategy used. */
@@ -303,11 +305,7 @@ export type ReqRef = Partial<Record<keyof ReqRefDefaults, unknown>>;
 /**
  * Utilities for merging request refs and other things
  */
-export type MergeType<T, U> = {
-    [K in keyof T]: K extends keyof U
-        ? U[K]
-        : T[K];
-};
+export type MergeType<T, U> = Omit<T, keyof U> & U;
 
 export type MergeRefs<T extends ReqRef> = MergeType<ReqRefDefaults, T>;
 
--- a/typescript.md
+++ b/typescript.md
@@ -1,0 +1,800 @@
+
+## TypeScript Support
+
+hapi ships built-in TypeScript definitions (`.d.ts`) — no `@types/hapi` package needed.
+The type system is designed around two complementary patterns:
+
+- **Module augmentation** — declare global types that apply to every route (e.g. `UserCredentials`, `ServerApplicationState`).
+- **Generic refs** — pass per-route type overrides via `ServerRoute<Refs>`, `Request<Refs>`, and `Lifecycle.Method<Refs>`.
+
+Both patterns can be used together. Module augmentation sets the baseline; generic refs narrow types for individual routes.
+
+
+## Quick Start
+
+```typescript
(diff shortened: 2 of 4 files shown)


fjeldstad:
This should have been part of a major version bump as the type changes are breaking in some scenarios. For example when extracting a header value; that used to be `string | undefined` while now it is `string | string[] | undefined` meaning that it cannot be assigned to a string variable anymore. Even if it is more _correct_ now, I don't think consumers expect the build to break from installing a patch version bump.

Marsup:
@fjeldstad we already answered that in https://github.com/hapijs/hapi/issues/4563. Breaking types is inherent to the typesrcipt ecosystem, nobody releases a major version just for types correctness.

kanongil:
> nobody releases a major version just for types correctness

FYI, some projects do try to do that by eg. adhering to the proposed [semver-ts spec](https://www.semver-ts.org). However, even with that, changes in minor releases can cause type compilation issues! As [they state](https://www.semver-ts.org/formal-spec/index.html#:~:text=Accordingly%2C%20we%20propose%20the%20rules%20below%2C%20with%20the%20caveat%20that%20(as%20noted%20in%20several%20places%20throughout)%20they%20will%20not%20prevent%20all%20possible%20breakage—only%20the%20majority%20of%20it) (paraphrased):

> We propose the rules below, with the caveat that they *will not prevent all possible breakage*—only the majority of it.

As such, for typescript development, one must expect that any typings update can potentially cause "breakage".

lib/types/request.d.ts
```diff
@@ -250,7 +252,7 @@ export interface RequestLog {
 }
 
 export interface RequestQuery {
-    [key: string]: any;
+    [key: string]: string | string[] | undefined;
```
kanongil, line 255: What is going on here? This is a bad choice. After validation the value can be anything. Parsing to a `number`  or a `Date` are valid choices. This should probably just be `unknown`.

lib/types/request.d.ts
```diff
@@ -269,9 +271,9 @@ export interface InternalRequestDefaults {
 
     Payload: stream.Readable | Buffer | string | object;
     Query: RequestQuery;
-    Params: Record<string, any>;
+    Params: Record<string, string>;
```
kanongil, line 274: Again, the actual type is the result of the validation step.

lib/types/request.d.ts
```diff
@@ -269,9 +271,9 @@ export interface InternalRequestDefaults {
...
     Pres: Record<string, any>;
-    Headers: Record<string, any>;
+    Headers: Record<string, string | string[] | undefined>;
```
kanongil, line 276: Again, the actual type is the result of the validation step.

typescript.md
```diff
@@ -0,0 +1,800 @@
...
+| `RouteApp`             | `RouteOptionsApp`                                 | `route.options.app`                            |
+| `Server`               | `Server`                                          | `request.server`                               |
+
+### `ReqRefDefaults`
+
+```typescript
+interface ReqRefDefaults extends InternalRequestDefaults {}
+```
+
+This is the interface you augment via `declare module` to change defaults globally. Any key you add here overrides `InternalRequestDefaults` for all routes that don't provide their own refs.
```
kanongil, line 73: With the current narrow typing, it is impossible to change eg. the `Query` default to another type. It can only be further narrowed.

kanongil (commented):
@fjeldstad argued that the narrowed types should have been in a breaking release. As already discussed, breakage is somewhat expected.

I will now argue that the types should *never have been narrowed*, as evident from my inline comments. Maybe undo this mess?

lib/types/request.d.ts
damusix: The thing is: If you don't validate, it will most certainly be `string | string[] | undefined`. This is the actual possibility of runtime data types Hapi would produce.

If you do validate, it's whatever you've set with Joi, which is a custom definition, and depends on you not using `schema.strict()`, which would make `string | string[] | undefined` true again. At that point you might as well include a custom `ServerRoute` if you're depending on the types to be correct:

```ts
ServerRoute<{
  Headers: { 
    'x-my-custom-header': number 
    'x-date': Date
  }
}>
```

`unknown | unknown[] | undefined` would imply: reality depends on the user's Joi usage and Joi configuration for ALL routes— Joi is the source of runtime truth. This seems like the wrong assumption.

kanongil: I understand your point and it definitely has merits, but ultimately I think it is flawed.

Header, query, and param validation _and transformation_ is core to using Hapi, so it doesn't make sense to have a default type that only safely applies when it is not used. It does not have to be Joi, but some kind of validation and possible transformation is very much expected.

Whether the default should be `unknown` or `any` can be debated, but given Hapi tends to prefer safety and strictness I would probably go with `unknown`, so any parameter usage would have to be explicit.

Btw, for query params the types are even more uncertain, since it also depends on the `parser` server option. With eg. the suggested `qs` module it can also contain plain objects, which is now impossible to type, as it is not a subset of `string | string[] | undefined`.

```ts
declare module "@hapi/hapi" {
    interface ReqRefDefaults {
        Query: {
            [key: string]: string | string[] | Record<string, object> | undefined;
        };
    }
}
```

…fails with this error:

```
node_modules/@hapi/hapi/lib/types/request.d.ts:298:18 - error TS2430: Interface 'ReqRefDefaults' incorrectly extends interface 'InternalRequestDefaults'.
  Types of property 'Query' are incompatible.
    Type '{ [key: string]: string | string[] | Record<string, object> | undefined; }' is not assignable to type 'RequestQuery'.
      'string' index signatures are incompatible.
        Type 'string | string[] | Record<string, object> | undefined' is not assignable to type 'string | string[] | undefined'.
          Type 'Record<string, object>' is not assignable to type 'string | string[] | undefined'.
            Type 'Record<string, object>' is missing the following properties from type 'string[]': length, pop, push, concat, and 29 more.
```

ZachHaber:
FWIW, the new type definitions are getting in my way and I'm having to deal with hundreds of type errors from this PR due to strongly typing my server routes via a helper I created a while back to infer the types of the joi schemas and apply those to the relevant pieces of the request object. The issue I'm running into with it is that anytime I am working with a helper to use a `ResponseToolkit` or `Request` directly, it complains due to unmatched types between the typed Query/Params/etc (which definitely have things other than `string | string[]`) and the index signature of those. Which is definitely the point that kanongil was making.


> @fjeldstad argued that the narrowed types should have been in a breaking release. As already discussed, breakage is somewhat expected.
> 
> I will now argue that the types should _never have been narrowed_, as evident from my inline comments. Maybe undo this mess?

I'd appreciate not having to wait to #4578 is applied via a major version (unless there is planned to be a major breaking change soon) and have the type narrowing that was done in this PR reverted back to their old values until that next major breaking change. Otherwise we end up having to deal with either not updating Hapi until the next major version or go through two sets breaking changes (this version and the major one).

Issue #4565, opened 2026-02-20
Incorrect / overly strict types in 21.4.5


joshkel:
### Runtime

Node.js

### Runtime version

24.13.1

### Module version

21.4.5

### Last module version without issue

21.4.4

### What are you trying to achieve or the steps to reproduce?

Take advantage of Joi's type handling to directly reference appropriately typed parameters in `request.params` and `request.query`. For example:

```ts
function greetUser(id: number) {
  assert(typeof id === "number", "ID must be a number");
  return `Hello, user with ID ${id}!`;
}

server.route({
  method: "GET",
  path: "/hello/{id}",
  handler: (request, _h) => {
    // request.params.id is number, not string, because of Joi validation
    return { message: greetUser(request.params.id) };
  },
  options: {
    validate: {
      params: Joi.object({
        id: Joi.number().required(),
      }),
    },
  },
});
```

### What was the result you got?

Starting in Hapi 21.4.5:

```
Argument of type 'string' is not assignable to parameter of type 'number'.ts(2345)
```

### What result did you expect?

No errors.

I believe that the stricter types in #4562 are too strict. I appreciate stricter typing (even if it breaks compilation in patch releases - #4563) but this seems _too_ strict - from what I understand, relying on Joi to validate queries and route parameters and coerce / convert their types is a core feature of Hapi and is (I assume) a common practice, but the new types act like it isn't permitted / doesn't happen.

joshkel:
As a workaround within my own codebase, I'm adding the pre-21.4.5 `Params` and `Query` definitions using [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation).

```ts
declare module '@hapi/hapi' {
  interface ReqRefDefaults {
    // Revert overly strict types in Hapi 21.4.5:
    // https://github.com/hapijs/hapi/issues/4565
    Params: Record<string, any>;
    Query: Record<string, any>;
  }
}
```

damusix:
@joshkel you can set these yourself, and you should.

https://github.com/hapijs/hapi/blob/master/typescript.md#example

By default, without Joi doing type coercion for you via validations, params WILL be a string. Typescript has no knowledge of what you're doing to convert params into a specific type. Create your own refs and you solve the problem. `any` is ambiguous. The goal in the last PR was to make types LESS ambiguous.

kanongil:
[...]

PR #4578, opened 2026-05-12
fix: broaden Query/Params/Headers ref defaults to `unknown`

damusix:
## Summary

- The 21.4.5 narrowing of `Query` / `Params` / `Headers` defaults to concrete unions blocked global `ReqRefDefaults` augmentation — users could only narrow further, not widen. That breaks legitimate cases like the `qs` query parser (which yields nested objects) and Joi-driven coercion on path params and headers.
- Defaults are now `unknown`-based (`RequestQuery` index signature, `Params`, `Headers`), per @kanongil's review on #4562 ([comment](https://github.com/hapijs/hapi/pull/4562#discussion_r3201524421)). Addresses https://github.com/hapijs/hapi/issues/4565
- Per-route narrowing via `ServerRoute<{ Query: ... }>` is unchanged. `typescript.md` and `test/types/index.ts` updated to match.

## Replication

Reproducer from the inline review:

```ts
declare module "@hapi/hapi" {
    interface ReqRefDefaults {
        Query: { [key: string]: string | string[] | Record<string, object> | undefined };
    }
}
```

Verified end-to-end against a standalone consumer file (`tsc --strict`, no `--skipLibCheck`):

- Pre-fix lib types: `TS2430: Interface 'ReqRefDefaults' incorrectly extends interface 'InternalRequestDefaults'. Types of property 'Query' are incompatible. Type 'Record<string, object>' is not assignable to type 'string | string[] | undefined'.`
- Post-fix lib types: compiles clean.

The augmentation now lives in `test/types/index.ts`, followed by assertions that the augmented `Query` propagates through every relevant generic position:

- `ServerRoute` (no generic) — `Request<ReqRefDefaults>`
- `ServerRoute<{ Params: ... }>` — partial Refs override; `Query` falls through `MergeRefs` as the augmented type
- `Lifecycle.Method`
- Per-route `ServerRoute<{ Query: ... }>` override still replaces the augmented default outright

Regression check: reverting `lib/types/request.d.ts` to the pre-fix narrow defaults fails the suite at `const id: string = request.params.id` — pre-fix `Params: Record<string, string>` makes that assignment compile, so the `@ts-expect-error` directive becomes unused.

## Caveat

The pre-fix TS2430 surfaces inside `lib/types/request.d.ts` itself, which `lab -Y` runs with `--skipLibCheck`. Regression detection inside the suite therefore comes from the route-handler `@ts-expect-error` directives, not from the library-level error directly.

### DIFF
--- a/lib/types/request.d.ts
+++ b/lib/types/request.d.ts
@@ -252,7 +252,7 @@ export interface RequestLog {
 }
 
 export interface RequestQuery {
-    [key: string]: string | string[] | undefined;
+    [key: string]: unknown;
 }
 
 /**
@@ -271,9 +271,9 @@ export interface InternalRequestDefaults {
 
     Payload: stream.Readable | Buffer | string | object;
     Query: RequestQuery;
-    Params: Record<string, string>;
+    Params: Record<string, unknown>;
     Pres: Record<string, any>;
-    Headers: Record<string, string | string[] | undefined>;
+    Headers: Record<string, unknown>;
     RequestApp: RequestApplicationState;
 
     AuthUser: UserCredentials;
--- a/test/types/index.ts
+++ b/test/types/index.ts
@@ -5,6 +5,7 @@ import * as CatboxMemory from '@hapi/catbox-memory';
 import {
     Plugin,
     ReqRef,
+    ReqRefDefaults,
     Request,
     RequestRoute,
     ResponseToolkit,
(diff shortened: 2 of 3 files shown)


hueniverse:
Changing headers to unknown broke some code... (pretend I was never here)

Marsup:
Changing types always come with a few broken builds. As long as you properly define the types for the headers you expect (and validate them accordingly), I don't expect any breakage.

damusix:
@hueniverse should be able to put things back via augmentation. The prior change was noisy, too, and seemed wrong in hindsight. Since Joi is the defacto validator, and it coerces, I figured its better to let the user define what their things are. Perhaps headers was an over-reach since it's less commonly validated?

hueniverse:
I didn't dig deeper into what's the correct approach, just wanted to provide feedback. It's going to be someone else's job to fix it... 😄

PR #4579, opened 2026-05-18
fix: replace deprecated 'aborted' event with 'close' on IncomingMessage

### LINKED ISSUE #4575: The `abort` event on nodejs is deprecated
### Runtime

node.js

### Runtime version

24

### Module version

21.4.4

Node 17

### Used with

hapi application

The `Request.active()` function always returns true. 

The `abort` event on NodeJS's `http.ClientRequest` has been deprecated since NodeJS v17, and `hapi/lib/request.js` is depending on that function.

[1] https://github.com/hapijs/hapi/blob/master/lib/request.js#L328 
[2] https://nodejs.org/api/http.html#event-abort

When spamming my F5 key on my browser with a request with an intentional delay, the request never got cancelled, and `active()` always returned true. 

The requests were never cancelled. 

I wrote a small plugin to work around the issue:
```ts
export const plugin: Plugin<object> = {
  name: 'earlyCancel',
  register(server) {
    server.ext('onPreAuth', (request, tk) => {
      request.app.disconnected = false;
      request.raw.req.on('close', () => {
        if (!request.raw.res.writableEnded) {
          request.app.disconnected = true;
        }
      });
      return tk.continue;
    });
  },
};
```
And verify that when spamming F5, I was able to get some requests to be cancelled.

sapirbaruch:
The `'aborted'` event on `http.IncomingMessage` was deprecated in Node.js v17 and no longer fires in v24+. As a result, `Request.active()` always returns `true` because `_eventContext.request` is never set to `null` on client disconnects.

The fix replaces `req.on('aborted', ...)` with `req.on('close', ...)` and guards the abort logic with a `res.writableEnded` check — the same guard used by the existing `res.on('close')` handler — to distinguish a premature client disconnect from a normal connection teardown.

Fixes #4575.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -325,7 +325,7 @@ exports = module.exports = internals.Request = class {
 
         this.raw.res.on('close', internals.event.bind(this.raw.res, this._eventContext, 'close'));
         this.raw.req.on('error', internals.event.bind(this.raw.req, this._eventContext, 'error'));
-        this.raw.req.on('aborted', internals.event.bind(this.raw.req, this._eventContext, 'abort'));
+        this.raw.req.on('close', internals.aborted.bind(this.raw.req, this._eventContext, this.raw.res));
         this.raw.res.once('close', internals.closed.bind(this.raw.res, this));
     }
 
@@ -712,6 +712,19 @@ internals.closed = function (request) {
 
     request._closed = true;
 };
+
+
+internals.aborted = function (eventContext, res) {
+
+    // The 'aborted' event on IncomingMessage was deprecated in Node.js v17 and removed in v24.
+    // Use the 'close' event on the request socket instead, guarded by a check that the response
+    // has not already been written (to distinguish a client abort from a normal connection close).
+
+    if (!res.writableEnded) {
+        internals.event(eventContext, 'abort');
+    }
+};
+
 
 internals.event = function ({ request }, event, err) {
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -325,7 +325,13 @@ exports = module.exports = internals.Request = class {
 
         this.raw.res.on('close', internals.event.bind(this.raw.res, this._eventContext, 'close'));
         this.raw.req.on('error', internals.event.bind(this.raw.req, this._eventContext, 'error'));
-        this.raw.req.on('close', internals.aborted.bind(this.raw.req, this._eventContext, this.raw.res));
+
+        // 'aborted' was deprecated in Node.js v17 and removed in v24.  It remains here for
+        // compatibility with older Node.js versions where it is the most reliable abort signal.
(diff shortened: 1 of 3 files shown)


damusix:
@sapirbaruch Can you please add tests for this?

sapirbaruch:
Added a regression test in `test/request.js` that verifies `request.active()` returns `false` when the client closes the connection before a response is sent.

The test uses `Net.connect` to make an actual TCP connection to `localhost` so it works regardless of hostname resolution, and polls `active()` directly without depending on the `disconnect` event.

damusix (approved):

damusix:
@Marsup can we get an approval to run the CI on this?

Marsup:
As expected, it fails on "old" node.js versions. I think we need both subscriptions to maintain compatibility.

EDIT: actually it even fails on latest, so I'm not sure it's quite right.

Nican:
Howdy @damusix  , looks like the tests are failing due to line coverage. 
What can we do to get this PR finished up? :o

lib/request.js
```diff
@@ -739,7 +747,12 @@ internals.event = function ({ request }, event, err) {
 
     request._eventContext.request = null;
 
-    if (event === 'abort') {
+    // On Node.js v24+, 'aborted' was removed.  A 'close' on the response before writableEnded
+    // means the client disconnected mid-request — treat it as an abort.  We exclude @hapi/shot
+    // inject requests (identified by req._shot) because shot fires req.emit('close') for its
+    // simulate.close scenario but that should not be treated as a client abort.
+    if (event === 'abort' ||
+        (event === 'close' && !request.raw.res.writableEnded && !request.raw.req._shot)) {
```
kanongil: We just published a new version of shot that might change the behaviour. Regardless, we should probably not specialcase shot here, but rather revise the tests. Treating it as an aborted request seems quite appropriate.

kanongil:
> looks like the tests are failing due to line coverage. What can we do to get this PR finished up? :o

I guess add more tests 😉

FYI, you can do `npm run test-cov-html` to generate a `coverage.html` that will provide more useful details on why it fails the check.

sapirbaruch:
Hey @kanongil — just wanted to flag that the coverage issues you mentioned on May 27 were addressed in the May 28 push. All CI checks are green across Ubuntu and Windows (Node 14–latest). Would appreciate a second look when you get a chance!

kanongil (changes requested):
Wait, what is going on here? The added test doesn't fail for me when using the current logic (on any node release).

Also the claim that "IncomingMessage 'aborted' event was removed in Node.js v24." is patently false, and a pure hallucination.

kanongil:
There might be an underlying bug in the original issue, but this PR is just wasting our time.

PR #4580, opened 2026-05-22
Emit request errors after aborted requests

### LINKED ISSUE #4567: Error-tracking software doesn't catch hapi application errors if an incoming request gets prematurely closed
### Runtime

Node.js

### Runtime version

v25.7.0

### Module version

v21.4.7

_No response_

### Used with

Sentry, Rollbar, DataDog

Sentry, Rollbar, and DataDog support tracking errors in hapi applications, relying on either hapi's `request` `error` event or on `onPreResponse` lifecycle method. See:
- https://github.com/getsentry/sentry-javascript/blob/master/packages/node/src/integrations/tracing/hapi/index.ts#L67
- https://docs.rollbar.com/docs/nodejs#using-hapi
- https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-instrumentations/src/hapi.js#L33

Here is example code that reproduces an issue where an error thrown in a handler bypasses both the `request` `error` event and `onPreResponse` lifecycle method calls.

```js
import Hapi from "@hapi/hapi";
import { setTimeout as sleep } from "node:timers/promises"

const server = Hapi.server({
    port: 3000,
    host: "localhost"
});

server.events.on({ name: "request", channels: ["error"] }, (request, event, tags) => {
    console.log("[server] on request error"); // NOT called
})

server.ext("onPreResponse", async (request, h) => {
   console.log("[server] onPreResponse"); // NOT called
   return request;
});

server.route({
    method: "GET",
    path: "/sleep/{seconds}/then-throw-error",
    handler: async (request, h) => {
        const { seconds } = request.params;
        for (let i = 0; i < seconds; i++) {
            console.log("[handler] Handling a request... (sleep %s)", i + 1);
            await sleep(1_000);
        }
        console.log("[handler] Throwing error");
        throw new Error("Boom!");
    }
});

await server.start();
console.log("[server] Running on %s", server.info.uri);

try {
    console.log("[fetch] Making a request");
    await fetch("http://localhost:3000/sleep/5/then-throw-error", { signal: AbortSignal.timeout(1_000) });
} catch (error) {
    if (error.name === "TimeoutError") {
        console.log("[fetch] Stopped waiting for the response");
    } else {
        console.log(error);
    }
}
```

Here is the output produced by the example code:

```plain
[server] Running on http://localhost:3000
[fetch] Making a request
[handler] Handling a request... (sleep 1)
[fetch] Stopped waiting for the response
[handler] Handling a request... (sleep 2)
[handler] Handling a request... (sleep 3)
[handler] Handling a request... (sleep 4)
[handler] Handling a request... (sleep 5)
[handler] Throwing error
```

Neither `server.events.on({ name: "request", channels: ["error"] }, fn)` nor `server.ext("onPreResponse", fn)` callback functions are called if the incoming request gets prematurely closed. However, if we increase the request timeout, e.g., `AbortSignal.timeout(10_000)`, both callback functions will get called.

I'm not sure if the shown hapi behavior is a bug or if it works as intended, especially because the `onPreResponse ` documentation [mentions](https://hapi.dev/api/?v=21.4.3#request-lifecycle) that it is “always called, _unless the request is aborted_.” Nevertheless, it's evident that there is existing error-tracking software that seems to be expecting to capture _all_ errors using the approach. The fact that some errors are not getting captured is an issue. I would like to start a discussion to figure out a way forward on how to improve error tracking around aborted requests.

puneetdixit200:
Summary:
- emit the public request error event when a late 500 error is produced after an aborted request has already finalized
- reuse the same error-response logging helper for normal finalize and completed-response paths
- add a regression test for a handler that throws after the client disconnects

Tests:
- `npm_config_cache=/tmp/hapi-npm-cache npx lab -a @hapi/code -m 5000 test/request.js -g "emits request-error when handler fails after abort|does not fail on abort|emits request-error once"`
- `source ~/.nvm/nvm.sh && nvm use 18.20.8 >/dev/null && npm_config_cache=/tmp/hapi-npm-cache npx lab -a @hapi/code -m 5000 test/request.js -g "emits request-error when handler fails after abort|does not fail on abort|emits request-error once"`
- `git diff --check`

Fixes #4567

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -544,8 +544,17 @@ exports = module.exports = internals.Request = class {
 
     _logErrorResponse(response) {
 
-        const error = response._error ?? (Boom.isBoom(response) ? response : null);
+        const isBoom = Boom.isBoom(response);
+        const error = response._error ?? (isBoom ? response : null);
         const statusCode = response.statusCode ?? response.output?.statusCode;
+
+        if (isBoom &&
+            response.output.statusCode === this.route.settings.response.disconnectStatusCode &&
+            ['Request aborted', 'Request close', 'Response error'].includes(response.message)) {
+
+            return;
+        }
+
         if (statusCode !== 500 ||
             !error) {
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -507,14 +507,9 @@ exports = module.exports = internals.Request = class {
 
         this._eventContext.request = null;              // Disable req events
 
+        this._logErrorResponse(this.response);
+
         if (this.response._close) {
-            if (this.response.statusCode === 500 &&
-                this.response._error) {
-
-                const tags = this.response._error.isDeveloperError ? ['internal', 'implementation', 'error'] : ['internal', 'error'];
-                this._log(tags, this.response._error, 'error');
-            }
-
             this.response._close();
         }
 
@@ -538,12 +533,27 @@ exports = module.exports = internals.Request = class {
         }
 
         if (this.info.completed) {
+            this._logErrorResponse(response);
             response._close?.();
 
             return;
         }
 
         this.response = response;
+    }
+
+    _logErrorResponse(response) {
+
+        const error = response._error ?? (Boom.isBoom(response) ? response : null);
+        const statusCode = response.statusCode ?? response.output?.statusCode;
+        if (statusCode !== 500 ||
+            !error) {
+
+            return;
+        }
+
+        const tags = error.isDeveloperError ? ['internal', 'implementation', 'error'] : ['internal', 'error'];
+        this._log(tags, error, 'error');
     }
 
     _setState(name, value, options) {
--- a/test/request.js
+++ b/test/request.js
@@ -766,6 +766,43 @@ describe('Request', () => {
             await server.stop();
         });
 
+        it('emits request-error when handler fails after abort', { retry: true }, async (flags) => {
+
+            const server = Hapi.server({ debug: false });
+            const team = new Teamwork.Team();
+
+            const handler = async (request) => {
+
+                clientRequest.destroy();
+                await Hoek.wait(10);
+                team.attend();
+                throw new Error('late failure');
+            };
+
+            server.route({ method: 'GET', path: '/', handler });
+
+            const log = server.events.once({ name: 'request', channels: 'error' });
+
+            await server.start();
+            flags.onCleanup = () => server.stop();
+
+            const clientRequest = Http.request({
+                hostname: 'localhost',
+                port: server.info.port,
+                method: 'GET'
+            });
+
+            clientRequest.on('error', Hoek.ignore);
+            clientRequest.end();
+
+            await team.work;
+
+            const [, event] = await log;
+            expect(event.error.message).to.equal('late failure');
+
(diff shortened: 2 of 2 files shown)


damusix:
@puneetdixit200 Thanks for the contribution ... something I picked up doing some testing as an edge case: double `request-error` emit when `disconnectStatusCode: 500`

The fix works correctly with the default `disconnectStatusCode` (499), but double-emits when a route sets it to 500:

```js
server.route({
    method: 'GET',
    path: '/',
    options: {
        response: { disconnectStatusCode: 500 }
    },
    handler: async (request) => {

        const result = await someSlowOperation();  // client disconnects during this
        return result;                               // then this throws
    }
});
```

**What happens:**
1. Client disconnects → `_finalize()` → `_logErrorResponse()` logs the abort Boom (500) — 1st emit
2. Handler throws after finalize → `_setResponse()` → `_logErrorResponse()` logs the handler error (500) — 2nd emit

With the default 499, step 1 skips logging (not a 500), so only step 2 fires. But `disconnectStatusCode` validates as any integer >= 400, so 500 is legal.

**Reproduction test** (currently fails against this PR — `emitCount` is 2):

```js
it('does not double-emit request-error when disconnectStatusCode is 500', { retry: true }, async (flags) => {

    const server = Hapi.server({ debug: false });
    const team = new Teamwork.Team();
    let emitCount = 0;
    const errors = [];

    server.events.on({ name: 'request', channels: 'error' }, (request, event) => {

        emitCount++;
        errors.push(event.error.message);
    });

    server.route({
        method: 'GET',
        path: '/',
        handler: async (request) => {

            clientRequest.destroy();
            await Hoek.wait(10);
            team.attend();
            throw new Error('late failure');
        },
        options: {
            response: {
                disconnectStatusCode: 500
            }
        }
    });

    await server.start();
    flags.onCleanup = () => server.stop();

    const clientRequest = Http.request({
        hostname: 'localhost',
        port: server.info.port,
        method: 'GET'
    });

    clientRequest.on('error', Hoek.ignore);
    clientRequest.end();

    await team.work;
    await Hoek.wait(50);

    expect(emitCount).to.equal(1);
    expect(errors[0]).to.equal('late failure');

    await server.stop();
});
```

Perhaps, if we know aborted requests as 499, we could specifically check for it to prevent the double log? 

This is a very niche case and I have no idea who configures their routes this way.

puneetdixit200:
Pushed `106f641d` for the `disconnectStatusCode: 500` double-emit case.

What changed:
- hapi-generated disconnect Boom responses (`Request aborted`, `Request close`, `Response error`) are no longer logged as internal 500 request errors.
- late handler failures after an abort still emit `request-error` once.
- added the `disconnectStatusCode: 500` regression test from the review case.

Verified locally:
- `npx lab -a @hapi/code -m 5000 test/request.js -g "handler fails after abort|disconnectStatusCode is 500"`
- `npx lab -a @hapi/code -m 5000 test/request.js`
- `npx eslint lib/request.js test/request.js`
- `git diff --check`

kanongil (changes requested):
This is definitely the wrong fix. Requests should not error once aborted.

Issue #4527, opened 2024-08-28
The application crashes when setting an invalid cookie value


Hydrock:
### Runtime

node.js

### Runtime version

v18.20.0 / LTS

### Module version

21.3.2

### Any other relevant information

When the server tries to set an invalid cookie value, the server crashes with an error:

```
node:_http_outgoing:662
  validateHeaderValue(name, value);
  ^

TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["set-cookie"]
    at ServerResponse.setHeader (node:_http_outgoing:662:3)
    at internals.writeHead (/project/workspace/node_modules/@hapi/hapi/lib/transmit.js:336:21)
    at internals.transmit (/project/workspace/node_modules/@hapi/hapi/lib/transmit.js:104:15)
    at internals.fail (/project/workspace/node_modules/@hapi/hapi/lib/transmit.js:68:22)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Request._reply (/project/workspace/node_modules/@hapi/hapi/lib/request.js:456:9) {
  code: 'ERR_INVALID_CHAR',
  isBoom: true,
  isServer: true,
  data: null,
  output: {
    statusCode: 500,
    payload: {
      statusCode: 500,
      error: 'Internal Server Error',
      message: 'An internal server error occurred'
    },
    headers: {}
  }
}
```

Slack chat discussion: https://hapihour.slack.com/archives/C6CEEUE06/p1724861065721169

Problem example: https://codesandbox.io/p/devbox/hapi-cookie-test-forked-t2g89w?workspaceId=6ecda139-50d1-4062-98aa-206f54ef133e

### What are you trying to achieve or the steps to reproduce?

Problem example: https://codesandbox.io/p/devbox/hapi-cookie-test-forked-t2g89w?workspaceId=6ecda139-50d1-4062-98aa-206f54ef133e

Just start the server at the address "/" and you will immediately see the server crash. This is because of the line:
```
// INVALID cookie
.state("cookieName2", "тест");
```

in slack's chat, a user under the nickname "yoannma" wrote the following:
```
I think I found why node crash :
- hapi try to serialize the hapi.response header to node.response in writeHead (https://github.com/hapijs/hapi/blob/master/lib/transmit.js#L336)
- node throw an error because of the bad characters
- hapi catch it, remove the headers, throw a boomify error (https://github.com/hapijs/hapi/blob/master/lib/transmit.js#L345)
[...]

Issue #4316, opened 2021-11-11
Exception when using takeover response in onPreResponse and the server is configured with autoValue


sveisvei:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*: v14.17.6
* *module version with issue*:  "@hapi/hapi": "^20.2.1",
* *last module version without issue*: ?
* *environment* (e.g. node, browser, native):  node
* *used with* standalone
* *any other relevant information*:

#### What are you trying to achieve or the steps to reproduce?

I am using the cookie state autoValue for a persistent id. 
When trying to use the `onPreResponse` to takeover a response, it fails because request.state is null.

```
TypeError: Cannot use 'in' operator to search for 'tid' in null
    at exports.state (/node_modules/.pnpm/@hapi+hapi@20.2.1/node_modules/@hapi/hapi/lib/headers.js:88:57)
    at Object.internals.marshal (/node_modules/.pnpm/@hapi+hapi@20.2.1/node_modules/@hapi/hapi/lib/transmit.js:41:15)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async Object.exports.send (/node_modules/.pnpm/@hapi+hapi@20.2.1/node_modules/@hapi/hapi/lib/transmit.js:27:9)
    at async Request._reply (/node_modules/.pnpm/@hapi+hapi@20.2.1/node_modules/@hapi/hapi/lib/request.js:457:5)
```

Reproducible demo:
```js
const Hapi = require("@hapi/hapi");
const { isBoom } = require('@hapi/boom');

const server = new Hapi.Server({
  port: 3333,
});

server.state("tid", {
  ttl: 1000 * 60 * 60 * 24 * 31 * 3, // 3 months
  isSecure: false,
  isHttpOnly: true,
  encoding: "none",
  clearInvalid: true,
  strictHeader: true,
  autoValue: () => 'v',
  path: "/",
});

server.ext("onPreResponse", (request, h) => {
  const err = request.response;
  if (isBoom(err)) {
    return h.response("test").code(err.output.statusCode);
  }
  return h.continue;
});

server.start();
console.log(server.info.uri);

```

#### What was the result you got?

Exception.

#### What result did you expect?

I would expect that onPreResponse can be taken over.

sveisvei:
[...]

PR #4586, opened 2026-07-21
Fix cookie processing issues

### LINKED ISSUE #4297: Error responses include state
#### Support plan

#### Context

* *node version*: 
* *module version with issue*: 20.2.0
* *last module version without issue*: 

`'set-cookie` header is generated for responses here:

https://github.com/hapijs/hapi/blob/7b4d7d89bb5be734aa23b75a89117ae279c4069e/lib/headers.js#L104-L110

It does not use information from response to generate this. Only from the request.

Created a test case to demonstrate:

```js
    it('does not set cookie on response errors', async () => {

        const server = Hapi.server();
        server.state('data', { encoding: 'base64json' });
        server.route({ method: 'GET', path: '/', handler: (request, h) => h.response('ok').state('data', { b: 3 }) });
        server.ext('onPostHandler', () => Hoek.wait(50), { timeout: 1 });

        const res = await server.inject('/');
        expect(res.headers['set-cookie']).to.not.exist();
    });
```

```
Expected [ 'data=eyJiIjozfQ==; Secure; HttpOnly; SameSite=Strict' ] to not exist
```

Pass.

kanongil:
This fixes #4297, #4527 and #4316.

Not allowing cookies to be set on error 500 responses is an API change, but unlikely to affect usage. I made an effort to preserve cookie clears, since it can help avoid processing errors on future requests, and it doesn't have the same risk as setting a full cookie.

Not setting `autoValue` cookies when state parsing is disabled is not an API change as such, since this is how it already works in practice.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2540,7 +2540,7 @@ across multiple requests. Registers a cookie definitions where:
 
     - `domain` - the domain scope. Defaults to `null` (no domain).
 
-    - `autoValue` - if present and the cookie was not received from the client or explicitly set by the route handler, the cookie is automatically added to the response with the provided value. The value can be a function with signature `async function(request)` where:
+    - `autoValue` - if present and the cookie was not received from the client or explicitly set by the route handler, the cookie is automatically added to the response with the provided value, unless state parsing is disabled or the response code is 500. The value can be a function with signature `async function(request)` where:
 
         - `request` - the [request object](#request).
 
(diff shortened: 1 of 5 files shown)


kanongil:
Note that regular responses with code 500 will still set state cookies.

Marsup (approved):

PR #4588, opened 2026-08-25
fix(types): add QUERY http method

SimonSchick:
Idk if this requires a test since this seems like a very very basic addition.

https://www.rfc-editor.org/info/rfc10008/


lbesecker195 (approved):
Verified: `require('http').METHODS` includes `QUERY` on Node 24, and this type is documented as mirroring Node's supported method list. LGTM.

Marsup:
Thanks.

